PR #84, opened 2015-02-16
Netflixoss build

rspieldenner:


tgianos:
Hey @rspieldenner,

A few questions about this pull request I wanted to check with you on.

1. Why did you take the common dependencies from the root build.gradle and distribute them to all the subprojects? This seems to create more work when we want to go to update these packages as we need to remember to do it in all the projcts.
2. Why did you take the () wrappings off all the dependencies that didn't explicitly need them? Do they hurt something? I had left them on just to make it easier if we ever had to add some closures.
3. What is the value of putting jcenter() above mavenCentral() for the dependencies?
4. For the code analysis stuff I believe I put in the findbugs toolversion explicitly since we weren't on gradle 2.+ yet, now that we will be is that still needed?
5. I understand why you changed the lines like ```compile project.parent.childProjects.get('genie-common')``` to ```compile project(':genie-common')``` as this is the standard and cleaner way to do this. However it was done to support building internal Genie via submodule and wrapping the OSS build. I believe this way didn't work in the past which is why I had to change it. Can you change it back or we'll need to figure out another solution for this which may be more work than we can do before 2.1.0 release needs to go?

Thanks,
Tom

rspieldenner:
1) I have found that the grand majority of people use that top level dependencies block as a dumping ground for everything and aren't good about keeping them minimal. So my default is to have all dependencies for a project in one block. I also find it easier to reason about dependencies if they're all in one place instead of needing to look at two. If you don't like it this way you or I can change it back.
2) I try to keep my build.gradle as clean as possible so parentheses got deleted out of habit.
3) I like bintray and jcenter better than mavenCentral. mavenLocal should be a fallback not first.
4) didn't check
5) hopefully we can do your internal build cleaner than it currently is

PR #116, opened 2015-04-08
Added (simpler) Groovy sample and reviewed the Java client

liviutudor:
2 streams of changes here:

* on the Java client, the way the query is passed as a `FileAttachment` is way too convoluted and also incurs extra I/O due to going to disk when in fact everything can be done in memory.
* Since I know there are some of us in Netflix using Groovy, and due to the fact that Groovy has a lot of syntactical sugar around initialization and properties setting, I've put together a Groovy client sample too which is much more concise I feel than the Java one and as such might make it easier to understand the client (as well as help folks who rely on Groovy).

### DIFF
--- a/genie-client/build.gradle
+++ b/genie-client/build.gradle
@@ -6,7 +6,7 @@ dependencies {
      *******************************/
 
     compile project(":genie-common")
-    compile "org.codehaus.groovy:groovy:2.3.3"
+    compile "org.codehaus.groovy:groovy:$groovy_lang_version"
 
     /*******************************
      * Provided Dependencies
--- a/genie-client/build.gradle
+++ b/genie-client/build.gradle
@@ -1,16 +1,12 @@
 apply plugin: 'groovy'
 
-// module versions
-groovyVersion="2.3.3"
-
-
 dependencies {
     /*******************************
      * Compile Dependencies
      *******************************/
 
     compile project(":genie-common")
-    compile "org.codehaus.groovy:groovy:$groovyVersion"
+    compile "org.codehaus.groovy:groovy:2.3.3"
 
     /*******************************
      * Provided Dependencies
--- a/genie-client/build.gradle
+++ b/genie-client/build.gradle
@@ -1,9 +1,16 @@
+apply plugin: 'groovy'
+
+// module versions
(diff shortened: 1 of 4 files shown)


genie-client/src/main/groovy/com/netflix/genie/client/sample/GExecutionServiceSampleClient.groovy
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A sample client demonstrating usage of the Execution Service Client.
+ *
+ * @author skrishnan
+ * @author tgianos
```
tgianos: @liviutudor You probably want to set the author to yourself here rather than myself or Shriram

genie-client/build.gradle
```diff
@@ -1,9 +1,12 @@
+apply plugin: 'groovy'
+
 dependencies {
     /*******************************
      * Compile Dependencies
      *******************************/
 
     compile project(":genie-common")
+    compile "org.codehaus.groovy:groovy:2.3.3"
```
tgianos: @liviutudor Can you follow the pattern of putting this version into ```gradle.properties``` and then reusing it as a variable? We'd like to keep the versions for things in one place so we don't have to hunt through all the build files for them.

genie-client/src/main/groovy/com/netflix/genie/client/sample/GExecutionServiceSampleClient.groovy
```diff
@@ -0,0 +1,117 @@
...
+import java.util.List;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A sample client demonstrating usage of the Execution Service Client.
+ *
+ * @author liviutudor
```
liviutudor: Btw since this is on github I used my github username rather than the OneLogin one. Happy to use the Netflix one if it's better.

tgianos:
Merged. Thanks @liviutudor!

PR #147, opened 2015-08-04
Feature/99

tgianos:
Please review the changes for Java 8.

If necessary we can later set the genie client and genie common to be java 7 compile level after talking to Rob.


amitsharma1981:
Looks good. Ship it.

PR #172, opened 2015-12-07
Added async notification in the ExecutionClient

liviutudor:
Added a feature where user can supply a callback and ask to be notified async through the callback when job is finished.


liviutudor:
All efforts are now on version 3.0 so closing this without merge.

PR #183, opened 2016-02-10
Job Output Directory Support For Genie 3.0

tgianos:
This pull request enables handling of the job output directory for Genie 3.0.

Since Genie 3.0 runs using [Spring Boot](http://projects.spring.io/spring-boot/), which includes an embedded Tomcat, instead of within an independent Tomcat process the output directory processing had to change. 

In Genie 2.x we would configure the job output directory as another "web application" via symlink to a directory on the filesystem within the tomcat webapps directory. From there Tomcat would handle serving that content using the [DefaultServlet](https://tomcat.apache.org/tomcat-8.0-doc/default-servlet.html) and enabling directory listing.

This wouldn't work out of the box in Spring Boot. Also the old solution kicked the user out of the Genie context which would cause a lot of problems for the new security put in place via #180. Basically the anyone could have accessed the output files which obviously could contain data that people shouldn't have access to.

This PR contains code which adds a REST API under ```/api/v3/jobs/{id}/output/**``` which handles listing directories and serving static files. Static file support is done using Springs' [ResourceHttpRequestHandler](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/servlet/resource/ResourceHttpRequestHandler.html). This leaves in place all the advanced file handling they already have. Support for directories is accomplished by extending that class and checking if the requested resource on disk is a directory. If it is the handling bypasses Springs implementation and can return either a HTML or JSON representation of the directory contents.

The other important feature of this PR is forwarding of requests for these static resources. In Genie 2.x due to the design explained above with the symlinked directory there was no control in Genie what resources the users were trying to request as there was no application code around the requests. This PR adds a feature that allows admins to turn on or off Job Output Directory forwarding. When run as a cluster this allows users to send requests for resources to any Genie node and have it forwarded to the node where the resource "should" be (unless it was deleted by some janitor type process). This can be turned off if the job directory is backed by some shared NFS type directory like [EFS](https://aws.amazon.com/efs/). If it is on though it removes the "stickiness" of needing to send requests to the proper nodes in Genie 2 and allows running Genie behind a load balancer or similar.

The HTML output of the directories are likely to change but want to get this code merged into the baseline to avoid conflicts with other development efforts going on.


amitsharma1981:
Looks good.It's weird there is no library out there just lets us do the listing automatically instead of us rendering the page in Genie. Will work well for us for the time being though.

Issue #171, opened 2015-11-12
System periodically runs a monitor process on each node to check health/status of all jobs run on the node


amitsharma1981:
Each node in the Genie cluster periodically starts a Monitor process which checks the health/status of jobs running on the node. This process relies on the existence of the pid of the job to check if its running. If the pid does not exist it assumes that the process is done and look for the return code in the genie.done file. If it cannot find the genie.done file or the return code in it, it marks the job as failed. If it finds the return code, it updates the job status appropriately.

tgianos:
Handled by #189 and #190

PR #189, opened 2016-02-24
New Job Monitoring Subsystem

tgianos:
This pull request introduces a new Job Monitoring mechanism for Genie 3.0.

When a job is started (via the Job Execution system @amitsharma1981 is working on) a ```JobStartedEvent``` should be fired. When this happens the ```JobMonitoringCoordinator``` is listening to the events and will create a new ```JobMonitor``` for the running job. This ```JobMonitor``` will then be scheduled to run with fixed delay period with the system [TaskScheduler](http://docs.spring.io/autorepo/docs/spring/current/javadoc-api/org/springframework/scheduling/TaskScheduler.html).

The ```JobMonitor``` when its run method is invoked will execute a ```PS``` on the system for the job client PID. Once this comes back with a non-zero exit code it will fire a ```JobFinishedEvent``` which the ```JobMonitoringCoordinator``` is also listening for to cancel execution of the ```JobMonitor```.

This system was designed to use events so as to take advantage of extensibly adding features which need to key off key events like jobs being started and completed without us re-writing a lot of interfaces to take new classes.

This PR doesn't include complete checking for the length of stdout or stderr and also is missing a flexible way to schedule the duration of the time between checks of the job monitor. These features will be added later.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/events/JobFinishedEvent.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/events/JobFinishedEvent.java
@@ -1,0 +1,50 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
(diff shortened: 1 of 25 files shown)


tgianos:
This PR closes #171

PR #197, opened 2016-03-11
Job execution framework implementation

amitsharma1981:
This pull request adds logic for the re-design of the Job Execution framework in Genie. 

Below are the highlights of how the job is executed:

- A JobRequest is accepted by a coordinator containing all the details needed to select the cluster/command command combination as well as files/arguments needed to run the job. 
- User can also submit a multi-form upload request which contains the JobRequest as well as an array of attachments. If attachments are specified Genie downloads the attachment on the current host.
- The JobRequest object and its corresponding Job object is persisted using the job persistence service.
- The coordinator hands off the request to the submitter service. A local run implementation of the interface generates the working directory for the job.
- A File transfer service is implemented to fetch all the dependencies for the job to the current working directory.
- A run.sh file is then dynamically generated that sets all the environment variables , add source commands for setup files and generates the execute command.
- The run.sh script is then kicked off and the process id of the process is persisted as a part of the JobExecution object. The current thread creates a JobStartedEvent and is detached from this job flow execution.
- A centralized monitoring service running on each node on receiving the event for JobStart keeps periodically checking the status of the process id waiting for it to finish. Once the process no longer exists, the monitoring thread generates a JobCompletionEvent.
- A Job completing handler performs the following tasks on receiving this event: Update job status based on genie.done file, archive job directory if enabled and requested & send a job completion email if enabled and requested.


genie-common/src/main/java/com/netflix/genie/common/util/Constants.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Protected constructor for utility class.
+     */
+    protected Constants() {
+    }
+
+    /**
+     * Enum containing types of files that the entities hold.
+     */
+    public enum FileType {
```
tgianos: I would move these enum's to their own class files so they're easier to find/modify.

genie-common/src/main/java/com/netflix/genie/common/util/Constants.java
```diff
@@ -0,0 +1,235 @@
...
+package com.netflix.genie.common.util;
+
+/**
+ * A class holding some constants to be used everywhere.
+ *
+ * @author amsharma
+ * @author tgianos
+ * @since 3.0.0
+ */
+public final class Constants {
```
tgianos: Since most of these Constants are used in the core project would it make more sense to move this file there? Are any of them even used in Common or would be used by something like the client? Unlikely from the looks of the constants

genie-core/build.gradle
```diff
@@ -16,12 +16,18 @@ dependencies {
...
     compile("commons-httpclient:commons-httpclient")
     compile("commons-io:commons-io")
+    compile("org.apache.commons:commons-exec:${commons_exec_version}")
 
     // Spring Libs
     compile("org.springframework.boot:spring-boot-starter-aop")
     compile("org.springframework.boot:spring-boot-starter-data-jpa")
     compile("org.springframework.boot:spring-boot-starter-mail")
     compile("org.springframework.retry:spring-retry")
+    compile("org.springframework.cloud:spring-cloud-aws-context:1.0.4.RELEASE")
```
tgianos: Move the version of these libraries to gradle.properties and reference.

does the java sdk not come as a transitive dependency of the spring cloud AWS dependency?

genie-core/src/main/java/com/netflix/genie/core/jobs/JobDoneFile.java
```diff
@@ -17,24 +17,17 @@
...
-     *
-     * @param jr The job request object
-     * @return jobExecutionEnvironment
-     */
-    public JobExecutionEnvironment build(final JobRequest jr) {
-        return null;
-    }
+@Getter
+@Setter
+public class JobDoneFile {
```
tgianos: Is there a case to be made to add a "reason" type field to this done file that the execution script can leverage? That way when we populate the job status message field of the job object in the database we can use this field to populate it if need be?

genie-core/src/main/java/com/netflix/genie/core/jobs/JobExecutionEnvironment.java
```diff
@@ -46,112 +38,88 @@
...
+        private List<Application> bApplications = new ArrayList<>();
+        private File bJobWorkingDir;
+
+        /**
+         * Constructor.
+         *
+         * @param request The job request object.
+         * @param clusterObj The cluster object.
+         * @param commandObj The command object.
+         * @param dir The directory location for the jobs
```
tgianos: is this the directory for all the jobs or just this particular job?

PR #202, opened 2016-03-22
Application Overrides

tgianos:
This pull request adds the ability for users to override the default set of applications used to execute a job . It is a completely optional parameter intended to be used by advanced users or administrators testing new applications before releasing them for general use.

This is accomplished by adding a list of application id's in the Job Request in the ```applications``` field. If this isn't set the system uses the default applications linked to the command if any.

Also in this pull request is adding search filtering by application type. Lots of changes to various classes and cleanup. Moved all searching to JobSearchService as opposed to proxying through ```JobCoordinatorService```.


genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
```diff
@@ -361,6 +366,44 @@ public void setMemory(@Min(1) final int memory) {
...
+     * Gets the applications for the job as JSON array.
+     *
+     * @return applications
+     */
+    protected String getApplications() {
+        return this.applications;
+    }
+
+    /**
+     * Sets the dependencies for the job.
```
amitsharma1981: dependencies should be applications?

tgianos: Two different things. Dependencies are the files (queries, etc) the job needs to run. Applications are the id's of applications the user wants to override the default set of applications for the selected command.

PR #205, opened 2016-03-23
Added logic to handle kill signals in the bash script

amitsharma1981:
The run.sh script for each job now has a handle_kill_request method which gets triggered once it traps a SIGTERM (kill or kill -15) signal. This initiates a kill sequence as follows:

0. Writes the Kill exit code (999) to genie.done file.
1. The base process (run.sh) sends a kill to its children
2. Then it checks every second for the next 30 seconds if the children have been killed
3. If all children are dead it exits.
4. If all children are not dead it sends a kill -9 to all children

The behavior of kill is slightly different between Linux and Non-Linux OS.

For a linux OS , the system launches the run.sh script using setsid. This results in run.sh, its children and all its grandchildren to launch in the same process group which is the pid of run.sh. Once this is done the kill logic uses the "-g" flag to send the kill and kill -9 signals to all at the processes in the process group. This signal is send to run.sh itself in this case. The kill signal is handled by run.sh and ignored, but the kill -9 kills the run.sh as well, which is fine as at this point there is nothing else it can do.

For a Non-Linux OS we simply launch the run.sh as a normal process and use its pid to send kill and kill -9 signals to its immediate children only, and hope that they handle the killing of their own children.


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -107,6 +108,12 @@ public void executeTask(
             command.add("-u");
             command.add(this.jobExecEnv.getJobRequest().getUser());
         }
+
+        // If the OS is linux use setsid to launch the process so that the entire process tree
+        // is launched in process group id which is the same as the pid of the parent process
+        if (SystemUtils.IS_OS_LINUX) {
+            command.add("setsid");
+        }
```
tgianos: If we're going to start having lots of OS level differences I'd say we should create subclasses for LinuxJobKickoffTask, OSXJobKickoffTask, etc... that are created as beans based on what OS it is. That may well be overkill for right now but it's something to keep in mind.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKillLogicTask.java
```diff
@@ -0,0 +1,33 @@
...
+    public void executeTask(
+        @NotNull
+        final Map<String, Object> context
+    ) throws GenieException {
+
+        log.debug("Executing JobKillLogic Task in the workflow.");
+        super.executeTask(context);
+
+        // Append logic for handling job kill signal
+        Utils.appendToWriter(writer, JobConstants.JOB_KILL_HANDLER_LOGIC.toString());
```
tgianos: Instead of calling toString over and over here you should just call it once in the constant class when it is created.

PR #206, opened 2016-03-28
Added failure handling capabilities to run.sh

amitsharma1981:
- Added logic in run.sh to handle failures from exports/source statements.
- Added more formatting to run.sh
- Added an integration test to validate contents of the dynamically generated run.sh with a static expected run.sh based on OS
- Made logic of archiving a job non optional
- Job Completion Handler no longer throws any exceptions as it runs in an asynchronous manner. Added TODOs in the the class to add metrics instead which we would eventually need to set alerts or maybe dynamically mark the node as unhealthy.


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ApplicationTask.java
```diff
@@ -80,7 +81,14 @@ public void executeTask(
...
                     this.fts.getFile(applicationSetupFile, localPath);
-                    Utils.appendToWriter(writer, "source " + localPath + ";");
+                    Utils.appendToWriter(writer, "# Sourcing setup file from application " + application.getId());
+                    Utils.appendToWriter(writer,
+                        JobConstants.SOURCE
+                        + localPath.replace(this.jobWorkingDirectory, "${" + JobConstants.GENIE_JOB_DIR_ENV_VAR + "}")
+                        + JobConstants.SEMICOLON_SYMBOL);
+
+                    // Append new line
+                    Utils.appendToWriter(writer, " ");
```
tgianos: why does the new line need to be a space? why not just a ```\n```

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/CommandTask.java
```diff
@@ -78,7 +79,15 @@ public void executeTask(
...
-            Utils.appendToWriter(writer, "source " + localPath + ";");
+
+            Utils.appendToWriter(writer, "# Sourcing setup file from command " + jobExecEnv.getCommand().getId());
+            Utils.appendToWriter(writer,
+                JobConstants.SOURCE
+                    + localPath.replace(this.jobWorkingDirectory, "${" + JobConstants.GENIE_JOB_DIR_ENV_VAR + "}")
+                    + JobConstants.SEMICOLON_SYMBOL);
+
+            // Append new line
+            Utils.appendToWriter(writer, " ");
```
tgianos: this looks like pretty much the exact same code that is in the cluster task file? Can't it be factored out into reusable?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobTask.java
```diff
@@ -69,7 +69,15 @@ public void executeTask(
...
-            Utils.appendToWriter(writer, JobConstants.SOURCE + localPath + JobConstants.SEMICOLON_SYMBOL);
+
+            Utils.appendToWriter(writer, "# Sourcing setup file specified in job request");
+            Utils.appendToWriter(writer,
+                JobConstants.SOURCE
+                    + localPath.replace(this.jobWorkingDirectory, "${" + JobConstants.GENIE_JOB_DIR_ENV_VAR + "}")
+                    + JobConstants.SEMICOLON_SYMBOL);
+
+            // Append new line
+            Utils.appendToWriter(writer, " ");
```
tgianos: Same code again

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobTask.java
```diff
@@ -101,7 +114,11 @@ public void executeTask(
         // Wait for the above process started in background mode. Wait lets us get interrupted by kill signals.
         Utils.appendToWriter(writer, "wait $!");
 
+        // Append new line
+        Utils.appendToWriter(writer, " ");
+
         // capture exit code and write to genie.done file
+        Utils.appendToWriter(writer, "# Write the return code from the command in the done file.");
         Utils.appendToWriter(writer, JobConstants.GENIE_DONE_FILE_CONTENT_PREFIX + JobConstants.GENIE_DONE_FILE_NAME);
```
tgianos: Instead of calling appendToWriter so many times why not build up a List<String> and pass the whole list at the end to the method once and just iterate through and write it out. Puts less function calls on the stack. Not sure if the compiler would optimize that out or not but seems like it might clean the code up a bit. Though possible that calling list.add would be the same.... not sure just seems like there's a lot of calls to this util method that maybe there is a cleaner way.

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionHandler.java
```diff
@@ -214,23 +223,28 @@ public void archivedJobDir(
...
-                message,
-                message
-            );
+        try {
+            log.debug("Got a job finished event. Sending email.");
+
+            final JobRequest jobRequest = this.jobSearchService.getJobRequest(jobId);
+            final Job job = this.jobSearchService.getJob(jobId);
+
+            if (org.apache.commons.lang3.StringUtils.isNotBlank(jobRequest.getEmail())) {
```
tgianos: fully qualified usage here why?

PR #223, opened 2016-04-20
Python admin library.

stephen-mw:
I've temporarily moved all genie/ to genie3/ because of namespace collisions in the internal bdp genie libraries. The ```setup.py``` changes reflect that ephemeral change.


stephen-mw:
I'll go ahead and close this for now as I continue to develop the bdp integration. I'd rather not Jenkins kick off with every update.

PR #239, opened 2016-05-11
Add genie rest library

stephen-mw:
@irontable it looks like one file is too large to display inline (look at client.py). The is the large rest client library and I'd appreciate feedback.

I made updates regarding the namespace change to pygenie. See second commit for those changes specifically.


irontable:
@stephen-mw client.py is too large so I can't comment in-line with the code; you're passing in host to the Genie object's __init__(). Does it make more sense to pass in a GenieConf object and pull the default host and version from the conf object?

Issue #162, opened 2015-10-23
3.0 version of Java client developed


tgianos:
Need to re-write the java client from scratch.

- Make it easier to use
- Add ability to get job output via API
- Decouple from Eureka and Ribbon

liviutudor:
Based on the experience with v2.0 client, I think we ought to implement and/or make it easy to extend with the following:

* `getJobOutput(Job)` -- this will probably be a set of methods to allow output retrieveal, either parsed (through a CSV parser) or raw (text)
* `notifyOnJobFinish(...)` -- while the `waitForJobCompletion` is good for those apps which need to block until job is finished , it's also worth implementing an async set of methods where we supply a callback to be notified when the job is finished.
* since genie is used to run Hive, Presto etc queries, and the configuration of the type of query (and/or destination clusters to be run) is achieved via tags, we need to offer an easy way for users to define "tag templates" such that one can easily do something like "run this query as hive in prod", without having to keep track all the time on the caller side of the kind of tags that need to be supplied to Genie to achive this.

liviutudor:
Also, while I wouldn't discard using *RxJava* in the client, I don't think this is really something we should be that concerned about since I don't think users will be looking to shave off a few milliseconds off running a hive query for instance. However, there might be other reasons to justify going with something like that.

tgianos:
I think we should look into building the low levels of the next client based on [Retrofit](http://square.github.io/retrofit/). Seems to have taken as the industry "standard" over something like Feign.

PR #269, opened 2016-06-06
JobClient supports map of filename->inputstreams to create attachments while submitting jobs.

amitsharma1981:
JobClient supports map of filename->inputstreams to create attachments while submitting jobs.


tgianos:
Part of #162

Issue #276, opened 2016-06-11
Throwing checked exceptions on bad input


kilink:
This is mostly about the ergonomics of using Genie, but I find it onerous as a consumer of the API to be forced to catch useless checked exceptions that are thrown for things that are really programmer errors. There are numerous places throughout the codebase where passing a null into a method or an object in some illegal state results in a checked exception like GeniePreconditionException to be thrown.

But checked exceptions are [intended](http://allenlsy.com/NOTES-of-Effective-Java-8-Exceptions/#58.-use-checked-exceptions-for-recoverable-conditions-and-runtime-exceptions-for-programming-errors) for error cases where the caller can reasonably be expected to be able to recover (catch an IOException and optionally retry, etc). But if a programmer is passing bad input, how are they going to recover? It's a bug in the program, not a recoverable error condition.

Here is an example in [ApplicationClient](https://github.com/Netflix/genie/blob/f20e072a75165009ba1617b53880a40c2b88d92e/genie-client/src/main/java/com/netflix/genie/client/ApplicationClient.java#L218).  Here's what calling such an API feels like when you are forced to catch [useless exceptions](http://allenlsy.com/NOTES-of-Effective-Java-8-Exceptions/#59.-avoid-unnecessary-use-of-checked-exceptions):

```java
try {
    String applicationId = null;
    applicationClient.updateApplication(applicationId, application);
} catch (GenieClientException exc) {
    // oops, shouldn't have passed null to updateApplication, let me fix it and try again
   applicationId = "realApplicationId";
   // Call updateApplication again?
}
```

If we can agree that using checked exceptions for such cases harms the ergonomics of the API, and such precondition checks should throw runtime exceptions, then I would go further and say that custom exceptions for most of these precondition checks are unnecessary. Java already includes [standard runtime exceptions](http://allenlsy.com/NOTES-of-Effective-Java-8-Exceptions/#60.-favor-the-use-of-standard-exceptions) for dealing with such situations, such as IllegalArgumentException, NullPointerException, and IllegalStateException.

Is this something that we can address in 3.0?

kilink:
Fixed by #282.

Issue #275, opened 2016-06-11
JobClient.waitForCompletion throws ambiguous IOException on interruption


kilink:
I opened issue #230 regarding ambiguities around timeout vs interruption in JobClient.waitForCompletion, but unfortunately even on master / 3.0 the issue has not been resolved.

The [current code](https://github.com/Netflix/genie/blob/87a363ac1f1dec69b71a42987f6b0fc904051bb0/genie-client/src/main/java/com/netflix/genie/client/JobClient.java#L437) now throws IOException when the call to Thread.sleep is interrupted, without ever restoring the interrupt flag, so the caller is unable to correctly handle it. I also can't just catch the IOException and assume its an interruption, since the call to getJobStatus can also throw an IOException.

The method's throws clause was changed to include InterruptedException, so why not just let the exception propagate without trying to handle it?

kilink:
Fixed by #282.

PR #282, opened 2016-06-15
Refactored client exceptions and added logic for job dir clean up after job finishes.

amitsharma1981:
- The job dependencies are now deleted (configurable) before the archive is created. The deletion had a bug which is now fixed by deleting each application dependencies one-by-one

- The clients now throw IllegalArgumentException in case of missing params

- The wait method in JobClient now does not gobble up the the InterruptedException from the thread sleep method but lets it propagate up to the caller to handle it.

- All non 2xx response in the clients now throw GenieClientExceptions and IOExceptions are thrown for Network and other IO problems.

### DIFF
--- a/genie-client/src/main/java/com/netflix/genie/client/JobClient.java
+++ b/genie-client/src/main/java/com/netflix/genie/client/JobClient.java
@@ -129,14 +129,14 @@ public class JobClient extends BaseGenieClient {
      *
      * @return jobId The id of the job submitted.
      *
-     * @throws GenieClientException For any other error.
-     * @throws IOException If the response received is not 2xx.
+     * @throws GenieClientException If the response recieved is not 2xx.
+     * @throws IOException For Network and other IO issues.
      */
     public String submitJob(
         final JobRequest jobRequest
     ) throws IOException, GenieClientException {
         if (jobRequest == null) {
-            throw new GenieClientException("Job Request cannot be null.");
+            throw new IllegalArgumentException("Job Request cannot be null.");
         }
         return getIdFromLocation(jobService.submitJob(jobRequest).execute().headers().get("location"));
     }
@@ -149,15 +149,15 @@ public class JobClient extends BaseGenieClient {
      *
      * @return jobId The id of the job submitted.
      *
-     * @throws GenieClientException       For any other error.
-     * @throws IOException If the response received is not 2xx.
+     * @throws GenieClientException If the response recieved is not 2xx.
+     * @throws IOException For Network and other IO issues.
      */
     public String submitJobWithAttachments(
         final JobRequest jobRequest,
         final Map<String, InputStream> attachments
     ) throws IOException, GenieClientException {
         if (jobRequest == null) {
-            throw new GenieClientException("Job Request cannot be null.");
+            throw new IllegalArgumentException("Job Request cannot be null.");
         }
 
         final MediaType attachmentMediaType = MediaType.parse(APPLICATION_OCTET_STREAM);
@@ -194,8 +194,9 @@ public class JobClient extends BaseGenieClient {
      * Method to get a list of all the jobs.
      *
      * @return A list of jobs.
-     * @throws GenieClientException       For any other error.
-     * @throws IOException If the response received is not 2xx.
+     *
+     * @throws GenieClientException If the response recieved is not 2xx.
+     * @throws IOException For Network and other IO issues.
      */
     public List<JobSearchResult> getJobs() throws IOException, GenieClientException {
         return this.getJobs(null, null, null, null, null, null, null, null, null, null, null, null, null);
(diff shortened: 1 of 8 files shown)


genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -440,14 +462,15 @@ public void killJob(final String jobId) throws IOException {
      *                     GenieClientException will be thrown
      * @param pollTime     the time to sleep between polling for job status
      * @return The job status for the job after completion
-     * @throws GenieClientException       For timeout or other errors.
-     * @throws InterruptedException on thread errors
-     * @throws IOException If the response received is not 2xx.
+     *
+     * @throws InterruptedException on thread errors.
+     * @throws GenieClientException If the response received is not 2xx.
```
skwslide: The description of this exception is not accurate.

genie-client/src/main/java/com/netflix/genie/client/config/GenieNetworkConfiguration.java
```diff
@@ -30,14 +30,19 @@
 @Setter
 public class GenieNetworkConfiguration {
 
+    /**
+     * Default network timeout value if not specified.
+     */
+    public static final long DEFAULT_TIMEOUT = -1;
```
skwslide: Is -1 same as infinite? Infinite is bad (usually).

genie-client/src/main/java/com/netflix/genie/client/interceptor/ResponseMappingInterceptor.java
```diff
@@ -52,8 +53,13 @@ public Response intercept(final Chain chain) throws IOException {
         if (response.isSuccessful()) {
             return response;
         } else {
-            final JsonNode responseBody = mapper.readTree(response.body().string());
-            throw new GenieClientException(response.code(), response.message() + " : " + responseBody.get("message"));
+            try {
+                final JsonNode responseBody = mapper.readTree(response.body().string());
+                throw new GenieClientException(response.code(), response.message()
+                    + " : " + responseBody.get("message"));
+            } catch (JsonMappingException jme) {
```
skwslide: Catching JsonProcessingException here will encompass JsonMappingException and JsonParseException. Or catch both explicitly.

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -461,34 +484,32 @@ public JobStatus waitForCompletion(final String jobId, final long blockTimeout,
...
-                    Thread.sleep(pollTime);
-                } else {
-                    throw new GenieClientException("Timed out waiting for job to finish");
-                }
-            } catch (InterruptedException ie) {
-                throw new IOException("Received interreupted exception from wait thread.");
+            if (System.currentTimeMillis() - startTime < blockTimeout) {
+                Thread.sleep(pollTime);
+            } else {
+                throw new GenieClientException("Timed out waiting for job to finish");
```
skwslide: As already discussed, something like GenieTimeoutException will be better.

genie-client/src/main/java/com/netflix/genie/client/config/GenieNetworkConfiguration.java
amitsharma1981: -1 basically means its not set by user so i will let okhttp use defaults. which is believe is 10000.

skwslide: Got it.

genie-client/src/main/java/com/netflix/genie/client/interceptor/ResponseMappingInterceptor.java
```diff
@@ -57,7 +57,7 @@ public Response intercept(final Chain chain) throws IOException {
                 final JsonNode responseBody = mapper.readTree(response.body().string());
                 throw new GenieClientException(response.code(), response.message()
                     + " : " + responseBody.get("message"));
-            } catch (JsonMappingException jme) {
+            } catch (JsonProcessingException jpe) {
                 throw new GenieClientException(response.code(), response.message() + response.body().toString());
```
skwslide: I suspect calling response.body() a second time returns nothing.

amitsharma1981: Yea thats correct. I am going to cache it for reuse.

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -461,34 +485,33 @@ public JobStatus waitForCompletion(final String jobId, final long blockTimeout,
...
-     * @param id           the Genie job ID to wait for completion.
+     * @param jobId           the Genie job ID to wait for completion.
      * @param blockTimeout the time to block for (in ms), after which a
      *                     GenieClientException will be thrown.
      * @return The job status for the job after completion.
-     * @throws GenieClientException       For timeout or other errors.
-     * @throws InterruptedException on timeout/thread errors.
-     * @throws IOException If the response recieved is not 2xx.
+     *
+     * @throws InterruptedException on thread errors.
```
kilink: Nitpick: I would probably reword this, as an InterruptedException does not necessarily indicate a "thread error". Here's what the Javadoc for Future.get says: "if the current thread was interrupted while waiting", which seems like a more accurate description.

I think the same Javadoc comment for InterruptedException is repeated elsewhere, so if we decide to change it, we should fix it everywhere.

kilink:
LGTM in terms of addressing the issues I opened in #275 and #276. Thanks!

PR #283, opened 2016-06-16
Disabled ability to turn on automatic retries in okhttp and refactored client exceptions.

amitsharma1981:


genie-client/src/main/java/com/netflix/genie/client/exceptions/GenieClientException.java
```diff
@@ -39,7 +39,7 @@
      * @param msg human readable message
      */
     public GenieClientException(final int errorCode, final String msg) {
-        super(msg);
+        super(String.valueOf(errorCode) + msg);
```
skwslide: Something like a colon is needed to separate the error code and the message.

PR #312, opened 2016-07-14
fix setting session properties for PrestoJob in Python client

irontable:


irontable:
This is a bad pull request.

PR #313, opened 2016-07-14
Added health indicator for physical memory used on the instance. 

ajoymajumdar:
Also marks the service out-of-service if it crosses the threshold which is a configuration property - genie.threshold.used-physical-memory-percent. genie.threshold.used-physical-memory-percent should be a percentage of used memory. By default, it is 90. Health indicator displays the free and total physical memory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,8 +116,8 @@ subprojects {
         testRuntime("javax.el:javax.el-api")
         testRuntime("org.glassfish:javax.el")
 
-        testCompile 'org.spockframework:spock-core:1.0-groovy-2.4'
-        testCompile 'org.spockframework:spock-guice:1.0-groovy-2.4'
+        testCompile 'org.spockframework:spock-core'
+        testCompile 'org.spockframework:spock-spring'
     }
 
     task unitTests(type: Test, group: 'verification') {
--- a/build.gradle
+++ b/build.gradle
@@ -54,6 +54,7 @@ subprojects {
     apply plugin: 'nebula.optional-base'
     apply plugin: 'nebula.provided-base'
     apply plugin: 'com.gorylenko.gradle-git-properties'
+    apply plugin: 'groovy'
 
     group = "com.netflix.${githubProjectName}"
 
@@ -114,6 +115,9 @@ subprojects {
 
         testRuntime("javax.el:javax.el-api")
         testRuntime("org.glassfish:javax.el")
+
+        testCompile 'org.spockframework:spock-core:1.0-groovy-2.4'
+        testCompile 'org.spockframework:spock-guice:1.0-groovy-2.4'
     }
 
     task unitTests(type: Test, group: 'verification') {
--- a/genie-web/src/main/java/com/netflix/genie/web/configs/ServicesConfig.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/configs/ServicesConfig.java
@@ -345,7 +345,7 @@ public class ServicesConfig {
     }
 
     /**
-     * Get the .
+     * Returns the operating system MX bean.
      *
      * @return The operating system MX bean.
      */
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -114,6 +115,9 @@ subprojects {
 
         testRuntime("javax.el:javax.el-api")
         testRuntime("org.glassfish:javax.el")
+
+        testCompile 'org.spockframework:spock-core:1.0-groovy-2.4'
```
tgianos: You don't need to include these versions here as it is already included in the spring platform BOM we're using. We default to that as much as possible that way when we upgrade the BOM version the entire system upgrades compatible libraries. See [here](http://docs.spring.io/platform/docs/2.0.6.RELEASE/reference/htmlsingle/).

genie-web/src/main/java/com/netflix/genie/web/configs/ServicesConfig.java
```diff
@@ -341,4 +343,14 @@ public JobCoordinatorService jobCoordinatorService(
             eventPublisher
         );
     }
+
+    /**
+     * Get the .
```
tgianos: Doesn't look like you finished this comment.

genie-web/src/main/java/com/netflix/genie/web/health/MemoryHealthIndicator.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param maxUsedPhysicalMemoryPercentage The maximum physical memory threshold
+     * @param operatingSystemMXBean MX bean for operating system
+     */
+    @Autowired
+    public MemoryHealthIndicator(
+            @Value("${genie.threshold.used-physical-memory-percent:90}")
```
tgianos: We haven't used dashes in any of our property names. I would switch this to camel case for consistency. We might also want to group them all under a ```genie.health``` level like this could be ```genie.health.memory.usedPhysicalMemoryPercent```

genie-web/src/main/java/com/netflix/genie/web/health/MemoryHealthIndicator.java
```diff
@@ -0,0 +1,64 @@
+package com.netflix.genie.web.health;
```
tgianos: Can you add the copyright header?

genie-web/src/test/groovy/com/netflix/genie/web/health/MemoryHealthIndicatorSpec.groovy
```diff
@@ -0,0 +1,35 @@
+package com.netflix.genie.web.health
```
tgianos: Copyright header here too

genie-web/src/test/groovy/com/netflix/genie/web/health/MemoryHealthIndicatorSpec.groovy
```diff
@@ -0,0 +1,35 @@
...
+class MemoryHealthIndicatorSpec extends Specification{
+    OperatingSystemMXBean operatingSystemMXBean
+    MemoryHealthIndicator memoryHealthIndicator;
+
+    def setup(){
+        operatingSystemMXBean = Mock(OperatingSystemMXBean)
+        memoryHealthIndicator = new MemoryHealthIndicator(80, operatingSystemMXBean)
+    }
+
+    def testHealth(){
```
tgianos: I like this you'll have to explain to me in person what we can do with this

genie-web/src/test/resources/application-integration.yml
```diff
@@ -22,6 +22,8 @@ genie:
       location: "file:///tmp/"
     archive:
       location: /archive/location
+  threshold:
+    used-physical-memory-percent: 100
```
tgianos: Can you add a default value to the ```src/main/resources/application.yml```? We try to put default values in there so that users know most of the properties that are available.

genie-web/src/test/resources/application-integration.yml
```diff
@@ -17,13 +17,14 @@
 ##
 
 genie:
+  health:
+    memory:
+      usedPhysicalMemoryPercent: 100
```
tgianos: 100 seems a bit high no? it'll never really go down in that case

ajoymajumdar: Instance memory is indeterministic where these integration tests are run. So setting 100 will be safe to have all the integration tests run successfully always.

PR #319, opened 2016-07-20
Synchronize Job Coordination

tgianos:
This PR synchronizes the entire Job Coordination code to make sure only one job per node can enter that block at a time and all metrics/job counts are consistent before another job is allowed.

This may slow down performance somewhat but the consistency tradeoff should be worth it. Will observe performance change in testing under load.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/services/JobCoordinatorService.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/JobCoordinatorService.java
@@ -159,7 +159,9 @@ public class JobCoordinatorService {
                 .withStatus(JobStatus.FAILED)
                 .withStatusMsg("Unable to run job due to host being too busy during request.");
             this.jobPersistenceService.createJob(jobBuilder.build());
-            throw new GenieServerUnavailableException("Reached max running jobs on this host. Unable to run job.");
+            throw new GenieServerUnavailableException(
+                "Running max running jobs (" + this.maxRunningJobs + ") on this host. Unable to run job."
+            );
         }
     }
 
--- a/genie-core/src/main/java/com/netflix/genie/core/services/JobCoordinatorService.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/JobCoordinatorService.java
@@ -136,32 +136,40 @@ public class JobCoordinatorService {
             .withId(jobRequest.getId())
             .withTags(jobRequest.getTags());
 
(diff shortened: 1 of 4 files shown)


tgianos:
@ajoymajumdar if you have time give this a quick peer review to see if you see any potential issues.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCountServiceImpl.java
```diff
@@ -49,7 +49,7 @@ public JobCountServiceImpl(@NotNull final JobSearchService jobSearchService, @No
      * {@inheritDoc}
      */
     @Override
-    public int getNumJobs() {
+    public synchronized int getNumJobs() {
```
ajoymajumdar: Why do you need to synchronize this method?

genie-core/src/main/java/com/netflix/genie/core/services/JobCoordinatorService.java
```diff
@@ -136,30 +136,40 @@ public String coordinateJob(
...
-        if (this.canRunJob()) {
-            jobBuilder
-                .withStatus(JobStatus.INIT)
-                .withStatusMsg("Job Accepted and in initialization phase.");
-            // TODO: if this throws exception the job will never be marked failed
-            this.jobPersistenceService.createJob(jobBuilder.build());
-            try {
-                final Future<?> task
-                    = this.taskExecutor.submit(new JobLauncher(this.jobSubmitterService, jobRequest, this.registry));
+        synchronized (this) {
```
ajoymajumdar: This fix looks correct to really coordinate the max number of jobs on an instance. Instead of synchronized and getting the count, cant we have a thread pool with a queue capacity of max jobs. If the number goes up, the pool will reject it and you can either return back with a failure or redirect to another instance.

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitoringCoordinator.java
```diff
@@ -201,7 +201,7 @@ public synchronized void onJobFinished(final JobFinishedEvent event) {
      *
      * @return the number of jobs currently running on this node
      */
-    public int getNumJobs() {
+    public synchronized int getNumJobs() {
```
ajoymajumdar: Why do you need to synchronize this?

PR #322, opened 2016-07-20
Latency Metrics

tgianos:
This PR adds a significant number of metrics tracking the latency of various background tasks in the Genie system.

In particular it focusses on how long it takes to download and upload files to S3 as well as how long it takes to every phase of the job workflow.


genie-core/src/main/java/com/netflix/genie/core/jobs/JobLauncher.java
```diff
@@ -62,34 +64,38 @@ public JobLauncher(
...
-            } else if (e instanceof GenieTimeoutException) {
-                this.registry.counter(MetricsConstants.GENIE_EXCEPTIONS_TIMEOUT_RATE).increment();
-            } else {
-                this.registry.counter(MetricsConstants.GENIE_EXCEPTIONS_OTHER_RATE).increment();
+        this.submitTimer.record(() -> {
+                try {
+                    this.jobSubmitterService.submitJob(this.jobRequest);
+                } catch (final GenieException e) {
+                    log.error("Unable to submit job due to exception: {}", e.getMessage(), e);
+                    if (e instanceof GenieBadRequestException) {
```
ajoymajumdar: I would suggest adding one metric with tag as exception type.

tgianos: As in the String representation of the class name? That could be cleaner. We'll try to add it after I get through some more functional things.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ApplicationTask.java
```diff
@@ -41,6 +44,17 @@
 @Slf4j
 public class ApplicationTask extends GenieBaseTask {
 
+    private Timer timer;
```
ajoymajumdar: Create an @Aspect class to proxy all GenieBaseTask to log and monitor time.

PR #330, opened 2016-07-27
Enabled caching the files for cluster, application and command tasks.

ajoymajumdar:
Changes:

- Enabled caching the files that only re-downloads if the modified time of the downloaded file is old.
- Using a factory bean to get the right file system service(FileTransfer)
- Modified to pass in the FileTransferService to the GenieBaseTask constructor instead of passing it as part of the context.
- Removed the usage of GenieNodeStatisticsImpl
- Added option StandardCopyOption.REPLACE_EXISTING when copying files using local file system


build.gradle
```diff
@@ -118,6 +118,7 @@ subprojects {
 
         testCompile 'org.spockframework:spock-core'
         testCompile 'org.spockframework:spock-spring'
+        testCompile 'cglib:cglib-nodep:2.2.2'
```
tgianos: If this dependency isn't available as part of the spring platform bom please move it to gradle.properties so we can try to keep it consistent across project if we end up using it in multiple places

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ClusterTask.java
```diff
@@ -44,14 +44,18 @@
...
+    private final GenieFileTransferService fts;
 
     /**
      * Constructor.
      *
      * @param registry The metrics registry to use
+     * @param fts File transfer service
      */
-    public ClusterTask(@NotNull final Registry registry) {
+    public ClusterTask(@NotNull final Registry registry,
```
tgianos: We seem to be repeating a lot of this signature in the constructors I'd move it into an abstract base class for reuse

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobTask.java
```diff
@@ -44,20 +44,27 @@
...
      */
     public JobTask(
-        @NotNull final AttachmentService attachmentService,
-        @NotNull final Registry registry
+            @NotNull
+            final AttachmentService attachmentService,
+            @NotNull
+            final Registry registry,
+            @NotNull
+            final GenieFileTransferService fts
```
tgianos: I wonder if we can somehow combine the attachment service and the cache into one as they're pretty similar. Not necessarily required for this feature but a passing thought

genie-core/src/main/java/com/netflix/genie/core/services/FileTransferFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.netflix.genie.core.services;
+
+/**
+ * Factory for FileTransfer implementation based on the scheme.
+ *
+ * Created by amajumdar on 7/21/16.
```
tgianos: Annoying nitpick but we're trying to add ```@since``` to the classes so we can know when things were introduced and backtrace. Will help as we deprecate things going forward.

genie-core/src/main/java/com/netflix/genie/core/services/impl/CacheGenieFileTransferService.java
```diff
@@ -0,0 +1,152 @@
...
+ * Created by amajumdar on 7/22/16.
+ */
+@Slf4j
+public class CacheGenieFileTransferService extends GenieFileTransferService {
+    //File cache location
+    private final String baseCacheLocation;
+    //File transfer service to get/put files on a local system
+    private final FileTransfer localFileTransfer;
+    //File cache
+    private final LoadingCache<String, File> fileCache = CacheBuilder.newBuilder()
```
tgianos: Nice I didn't know such a thing existed. We can possibly use this for on node caching of security tokens with a TTL if we want to get around the current bottleneck quickly.

genie-core/src/main/java/com/netflix/genie/core/services/impl/CacheGenieFileTransferService.java
```diff
@@ -0,0 +1,152 @@
...
+    public CacheGenieFileTransferService(
+            @NotNull final FileTransferFactory fileTransferFactory,
+            @NotNull final String baseCacheLocation,
+            @NotNull final FileTransfer localFileTransfer,
+            @NotNull final Registry registry) throws GenieException {
+        super(fileTransferFactory);
+        this.baseCacheLocation = baseCacheLocation;
+        createDirectories(baseCacheLocation);
+        this.localFileTransfer = localFileTransfer;
+        registry.gauge("genie.jobs.file.cache.hitRate", fileCache,
```
tgianos: Since you're trying to get a rate wouldn't you rather using a counter than a gauge? Cause the gauge will just go up and down rather than capture how often it's happening.

genie-core/src/main/java/com/netflix/genie/core/services/impl/CacheGenieFileTransferService.java
```diff
@@ -0,0 +1,152 @@
...
+                (ToDoubleFunction<LoadingCache<String, File>>) value -> value.stats().hitRate());
+        registry.gauge("genie.jobs.file.cache.missRate", fileCache,
```
tgianos: Since these cache's aren't used by only job files (in fact we don't use them for any job files really just clusters, commands and apps right?) I would rename metric to more closely match the ones I did for s3: genie.files.s3.download.timer... just under ```genie.files.cache.whatever```

genie-core/src/main/java/com/netflix/genie/core/services/impl/CacheGenieFileTransferService.java
```diff
@@ -0,0 +1,152 @@
...
+import java.nio.file.Path;
+import java.util.UUID;
+import java.util.function.ToDoubleFunction;
+
+/**
+ * Caches the downloaded file from the remote location.
+ * Created by amajumdar on 7/22/16.
+ */
+@Slf4j
+public class CacheGenieFileTransferService extends GenieFileTransferService {
```
tgianos: Isn't this missing a put file definition?

genie-core/src/main/java/com/netflix/genie/core/services/impl/GenieFileTransferService.java
```diff
@@ -36,19 +38,19 @@
 @Slf4j
 public class GenieFileTransferService {
 
-    private final List<FileTransfer> fileTransferList;
+    protected final FileTransferFactory fileTransferFactory;
```
tgianos: why protected instead of private?

genie-core/src/main/java/com/netflix/genie/core/services/impl/S3FileTransferImpl.java
```diff
@@ -63,6 +65,7 @@ public S3FileTransferImpl(
         this.s3Client = amazonS3Client;
         this.downloadTimer = registry.timer("genie.files.s3.download.timer");
         this.uploadTimer = registry.timer("genie.files.s3.upload.timer");
+        this.getTimer = registry.timer("genie.files.s3.getObjectMetadata.timer");
```
tgianos: please add this to dashboard once this goes live

genie-web/src/main/java/com/netflix/genie/web/configs/aws/AwsS3Config.java
```diff
@@ -100,12 +100,12 @@ public AmazonS3Client genieS3Client(
      * @return An s3 implementation of the FileTransfer interface
      * @throws GenieException if there is any problem
      */
-    @Bean
+    @Bean(name = "s3")
```
tgianos: these names seem kind of risky as they're so short and possibly conflicting with any beans from other boot services. can we use s3:// in there or something? Seems unlikely

PR #337, opened 2016-08-01
Removed @DirtiesContext so as to not restarting the application across tests

ajoymajumdar:
Also modified genie.file.cache.location to correctly point to /tmp.

### DIFF
--- a/genie-web/src/main/resources/application.yml
+++ b/genie-web/src/main/resources/application.yml
@@ -26,7 +26,7 @@ eureka:
 genie:
   file:
     cache:
-      location: "file:///tmp/genie/cache"
+      location: /tmp/genie/cache
   health:
     memory:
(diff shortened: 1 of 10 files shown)


genie-web/src/main/resources/application.yml
```diff
@@ -26,7 +26,7 @@ eureka:
 genie:
   file:
     cache:
-      location: "file:///tmp/genie/cache"
+      location: /tmp/genie/cache
```
tgianos: My concern here is the difference in formatting between this and the jobs directory configuration. The way the jobs dir is used it HAS to have that ```file://``` prefix or some pieces of the code weren't working properly. I believe it was in the bean setup when we try to create the directory if it doesn't exist. I forget exactly just know it's a problem so I would like these to be consistent to avoid problems in user configuration down the road.

PR #341, opened 2016-08-03
JWT support for OAuth2 authentication with Ping Federate

tgianos:
This PR adds the option for a user to enable authentication to APIs via JSON Web Tokens (JWT). This saves the server from sending another request to the Ping Federate server to verify that a token is valid.

All validation is done against the signed JWT locally within the Genie JVM.

The code currently expects three claims to exist in the JWT:
```exp``` the expiration time of the token
```client_id``` which will be the principal (user) within the Spring Security Context.
```scope``` which is a collection of Strings each of which will be converted to ROLE_{scope}. If the scopes are prefixed with ```genie_``` the prefix is stripped off. These roles will become the Authorities within the Spring Security context.

### CODE REVIEW

File: genie-web/src/main/java/com/netflix/genie/web/security/oauth2/pingfederate/PingFederateRemoteTokenServices.java [hunk 1]
```diff
@@ -85,16 +89,29 @@
...
         @NotNull final AccessTokenConverter converter,
         @NotNull final Registry registry
     ) {
+        super();
         this.tokenValidationError = registry.createId("genie.security.oauth2.pingFederate.tokenValidation.error.rate");
         this.authenticationTimer = registry.timer(AUTHENTICATION_TIMER_NAME);
         this.pingFederateAPITimer = registry.timer(API_TIMER_NAME);
-        this.restTemplate = new RestTemplate();
-        this.restTemplate.setErrorHandler(
+        final RestTemplate restTemplate = new RestTemplate();
```
ajoymajumdar: Set read and connect timeout.

PR #342, opened 2016-08-03
Using RestTemplate instead of direct htppclient so as to cleanly close the http connection

ajoymajumdar:
Modified to use RestTemplate to make remotes calls to other nodes. Also added logs if the remote calls fail.


genie-web/src/main/java/com/netflix/genie/web/configs/MvcConfig.java
```diff
@@ -86,14 +86,22 @@ public String hostName() throws UnknownHostException {
...
-     * Get an HttpClient for calling between Genie nodes.
+     * Get RestTemplate for calling between Genie nodes.
      *
-     * @return The http client to use
+     * @param httpConnectTimeout http connection timeout in milliseconds
+     * @param httpReadTimeout http read timeout in milliseconds
+     * @return The rest template to use
      */
     @Bean
     @ConditionalOnMissingBean
```
tgianos: Did you verify that no where else in the spring boot auto configuration stack another rest template wasn't being created with this name that would cause yours to not be instantiated due to this conditional? I ask cause it seems risky as we have specific things in here we want to be setup and we can't guarantee another instance has it. May also want to use qualifiers where it is wired in.

genie-web/src/main/java/com/netflix/genie/web/configs/aws/AwsMvcConfig.java
```diff
@@ -44,33 +40,27 @@
...
-            final String hostname = EntityUtils.toString(publicHostnameResponse.getEntity(), UTF_8);
-            log.debug("AWS Public Hostname: {}", hostname);
-            return hostname;
+    public String hostName(final RestTemplate restTemplate) throws IOException {
+        String result = null;
+        try {
+            result = restTemplate.getForObject(publicHostNameGet, String.class);
+            log.debug("AWS Public Hostname: {}", result);
+        } catch (Exception e) {
+            result = restTemplate.getForObject(localIPV4HostNameGet, String.class);
```
tgianos: I assume either of these calls will throw an IOException if they can't actually reach the host and startup will fail?

genie-web/src/main/java/com/netflix/genie/web/controllers/JobRestController.java
```diff
@@ -227,27 +227,27 @@ public JobRestController(
...
         return this.handleSubmitJob(jobRequest, attachments, clientHost, userAgent, httpServletRequest);
     }
 
     private ResponseEntity<Void> handleSubmitJob(
-        final JobRequest jobRequest,
-        final MultipartFile[] attachments,
-        final String clientHost,
-        final String userAgent,
-        final HttpServletRequest httpServletRequest
+            final JobRequest jobRequest,
```
tgianos: It looks like your IDE is formatting things differently. Is it not picking up the .editorconfig file?

PR #344, opened 2016-08-04
Fixed issue where the job output request was responding back with a 401 error

ajoymajumdar:
This error only happened the first time the request is made with a new session or stale one. Issue was that while forwarding the request to a different node, the session cookie was not forwarded.


genie-web/src/main/java/com/netflix/genie/web/controllers/JobRestController.java
```diff
@@ -110,6 +111,7 @@
 public class JobRestController {
 
     private static final String FORWARDED_FOR_HEADER = "X-Forwarded-For";
+    private static final String NAME_HEADER_COOKIE = "cookie";
```
tgianos: COOKIE_HEADER_NAME seems clearer to me

tgianos:
Are there any test modifications for this?

PR #345, opened 2016-08-05
Handle jobs stuck in INIT state during error conditions

tgianos:
PR makes sure jobs that are in the INIT state if a node crashes and comes back are marked failed. Also makes sure if a node completely falls out of the cluster and the jobs are lost they are marked failed by the ClusterCheckerTask along with any jobs thought to be running on that node.

### DIFF
--- a/genie-common/src/main/java/com/netflix/genie/common/dto/JobStatus.java
+++ b/genie-common/src/main/java/com/netflix/genie/common/dto/JobStatus.java
@@ -34,27 +34,27 @@ public enum JobStatus {
     /**
      * Job has been initialized, but not running yet.
      */
-    INIT,
+    INIT(true),
     /**
      * Job is now running.
      */
-    RUNNING,
+    RUNNING(true),
     /**
      * Job has finished executing, and is successful.
      */
-    SUCCEEDED,
+    SUCCEEDED(false),
     /**
      * Job has been killed.
      */
-    KILLED,
+    KILLED(false),
     /**
      * Job failed.
      */
-    FAILED,
+    FAILED(false),
     /**
      * Job cannot be run due to invalid criteria.
      */
-    INVALID;
+    INVALID(false);
 
     private static final Set<JobStatus> ACTIVE_STATUSES = Collections.unmodifiableSet(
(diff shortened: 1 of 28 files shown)


genie-common/src/main/java/com/netflix/genie/common/dto/JobStatus.java
```diff
@@ -78,7 +89,7 @@ public static JobStatus parse(final String value) throws GeniePreconditionExcept
      * @return True if the job is still actively processing in some manner
      */
     public boolean isActive() {
-        return this == INIT || this == RUNNING;
```
ajoymajumdar: Enum should know that it is active or not during initialization. Create a constructor JobStatus(boolean active).

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -193,28 +194,58 @@ public JpaJobSearchServiceImpl(
...
+        final Root<JobEntity> root = query.from(JobEntity.class);
+        final Join<JobEntity, JobExecutionEntity> executions = root.join(JobEntity_.execution);
+
+        final Set<Predicate> statusPredicates = JobStatus
+            .getActiveStatuses()
+            .stream()
+            .map(status -> cb.equal(root.get(JobEntity_.status), status))
+            .collect(Collectors.toSet());
+
+        final Predicate statusPredicate = cb.or(statusPredicates.toArray(new Predicate[statusPredicates.size()]));
```
ajoymajumdar: Can we use an in clause?

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -193,28 +194,58 @@ public JpaJobSearchServiceImpl(
...
+
+        final Predicate hostPredicate = cb.equal(executions.get(JobExecutionEntity_.hostName), hostName);
+
+        query
+            .distinct(true)
```
ajoymajumdar: why distinct?

PR #346, opened 2016-08-08
Add GENIE_VERSION as environment variable for jobs

tgianos:


tgianos:
@danielcweeks note I changed it to GENIE_VERSION instead of genie.version to be a more proper environment variable

ajoymajumdar:
Looks good.

PR #352, opened 2016-08-11
Refactored ClusterCheckerTask class

ajoymajumdar:
- Refactored ClusterCheckerTask class
- Mark a node unhealthy if node status is up for all health endpoints excluding a few selected ones like genie,memory.
- Added more error checks.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
@@ -96,7 +96,7 @@ public class ClusterCheckerTask extends LeadershipTask {
         @NotNull final JobPersistenceService jobPersistenceService,
         @Qualifier("genieRestTemplate") @NotNull final RestTemplate restTemplate,
         @NotNull final ManagementServerProperties managementServerProperties,
-        @Value("${genie.tasks.clusterChecker.healthIndicatorsToIgnore:memory,genie}")
+        @Value("${genie.tasks.clusterChecker.healthIndicatorsToIgnore:memory,genie,discoveryComposite}")
         final String healthIndicatorsToIgnore,
         @NotNull final Registry registry
     ) {
@@ -199,8 +199,7 @@ public class ClusterCheckerTask extends LeadershipTask {
                 for (Map.Entry<String, Object> responseEntry : responseMap.entrySet()) {
                     if (responseEntry.getValue() instanceof Map
                         && !healthIndicatorsToIgnore.contains(responseEntry.getKey())
-                        && Status.OUT_OF_SERVICE.getCode()
-                        .equals(((Map) responseEntry.getValue()).get(PROPERTY_STATUS))) {
+                        && !Status.UP.getCode().equals(((Map) responseEntry.getValue()).get(PROPERTY_STATUS))) {
                         result = false;
                         break;
                     }
--- a/genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
@@ -17,6 +17,10 @@
  */
 package com.netflix.genie.web.tasks.leader;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.google.common.base.Splitter;
+import com.google.common.base.Throwables;
 import com.netflix.genie.common.dto.Job;
 import com.netflix.genie.common.dto.JobExecution;
 import com.netflix.genie.common.dto.JobStatus;
@@ -30,16 +34,18 @@ import com.netflix.spectator.api.Registry;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.actuate.autoconfigure.ManagementServerProperties;
+import org.springframework.boot.actuate.health.Status;
 import org.springframework.stereotype.Component;
+import org.springframework.web.client.HttpStatusCodeException;
 import org.springframework.web.client.RestTemplate;
 
 import javax.validation.constraints.NotNull;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * A task which checks to see if this leader node can communicate with all other nodes in the cluster. If it can't
@@ -53,6 +59,7 @@ import java.util.stream.Collectors;
 @Component
 @Slf4j
 public class ClusterCheckerTask extends LeadershipTask {
+    private static final String PROPERTY_STATUS = "status";
 
     private final String hostName;
     private final ClusterCheckerProperties properties;
@@ -61,6 +68,8 @@ public class ClusterCheckerTask extends LeadershipTask {
     private final RestTemplate restTemplate;
     private final String scheme;
     private final String healthEndpoint;
+    private final ObjectMapper mapper = new ObjectMapper();
+    private final List<String> ignoreHealthEndPoints;
 
     private final Map<String, Integer> errorCounts = new HashMap<>();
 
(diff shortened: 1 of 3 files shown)


genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -87,6 +97,7 @@ public ClusterCheckerTask(
         @NotNull final JobPersistenceService jobPersistenceService,
         @Qualifier("genieRestTemplate") @NotNull final RestTemplate restTemplate,
         @NotNull final ManagementServerProperties managementServerProperties,
+        @Value("${genie.tasks.clusterChecker.ignoreHealthEndPoints:memory,genie}") final String ignoreHealthEndPoints,
```
tgianos: There is only one endpoint I would go with healthIndicatorsToIgnore

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
     @Override
     public void run() {
         log.info("Checking for cluster node health...");
-        final Set<String> badNodes = new HashSet<>();
         this.jobSearchService.getAllHostsWithActiveJobs()
             .stream()
             .filter(host -> !this.hostName.equals(host))
```
tgianos: I wonder about this cause we basically don't check the health of the "leader node" but it itself could be unhealthy. We can either leave this in or implement logic in the LeadershipCoordinator which gives up leadership based on the Health state that way it's always on a healthy node.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
...
+            restTemplate.getForObject(this.scheme + host + this.healthEndpoint, String.class);
+        } catch (final HttpStatusCodeException e) {
+            log.error("Failed validating host {}", host, e);
+            try {
+                final Map<String, Object> responseMap = mapper.readValue(e.getResponseBodyAsByteArray(),
+                    TypeFactory.defaultInstance().constructMapType(Map.class, String.class, Object.class));
+                for (Map.Entry<String, Object> responseEntry : responseMap.entrySet()) {
+                    if (responseEntry.getValue() instanceof Map
+                        && !ignoreHealthEndPoints.contains(responseEntry.getKey())
+                        && Status.OUT_OF_SERVICE.getCode()
```
tgianos: May as well make this a constant

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
...
+        try {
+            restTemplate.getForObject(this.scheme + host + this.healthEndpoint, String.class);
+        } catch (final HttpStatusCodeException e) {
+            log.error("Failed validating host {}", host, e);
+            try {
+                final Map<String, Object> responseMap = mapper.readValue(e.getResponseBodyAsByteArray(),
+                    TypeFactory.defaultInstance().constructMapType(Map.class, String.class, Object.class));
+                for (Map.Entry<String, Object> responseEntry : responseMap.entrySet()) {
+                    if (responseEntry.getValue() instanceof Map
+                        && !ignoreHealthEndPoints.contains(responseEntry.getKey())
```
tgianos: I'd put this condition first so it fails fast before even bothering checking the instance type

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
tgianos: I think you need to check if it's just not UP... basically any state other than that is bad. Looking at the internal Healthchecks we'll probably want to add discovery to the ones to ignore due to it being in UNKNOWN state but we don't care.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
...
+                        && Status.OUT_OF_SERVICE.getCode()
+                        .equals(((Map) responseEntry.getValue()).get(PROPERTY_STATUS))) {
+                        result = false;
```
tgianos: Why not just return here if your result is already false and save yourself the remaining iterations of the loop?

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
...
-            .filter(host -> !badNodes.contains(host))
-            .collect(Collectors.toSet());
-        toRemove.forEach(this.errorCounts::remove);
+    private void validateHostAndUpdateErrorCount(final String host) {
+        final boolean isNodeHealthy = isNodeHealthy(host);
+        //
+        // If node is healthy, remove the entry from the errorCounts.
+        // If node is not healthy, update the entry in errorCounts
+        //
+        if (isNodeHealthy) {
```
tgianos: Just call the method within the if statement, no need to save it to a local variable

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
...
+        this.errorCounts.entrySet().removeIf(entry -> {
+            final String host = entry.getKey();
+            boolean result = true;
+            if (entry.getValue() >= properties.getLostThreshold()) {
+                try {
+                    updateJobsToFailedOnHost(host);
+                } catch (Exception e) {
+                    log.error("Unable to update jobs on host {} due to exception", host, e);
+                    unableToUpdateJobCounter.increment();
+                    result = false;
```
tgianos: just return here

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -109,70 +121,99 @@ public ClusterCheckerTask(
...
                 }
-            );
+            } else {
+                result = false;
```
tgianos: you really don't need an else right? if your statement is true and you don't get exception return true, else false. so just have return false outside of the If block

genie-web/src/test/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTaskUnitTests.java
```diff
@@ -115,7 +119,28 @@ public void canRun() throws IOException, GenieException {
             .thenThrow(new RestClientException("blah"))
             .thenReturn("")
             .thenReturn("")
-            .thenThrow(new RestClientException("blah"))
+            .thenThrow(new HttpServerErrorException(HttpStatus.INTERNAL_SERVER_ERROR, "",
+                ("{\"status\":\"OUT_OF_SERVICE\", \"genie\": { \"status\": \"OUT_OF_SERVICE\"}, "
```
tgianos: Add tests for other status cases other than just OUT_OF_SERVICE... like DOWN or UNKNOWN

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
ajoymajumdar: Do you want to add the logic for revoking the leadership?

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
ajoymajumdar: Always want to have the return at the end. This logic will remove the entry if it returns true.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
ajoymajumdar: Strict check against any element that is not a Map

PR #362, opened 2016-08-18
Fix missing validation on Job input

tgianos:
Bean validation on job requests and other beans wasn't happening due to services not being interfaces so no proxies could be created. Also some of them weren't annotated with validated.


ajoymajumdar:
Good to go.

PR #375, opened 2016-08-24
Make job finished events asynchronous

tgianos:
This PR makes job finished events asynchronous instead of synchronous. This disconnects the thread throwing the events from the handling of the events and freeing up the threads to act on each other. Also should improve performance as all event handlers are independent.


ajoymajumdar:
Looks good.

PR #386, opened 2016-09-06
add session_properties convenience method to PrestoJob in python client

jessicasw:


genie-client/src/main/python/pygenie/jobs/presto.py
```diff
@@ -188,3 +188,25 @@ def session(self, name, value):
         self._set_command_option('--session', name, value)
 
         return self
+
+    @unicodify
+    @add_to_repr('append')
```
irontable: Probably shouldn't call `@add_to_repr()` here because it'll also add to repr when calling `self.session()`.

jessicasw:
Don't need this, closing

PR #402, opened 2016-09-28
Fixes in email, Job.GetApplications, Job cleanup and archiving and Job submission with same user/group.

amitsharma1981:
GetApplications for a job returns an empty list instead of NotFoundException. Emails are more detailed.Archival and Cleanup of job directories only happens if the job runs. Fixed an edge case where job submission failed if user/group specified were the same in the jobrequest.Emails upon job completion are more detailed.

Added start and end time to genie log file for debugging.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
@@ -203,8 +203,9 @@ public class JobKickoffTask extends GenieBaseTask {
         } catch (final IOException ioe) {
             log.debug("User does not exist. Creating it now.");
 
-            // Create the group for the user.
-            if (group != null) {
+            // Create the group for the user if its not the same as the user.
+            if (group != null && !group.equals(user) && StringUtils.isNotBlank(group)) {
+                log.debug("Group and User are different so creating group now.");
                 final CommandLine groupCreateCommandLine = new CommandLine("sudo");
                 groupCreateCommandLine.addArgument("groupadd");
                 groupCreateCommandLine.addArgument(group);
@@ -212,6 +213,7 @@ public class JobKickoffTask extends GenieBaseTask {
                 // We create the group and ignore the error as it will fail if group already exists.
                 // If the failure is due to some other reason, then user creation will fail and we catch that.
                 try {
+                    log.debug("Running command [" + groupCreateCommandLine.toString() + "]");
                     this.executor.execute(groupCreateCommandLine);
                 } catch (IOException ioexception) {
                     log.debug("Group creation threw an error as it might already exist");
@@ -222,7 +224,7 @@ public class JobKickoffTask extends GenieBaseTask {
             userCreateCommandLine.addArgument("useradd");
             userCreateCommandLine.addArgument(user);
 
-            if (StringUtils.isNotBlank(group)) {
+            if (StringUtils.isNotBlank(group) && !group.equals(user)) {
                 userCreateCommandLine.addArgument("-G");
                 userCreateCommandLine.addArgument(group);
             }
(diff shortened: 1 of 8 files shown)


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -203,15 +203,17 @@ protected synchronized void createUser(final String user, final String group) th
         } catch (final IOException ioe) {
             log.debug("User does not exist. Creating it now.");
 
-            // Create the group for the user.
-            if (group != null) {
+            // Create the group for the user if its not the same as the user.
+            if (group != null && !group.equals(user) && StringUtils.isNotBlank(group)) {
```
tgianos: The check for null and the call to StringUtils.isNotBlank are redundant. IsNotBlank checks for null so I would just put that first.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -203,15 +203,17 @@ protected synchronized void createUser(final String user, final String group) th
...
+                log.debug("Group and User are different so creating group now.");
                 final CommandLine groupCreateCommandLine = new CommandLine("sudo");
                 groupCreateCommandLine.addArgument("groupadd");
                 groupCreateCommandLine.addArgument(group);
 
                 // We create the group and ignore the error as it will fail if group already exists.
                 // If the failure is due to some other reason, then user creation will fail and we catch that.
                 try {
+                    log.debug("Running command [" + groupCreateCommandLine.toString() + "]");
```
tgianos: Can you change this to Running command to create group so that it's clearer when we're grepping for it?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -222,14 +224,15 @@ protected synchronized void createUser(final String user, final String group) th
             userCreateCommandLine.addArgument("useradd");
             userCreateCommandLine.addArgument(user);
 
-            if (StringUtils.isNotBlank(group)) {
+            if (StringUtils.isNotBlank(group) && !group.equals(user)) {
```
tgianos: Since this check seems equivalent to the one on line 207 why don't you create a boolean and assign it to the result of these statements so that you don't have to run the method calls twice?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -222,14 +224,15 @@ protected synchronized void createUser(final String user, final String group) th
...
                 userCreateCommandLine.addArgument("-G");
                 userCreateCommandLine.addArgument(group);
             }
 
             userCreateCommandLine.addArgument("-M");
 
             try {
+                log.debug("Running command [" + userCreateCommandLine.toString() + "]");
```
tgianos: More detailed log message for easier searching

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -268,29 +268,31 @@ private Void updateJob(final Job job, final JobFinishedEvent event, final Map<St
...
-
-                    // The process group should not exist and the above code should always throw and exception.
-                    // If it does not then the bash script is not cleaning up stuff well during kills
-                    // or the script is done but child processes are still remaining. This metric tracks all that.
-                    processGroupCleanupFailureRate.increment();
-                } catch (final Exception e) {
-                    log.debug("Received expected exception. Ignoring.");
-                }
-            });
+            if (!this.jobSearchService.getJobStatus(jobId).equals(JobStatus.INVALID)) {
```
tgianos: Have we gotten the job status anywhere else that we can pass in and save from making yet another DB call here?

amitsharma1981: Each method here is completely independent and the previous one's may or may not run. It's better to fetch it in the method itself.

PR #403, opened 2016-09-28
Adding thymeleaf back in

tgianos:
Needed to load index.html currently.


tgianos:
@cabhishek this is a temp fix to get us back up and running but you should find a way to get rid of the thymeleaf dependency as we're not actually using any of its functionality and I doubt we need it currently. It's just jar bloat.

Issue #395, opened 2016-09-22
Retrying on connection failures


kilink:
In BaseGenieClient, the underlying OkHttpClient instance is configured to [not retry on connection failures](https://github.com/Netflix/genie/blob/5dc76c8d89bcd9c64f38c85c6be35a8820a33d69/genie-client/src/main/java/com/netflix/genie/client/BaseGenieClient.java#L71). According to @amitsharma1981, the reasoning was to avoid retrying on non-idempotent requests such as POSTs.

However, if you look at the outcome of this OkHttp [issue](https://github.com/square/okhttp/issues/2394), you'll see connections are only retried if no bytes have been written to the connection (see these [test cases](https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/okhttp3/CallTest.java#L775-L826) that demonstrate this behavior).

I suggest not configuring the OkHttpClient in this manner; we have run into issues where long running processes that poll Genie (using waitForCompletion) inevitably fail due to a stale connection.

PR #405, opened 2016-09-30
Make retrying on connection failures on by default

tgianos:
Can be disabled by setting a field in the GenieNetworkConfiguration. Upgrade retrofit and okhttp libraries to get all bug fixes


tgianos:
Should address #395. @skwslide you may want to be aware of this change for your code once you're back.

PR #406, opened 2016-10-09
Determine health based on memory used out of configured maximum

tgianos:
Previously Genie determined the health of a given node based total number of running jobs vs a configured maximum. This assumed the admins had configured the system with a maximum that somehow represented the available memory (e.g. assume jobs used 1.5 GB on average and divide available memory by that amount to get a safe maximum number of jobs). This works at small scale but as the types of jobs that can be run become more divers and memory intensive on the client side it becomes less and less safe to make this assumption.

This PR makes it so the admin configure a maximum amount of memory available to a Genie node as well as the default and maximum amount of memory a job can use. Admins can also set the amount of memory (all in MB) that specific commands will use by default (e.g. presto could use 512 MB while spark might use 2 GB). Users can also override the default memory by sending in a requested amount of memory with their job request. Once the memory is resolved if enough is available on the node the job is accepted and the memory is added to the amount of used memory.

The Genie node reports unhealthy status once the amount of available memory drops below the amount required to run one default job.

Genie will also export a new environment variable for job clients to use (GENIE_JOB_MEMORY) which best practices would say all commands should configure this as their Xmx and Xms so that memory used is consistent with what Genie thinks is being used.


genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -96,104 +133,166 @@ public JobCoordinatorServiceImpl(
...
-                    // Tell the system a new job has been scheduled so any actions can be taken
-                    log.info("Publishing job scheduled event for job {}", jobId);
-                    this.eventPublisher.publishEvent(new JobScheduledEvent(jobId, task, this));
-                } catch (final TaskRejectedException e) {
-                    final String errorMsg = "Unable to launch job due to exception: " + e.getMessage();
-                    this.jobPersistenceService.updateJobStatus(jobId, JobStatus.FAILED, errorMsg);
-                    throw new GenieServerException(errorMsg, e);
+                        // Tell the system a new job has been scheduled so any actions can be taken
+                        log.info("Publishing job scheduled event for job {}", jobId);
+                        this.eventPublisher.publishEvent(new JobScheduledEvent(jobId, task, memory, this));
```
ajoymajumdar: Do we want to publish the event from the JobLauncher?

tgianos: As discussed we wouldn't then have the handle to the Future returned from the scheduler that we can use to cancel the task. For now leaving as is.

PR #407, opened 2016-10-10
Optimized the logic to get the job status.

ajoymajumdar:
1. Optimized the logic to get the job status.
2. Added property to configure S3 retries. By default, the retry setting is 5.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
@@ -52,6 +52,7 @@ import org.springframework.data.domain.Sort;
 import org.springframework.transaction.annotation.Transactional;
 
 import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
 import javax.persistence.PersistenceContext;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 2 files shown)


genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -270,14 +271,14 @@ public Job getJob(
...
         log.debug("Called with id {}", id);
-        if (this.jobRepository.exists(id)) {
-            final CriteriaBuilder cb = this.entityManager.getCriteriaBuilder();
-            final CriteriaQuery<JobStatus> query = cb.createQuery(JobStatus.class);
-            final Root<JobEntity> root = query.from(JobEntity.class);
-            query.select(root.get(JobEntity_.status));
-            query.where(cb.equal(root.get(JobEntity_.id), id));
-            return this.entityManager.createQuery(query).getSingleResult();
-        } else {
+        final CriteriaBuilder cb = entityManager.getCriteriaBuilder();
```
tgianos: Since this is much less dynamic than some of our other search queries did you see if theres a way to use a repository or named query which is more static for pulling back the same info rather than building it up every time? Can we make this static somehow i guess is the core question

genie-web/src/main/java/com/netflix/genie/web/configs/aws/AwsS3Config.java
```diff
@@ -81,15 +83,19 @@ public STSAssumeRoleSessionCredentialsProvider awsCredentialsProvider(
...
      * A bean providing a client to work with S3.
      *
+     * @param noOfS3Retries No. of S3 request retries
      * @param awsCredentialsProvider A credentials provider used to instantiate the client.
      * @return An amazon s3 client object
      */
     @Bean
     @ConditionalOnBean(AWSCredentialsProvider.class)
     public AmazonS3Client genieS3Client(
+        @Value("${genie.retry.s3.noOfRetries:5}") final int noOfS3Retries,
```
tgianos: We should build up a type safe version of this so we don't bind to a string directly

PR #408, opened 2016-10-10
Removed the inverse relations in the job entities

ajoymajumdar:
1. Removed the inverse relations in the job entities - JobEntity, JobRequestEntity, JobExecutionEntity and JobMetadataEntity,. This was done to reduce the SQL calls to get the associated entities.
2. Modified certain queries to use named queries.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaJobRequestRepository.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaJobRequestRepository.java
@@ -22,6 +22,7 @@ import org.springframework.stereotype.Repository;
 
 import javax.validation.constraints.NotNull;
 import java.util.Date;
+import java.util.List;
 
 /**
  * Job repository.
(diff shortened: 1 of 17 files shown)


genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaJobRequestRepository.java
```diff
@@ -33,10 +34,17 @@
...
-     * Delete all job requests created before the given date.
+     * Returns all job requests created before the given date.
      *
-     * @param date The date before which all job requests should be deleted.
-     * @return The number of deleted records
+     * @param date The date before which all job requests were created.
+     * @return List of job requests
      */
-    Long deleteByCreatedBefore(@NotNull final Date date);
+    List<JobRequestEntity> findByCreatedBefore(@NotNull final Date date);
```
tgianos: As discussed my only concern here is memory if this list becomes too large. Currently this method is called daily as part of a task so we should be ok but it's something we might want to improve going forward either by returning just the Id's or paginating this list.

PR #413, opened 2016-10-18
Added retries to service level methods that make persistence calls.

ajoymajumdar:
Using spring AOP, added a retry aspect for service methods that make persistence calls. Retries happen on exceptions - CannotGetJdbcConnectionException, CannotAcquireLockException, DeadlockLoserDataAccessException, OptimisticLockingFailureException, PessimisticLockingFailureException, ConcurrencyFailureException, QueryTimeoutException, TransientDataAccessResourceException

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/aspect/DataServiceRetryAspect.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/aspect/DataServiceRetryAspect.java
@@ -1,0 +1,115 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


genie-web/src/main/java/com/netflix/genie/web/aspect/DataServiceRetryAspect.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Constructor.
+     * @param noOfRetries number of retries
+     * @param initialInterval initial interval in milliseconds before retrying the first attempt
+     * @param maxInterval maximum interval in milliseconds
+     */
+    @Autowired
+    public DataServiceRetryAspect(
+        @Value("${genie.data.service.retry.noOfRetries:5}") final int noOfRetries,
```
tgianos: Change these to bind to a configuration properties class so we can better respond to code changes.

Issue #46, opened 2013-11-08
Support patterns in the cluster/search Name field


jimcistaro:
It would be great if the cluster/search Name field supported patterns.

PR #418, opened 2016-10-21
Initial version of a Genie 3 demo built on Docker Compose.

tgianos:
Able to run demo without any installation using just gradle and docker.

### DIFF
--- a/genie-demo/src/main/docker/client/example/init_demo.py
+++ b/genie-demo/src/main/docker/client/example/init_demo.py
@@ -1,0 +1,65 @@
+#!/usr/bin/python2.7
+
+# Copyright 2016 Netflix, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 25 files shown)


genie-demo/src/main/docker/client/example/init_demo.py
```diff
@@ -0,0 +1,65 @@
...
+        return yaml.load(_file)
+
+genie_conf = GenieConf()
+genie_conf.genie.url = "http://genie:8080"
+
+genie = Genie(genie_conf)
+
+hadoop_application = load_yaml("applications/hadoop271.yml")
+hadoop_application_id = genie.create_application(hadoop_application)
+LOGGER.info("Created hadoop application with id = [%s]" % hadoop_application_id)
```
stephen-mw: You're setting the `LOGGER` to `ERROR` so I don't think you'll see these lines. Also #42, #46, etc.

genie-demo/src/main/docker/client/example/run_hadoop_job.py
```diff
@@ -0,0 +1,64 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+##################################################################################
+# This script assumes setup.py has already been run to configure Genie and that
+# this script is executed on the host where Genie is running. If it's executed on
+# another host change the localhost line below.
+##################################################################################
+
+from __future__ import absolute_import, division, print_function, unicode_literals
```
stephen-mw: It looks like you don't need any of these imports. Same in the other `run_*_job.py` files.

PR #420, opened 2016-10-21
Changing ApplicationReadyEvent to ContextRefreshedEvent

tgianos:
This makes it picked up in application AND war deployment models for things that need to happen at application startup.


tgianos:
haha sorry @aaronblohowiak for some reason tried to assign @ajoymajumdar and it put you in instead when I pressed enter

PR #424, opened 2016-10-26
Cache webpack bundle command in gradle builds

cabhishek:


genie-web/build.gradle
```diff
@@ -117,6 +117,8 @@ node {
 }
 
 task bundle(type: NpmTask) {
+    inputs.files(fileTree("${project.projectDir}/src/main/resources/static"))
```
tgianos: why not just use `inputs.dir` instead of `inputs.files` and creating a file tree manually?

tgianos: It also seems like you should add two `inputs.file` lines and add the `webpack.config.js` and `package.json` in case there are changes to them.

PR #426, opened 2016-10-27
Resolved issue where a genie job errors out  with meesage 'Text file busy' when running the run script.

ajoymajumdar:
Before running the script, added check to see if it is used by any other programs. We check for 3 times and even after that if it is being used by another program, we go ahead and try running it, instead of failing.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
@@ -63,6 +63,7 @@ public class JobKickoffTask extends GenieBaseTask {
     private final Executor executor;
     private final String hostname;
     private final Timer timer;
+    private final Registry registry;
     private final RetryTemplate retryTemplate;
 
     /**
@@ -85,6 +86,7 @@ public class JobKickoffTask extends GenieBaseTask {
         this.isUserCreationEnabled = userCreationEnabled;
         this.executor = executor;
         this.hostname = hostname;
+        this.registry = registry;
         this.timer = registry.timer("genie.jobs.tasks.jobKickoffTask.timer");
         retryTemplate = new RetryTemplate();
         retryTemplate.setBackOffPolicy(new ExponentialBackOffPolicy());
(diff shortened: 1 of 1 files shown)


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -82,6 +86,8 @@ public JobKickoffTask(
         this.executor = executor;
         this.hostname = hostname;
         this.timer = registry.timer("genie.jobs.tasks.jobKickoffTask.timer");
+        retryTemplate = new RetryTemplate();
+        retryTemplate.setBackOffPolicy(new ExponentialBackOffPolicy());
```
tgianos: Don't we already have a retry template as a bean somewhere? why isn't this passed into the constructor?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -162,20 +172,30 @@ public void executeTask(@NotNull final Map<String, Object> context) throws Genie
             } catch (final IOException ie) {
                 throw new GenieServerException("Unable to start command " + String.valueOf(command), ie);
             }
-            log.info("Finished Job Kickoff Task for job {}", jobExecEnv.getJobRequest().getId());
+            log.info("Finished Job Kickoff Task for job {}", jobRequest.getId());
```
tgianos: This will log out as [optional: etc]. Don't we already have the id somewhere referenced so we can just use it directly?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -162,20 +172,30 @@ public void executeTask(@NotNull final Map<String, Object> context) throws Genie
...
     }
 
+    private boolean canExecute(final String runScriptFile) {
+        try {
+            return retryTemplate.execute(c -> {
+                FileUtils.touch(new File(runScriptFile));
+                return true;
+            });
+        } catch (Exception e) {
+            log.warn("Failed touching the run script file", e);
```
tgianos: Maybe we should add a metric here to get a sense how often this happens?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobKickoffTask.java
ajoymajumdar: Don't want to use the same. I feel instantiating one for this class and using it internally is the best.

jelez:
genie-app-3.0.8.jar
We get multiple of those issues:
```
setsid: failed to execute /opt/genie/tmp/jobs/644c2420-602b-11e7-b9b8-06e23ce966f0/run: Text file busy
```
Is this related to this: https://bugs.openjdk.java.net/browse/JDK-8068370

PR #430, opened 2016-11-03
Modified handling the job events for better coordination

ajoymajumdar:
1. Modified to trigger the job launch when handling the JobScheduledEvent. Earlier there was a possibility of a race condition.
2. Use one concurrent map of JobInfo to track the jobs on a local instance. Synchronize on one job instead of all the jobs.
3. On job kill/interruption, stop running the rest of the tasks in the LocalJobRunner.

### DIFF
--- a/genie-client/src/test/java/com/netflix/genie/client/JobClientIntegrationTests.java
+++ b/genie-client/src/test/java/com/netflix/genie/client/JobClientIntegrationTests.java
@@ -51,6 +51,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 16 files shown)


genie-client/src/test/java/com/netflix/genie/client/JobClientIntegrationTests.java
```diff
@@ -151,6 +155,70 @@ public void canSubmitJob() throws Exception {
     }
 
     /**
+     * Method to test submitting/killing a job.
+     *
+     * @throws Exception If there is any problem.
+     */
+    @Test
+    public void submitAndKillJob() throws Exception {
```
tgianos: Don't we have a test for killing the job in the JobRestControllerIntegrationTests?

genie-core/src/main/java/com/netflix/genie/core/events/JobScheduledEvent.java
```diff
@@ -32,25 +36,37 @@
 @Getter
 public class JobScheduledEvent extends BaseJobEvent {
```
tgianos: So we talked about renaming this cause not it's more of a "schedule a job event" and that's something that might be better as a direct method call to a service than an event? The events are intended to be listened and acted on by multiple listeners that are idempotent. This seems more like a direct action that needs to be taken.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -251,30 +236,11 @@ public String coordinateJob(
...
-                            return jobId;
-                        } catch (final TaskRejectedException e) {
-                            throw new GenieServerException(
-                                "Unable to launch job due to exception: " + e.getMessage(),
-                                e
-                            );
-                        }
+                        // Tell the system a new job has been scheduled so any actions can be taken
+                        log.info("Publishing job scheduled event for job {}", jobId);
+                        this.eventPublisher.publishEvent(new JobScheduledEvent(jobId, jobRequest,
```
tgianos: What happens now if the scheduling causes an exception?

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitoringCoordinator.java
```diff
@@ -116,10 +123,10 @@ public JobMonitoringCoordinator(
         this.registry = registry;
         this.jobsDir = jobsDir.getFile();
         this.jobsProperties = jobsProperties;
+        this.jobSubmitterService = jobSubmitterService;
 
         // Automatically track the number of jobs running on this node
-        this.registry.mapSize("genie.jobs.running.gauge", this.jobMonitors);
-        this.registry.mapSize("genie.jobs.scheduled.gauge", this.scheduledJobs);
+        this.registry.mapSize("genie.jobs.running.gauge", this.jobs);
         this.registry.methodValue("genie.jobs.active.gauge", this, "getNumActiveJobs");
```
tgianos: maybe it's better to tag them with the status so that we can break them up that way?

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitoringCoordinator.java
```diff
@@ -146,10 +153,23 @@ public void onStartup(final ContextRefreshedEvent event) throws GenieException {
...
      */
     @EventListener
-    public synchronized void onJobScheduled(final JobScheduledEvent event) {
-        // Increment the amount of memory used to account for this job
-        this.jobMemories.put(event.getId(), event.getMemory());
-        this.scheduledJobs.put(event.getId(), event.getTask());
+    public void onJobScheduled(final JobScheduledEvent event) {
+        final String jobId = event.getId();
+        jobs.put(jobId, new JobInfo());
+        handle(event.getId(), () -> {
```
tgianos: oh fancy functional programming haha

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
ajoymajumdar: We update the job status to Failed but we do not raise JobFinishedEvent event.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -239,60 +211,43 @@ public String coordinateJob(
...
+                            + "/"
+                            + maxSystemMemory
+                            + " MB are used and requested "
+                            + memory
+                            + " MB"
+                    );
                 }
             }
+        } catch (GenieException e) {
+            jobStateService.done(jobId);
```
tgianos: Is there a reason we don't have the persistence steps done within the state machine? It seems to me they're tightly coupled no?

genie-web/src/main/java/com/netflix/genie/web/controllers/JobRestController.java
```diff
@@ -781,7 +781,9 @@ private void copyRequestHeaders(final HttpServletRequest request, final ClientHt
     private void copyResponseHeaders(final HttpServletResponse response, final ClientHttpResponse forwardResponse) {
         final HttpHeaders headers = forwardResponse.getHeaders();
         for (final Map.Entry<String, String> header : headers.toSingleValueMap().entrySet()) {
-            response.setHeader(header.getKey(), header.getValue());
+            if (!TRANSFER_ENCODING_HEADER.equalsIgnoreCase(header.getKey())) {
```
tgianos: I would add a comment here why you're doing this so that no one takes it out by accident going forward

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
ajoymajumdar: It does makes sense but wanted to keep it separate to being with. First draft to be in-memory. This should actually have all the job related persistence steps done from within here.

genie-web/src/main/java/com/netflix/genie/web/controllers/JobRestController.java
ajoymajumdar: Will add.

PR #437, opened 2016-11-17
Throw GenieConflictException if a job request comes in with an id that is already running.

ajoymajumdar:
Previous change had opened this bug where the JobCoordinator was updating the job that is already running with a failed status. Made changes to reject a job request if a job is submitted with an id that is already running.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
@@ -155,7 +155,6 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
         JobStatus jobStatus = JobStatus.FAILED;
         try {
             log.info("Called to schedule job launch for job {}", jobId);
-            jobStateService.init(jobId);
             // create the job object in the database with status INIT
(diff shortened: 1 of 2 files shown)


genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -240,6 +241,9 @@ public String coordinateJob(
                     );
                 }
             }
+        } catch (GenieConflictException e) {
```
tgianos: This will call done on an already running job if there is indeed a conflict. As discussed in person we need a better solution for this.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -242,15 +241,18 @@ public String coordinateJob(
                 }
             }
         } catch (GenieConflictException e) {
-            jobStateService.done(jobId);
             throw e;
```
tgianos: Lets put comments why the catches are designed like this and maybe a TODO: so when we fix the boundaries between responsibilities it's clear what we need to remove

PR #438, opened 2016-11-22
Add a health indicator for the CPU usage

ajoymajumdar:
2. Corrected the error message when a URI path is invalid.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/health/GenieCpuHealthIndicator.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/health/GenieCpuHealthIndicator.java
@@ -17,14 +17,15 @@
  */
 package com.netflix.genie.web.health;
 
+import com.netflix.genie.core.properties.HealthProperties;
 import com.sun.management.OperatingSystemMXBean;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.stereotype.Component;
 
+import javax.validation.constraints.NotNull;
 import java.lang.management.ManagementFactory;
 
 /**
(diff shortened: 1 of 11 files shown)


genie-web/src/test/java/com/netflix/genie/web/health/GenieHealthIndicatorUnitTests.java
```diff
@@ -41,7 +41,7 @@
     private static final int DEFAULT_JOB_MEMORY = 1_024;
     private static final int MAX_JOB_MEMORY = 5_120;
 
-    private GenieHealthIndicator genieHealthIndicator;
```
tgianos: You should rename this test class file if you're also renaming the class it's testing

genie-web/src/main/java/com/netflix/genie/web/health/GenieCpuHealthIndicator.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    @Autowired
+    public GenieCpuHealthIndicator(
+        @Value("${genie.health.cpu.maxLoadPercent:80}")
+        final double maxCpuLoadPercent
+    ) {
+        this(maxCpuLoadPercent, (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean());
+    }
+
+    GenieCpuHealthIndicator(final double maxCpuLoadPercent,
```
tgianos: I assume this is package protected for testing purposes?

genie-web/src/main/java/com/netflix/genie/web/health/GenieCpuHealthIndicator.java
```diff
@@ -0,0 +1,76 @@
...
+    private final double maxCpuLoadPercent;
+    private final OperatingSystemMXBean operatingSystemMXBean;
+    /**
+     * Constructor.
+     *
+     * @param maxCpuLoadPercent The maximum physical memory threshold
+     */
+    @Autowired
+    public GenieCpuHealthIndicator(
+        @Value("${genie.health.cpu.maxLoadPercent:80}")
```
tgianos: I'm becoming increasingly uncomfortable using these `@Value` Strings. They're hard to hunt down when we want to change something and when we give them a default value like this it doesn't really break anything if we say change the name elsewhere this will still work. It might be better to create a `@ConfigurationProperties` class that at least binds at compile time. Also I would add whatever default value you have to the properties file so we have a record of it when we go to document.

PR #439, opened 2016-12-08
Initial REST documentation

tgianos:
Initial REST API documentation via [Spring Rest Docs](https://projects.spring.io/spring-restdocs/) and [Asciidoctor](http://asciidoctor.org/).

Still needs to flesh out the "glue" sections explaining things like return codes, error codes, security, etc but all the API's are now documented and included in the tests so figured this was a good time to merge.


genie-web/src/test/java/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests.java
```diff
@@ -607,12 +902,88 @@ private void testForConflicts(
...
             .andExpect(MockMvcResultMatchers.status().isConflict());
     }
 
     /**
+     * Test the job submit method for success twice to validate the file cache use.
+     *
+     * @throws Exception If there is a problem.
+     */
+    @Test
+    public void testSubmitJobMethodTwiceSuccess() throws Exception {
```
tgianos: @ajoymajumdar can you verify that this is actually needed? it doesn't seem to be doing anything to me other than slowing the build down

PR #448, opened 2017-01-05
Modified CPU health indicator to be more reliable.

ajoymajumdar:
Earlier the CPU health indicator was erratic because of the erratic nature of cpu metric. Added two strategies to reduce it.
1. Take an average of the CPU metric instead of the current value.
2. Mark the service down only if the CPU metric crosses the threshold for a number of times (configured by a property. Default is 3)

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/properties/HealthProperties.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/properties/HealthProperties.java
@@ -33,4 +33,8 @@ public class HealthProperties {
      * Default to 80 percentage.
      */
     private double maxCpuLoadPercent = 80;
+    /**
+     * Default to 3.
+     */
+    private int maxCpuLoadConsecutiveOccurrences = 3;
 }
--- a/genie-core/src/main/java/com/netflix/genie/core/properties/HealthProperties.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/properties/HealthProperties.java
@@ -30,10 +30,16 @@ import lombok.Setter;
 @Setter
 public class HealthProperties {
     /**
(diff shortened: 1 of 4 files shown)


genie-web/src/main/java/com/netflix/genie/web/health/GenieCpuHealthIndicator.java
```diff
@@ -38,36 +41,58 @@
...
+        taskScheduler.scheduleAtFixedRate(() -> this.summaryCpuMetric
+            .record((long) (operatingSystemMXBean.getSystemCpuLoad() * 100)), 5000);
     }
 
     @Override
     public Health health() {
-        final double currentCpuLoadPercent = operatingSystemMXBean.getSystemCpuLoad() * 100;
+        final long cpuCount = summaryCpuMetric.count();
+        final long cpuTotal = summaryCpuMetric.totalAmount();
+        final double currentCpuLoadPercent = cpuCount == 0 ? 0 : (cpuTotal / (double) cpuCount);
```
dmuino: This will not be the current cpu load, but rather the average since the class was instantiated. `totalAmount` and `count` return cumulative values since the meter was created.

genie-core/src/main/java/com/netflix/genie/core/properties/HealthProperties.java
```diff
@@ -33,4 +33,8 @@
      * Default to 80 percentage.
      */
     private double maxCpuLoadPercent = 80;
+    /**
+     * Default to 3.
+     */
+    private int maxCpuLoadConsecutiveOccurrences = 3;
```
tgianos: Please add documentation for this property to the docs _properties.adoc file

PR #452, opened 2017-01-09
change request retry logic to retry everything except specified failure status codes

irontable:


genie-client/src/main/python/pygenie/adapter/genie_3.py
```diff
@@ -396,7 +397,9 @@ def submit_job(self, job, timeout=30):
              url='{}/{}'.format(job._conf.genie.url, Genie3Adapter.JOBS_ENDPOINT),
              files=files,
              timeout=None if self.disable_timeout else timeout,
-             auth_handler=self.auth_handler)
+             auth_handler=self.auth_handler,
+             failure_codes=409,
```
irontable: 409s are returned when submitting a job with an id that already exists.

`failure_codes` can be an array or just a single code.

PR #457, opened 2017-01-17
strip quotes when setting job name from script

irontable:


irontable:
In the Python client, if a job name is not specified it will take the contents of the script/command and set it as the job name. The client was not removing quotes from the script/command and when Genie tries to set the job name the quotes cause a failure in the underlying shell scripts.

PR #458, opened 2017-01-18
allow specifying multiple property files for Hive and Pig job

irontable:
This change will allow specifying multiple property files instead of just 1 for HiveJob (Python client).


irontable:
This PR applies to both HiveJob and PigJob.

Issue #463, opened 2017-01-30
Error when saving configuration to postgres


otterc:
Caused by: org.postgresql.util.PSQLException: ERROR: syntax error at or near "user"
  Position: 82
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2458) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2158) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:291) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:432) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:358) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:171) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:138) ~[postgresql-9.4.1211.jre7.jar!/:9.4.1211.jre7]
	at org.hibernate.engine.jdbc.internal.ResultSetReturnImpl.executeUpdate(ResultSetReturnImpl.java:208) ~[hibernate-core-4.3.11.Final.jar!/:4.3.11.Final]

tgianos:
Confirmed this is an issue with Postgres 3.0.0 schema and code. Unfortunately updating `@Column` definition to `"\"user\""` solves the issue for Postgres but creates issues in other stores. It makes more sense at this point to fix this issue by renaming the column to `genie_user` instead of constantly fighting against reserved keywords and workarounds.

tgianos:
@otterc This issue should be fixed in release `3.0.1` with the new schema. There are database upgrade scripts or entirely new schema files as well. I'll cherry-pick this into master as well so that the same fix will be in going forward.

PR #464, opened 2017-01-30
#463 quoted user column

chandnisingh:


tgianos:
Hello @chandnisingh I see you opened an issue against this. Clearly it breaks all the tests when the new annotation definition is defined the way you tried to change it. Are you attempting to generate your schema via JPA schema auto generation? Genie isn't set up to do this for Mysql or Postgres. We have schema definition files in the `genie-ddl` directory. Are you still having problems even when you run against that schema?

tgianos:
Note you can find the Schema file for 3.0.0 [here](https://github.com/Netflix/genie/blob/v3.0.0/genie-ddl/src/main/sql/postgresql/3.0.0-schema.postgresql.sql)

chandnisingh:
@tgianos I used the schema file for 3.0.0. 
It was giving the error which I posted. I was able to build the app jar locally with the change I made. After which I was able  to save configuration.

tgianos:
Hi @chandnisingh I've confirmed that as it's currently written it doesn't work with Postgres due to the `user` column. We believed this worked when the schema was done but since we don't use Postgres internally the integration tests on travis work against MySQL. Unfortunately while your fix does work for Postgres and should be the JPA 2.0 standard it looks like it breaks the tests for HSQLDB and possibly MySQL. I think at this point it's more prudent to rename the column from the protected `user` keyword to something like `genie_user`. I'll work on this when I get a chance and push a patch 3.0.1 release with an upgrade script for schemas. For now it sounds like you have a binary compiled locally you can work with. Let us know if you're blocked otherwise. @ajoymajumdar FYI

PR #488, opened 2017-04-03
add more info to warning error when making requests in Python client

irontable:
When the Python client makes a HTTP request to the server and it fails, the client logs a message with attempt, etc. Add request URL and method to this message.


stephen-mw:
LGTM.

PR #504, opened 2017-05-11
Fix calculation of available job capacity returned in HealthIndicator

mprimi:


genie-web/src/main/java/com/netflix/genie/web/health/GenieMemoryHealthIndicator.java
```diff
@@ -84,8 +84,8 @@ public Health health() {
             .withDetail(NUMBER_RUNNING_JOBS_KEY, this.jobMetricsService.getNumActiveJobs())
             .withDetail(USED_MEMORY_KEY, usedMemory)
             .withDetail(AVAILABLE_MEMORY, availableMemory)
-            .withDetail(AVAILABLE_DEFAULT_JOB_CAPACITY, availableMemory % defaultJobMemory)
-            .withDetail(AVAILABLE_MAX_JOB_CAPACITY, availableMemory % maxJobMemory)
+            .withDetail(AVAILABLE_DEFAULT_JOB_CAPACITY, availableMemory / defaultJobMemory)
```
mprimi: @tgianos These can now cause division by zero exception.
Is it worth worrying about it?
lombok does not seem to have any way to enforce this constraint in the setter. Applying javax validator setter does not prevent setting to zero.

PR #505, opened 2017-05-12
Update demo dockerfile for latest Genie client. Remove iPython

mprimi:


mprimi:
This is a cherry-pick of https://github.com/Netflix/genie/pull/499
However notice it does not carry-over the changes to `build.gradle`. 
They were conflicting so I left them out since they are out of scope anyway for this, and picked just the iPhython change.

PR #507, opened 2017-05-13
Display reason when job gets killed by monitor

mprimi:


mprimi:
@tgianos @ajoymajumdar This is ready for review. Passes all tests, although I'd like to add some kind of end-to-end test, but I'm not sure how to go about it.

.travis.yml
```diff
@@ -39,3 +39,5 @@ env:
 notifications:
   slack:
     secure: H5nS+GX6TYTU27ur6YFG5OgrQeUbzXLok5ub6+xcmyYEeVPpnQ1Gg/wKqTAGsP9j6tAkqPpxgYT9i9Do6eyTEplK6bTvQVyhilsEDtxGJbUO8XOE9TSo6jAe/lD3EB5l46gxFID+Hg9IkPii4LwEabP7PVehrB1JfNZ6QDgSRRM=
+after_failure:
+  - cat ./genie-client/build/test-results/test/TEST-com.netflix.genie.client.ClusterClientIntegrationTests.xml
```
ajoymajumdar: Is this temporary? If not, should print out all the output files under test-results/test.

mprimi: Yeah it's meant to be temporary

mprimi:
Moved out some test improvements into a different PR: https://github.com/Netflix/genie/pull/513 so this is more self-contained. Notice 2 commits doing two tangential thing, we could skip the one adding Qualifier, it's not required.
This is ready for review and consistently passing tests.

genie-common/src/main/java/com/netflix/genie/common/dto/JobStatusMessage.java
```diff
@@ -0,0 +1,72 @@
...
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.common.dto;
+
+/**
+ * Constant strings for status message attached to a job after it terminates.
+ *
+ * @author mprimi
```
tgianos: Please add `@since 3.0.7` so we know about when the files were created

genie-common/src/main/java/com/netflix/genie/common/dto/JobStatusMessage.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public final class JobStatusMessage {
```
tgianos: It's possible we could make these eventually part of the JobStatus enum itself as another field if we were to expand the job status enums to more states. Would remove need for another class.

genie-core/src/main/java/com/netflix/genie/core/jobs/JobKillReasonFile.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+/**
+ * Class that represents the structure of the kill-reason file created when a job is killed.
+ *
+ * @author mprimi
+ * @since 3.0.8
```
tgianos: think you mean 3.0.7 here.

genie-core/src/main/java/com/netflix/genie/core/jobs/JobKillReasonFile.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+@NoArgsConstructor
```
tgianos: Use `@Data` should suffice for all these.

I prefer Dto's to be read only. Perhaps what you should do here (if it's intended to be bound to json) is just create the constructor yourself and annotate it with `@JsonCreator` that way jackson can bind it but we can still make it read only.

genie-core/src/main/java/com/netflix/genie/core/jobs/JobKillReasonFile.java
```diff
@@ -0,0 +1,37 @@
...
+public class JobKillReasonFile {
+    private String killReason;
```
tgianos: Once this is read only make this final.

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -57,6 +61,8 @@
     private final Executor executor;
     private final boolean runAsUser;
     private final ApplicationEventPublisher eventPublisher;
+    private final File baseWorkingDir;
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
tgianos: I believe there is already an `ObjectMapper` available as a bean within the spring context you should be able to inject here rather than creating another one. Also makes it hard to test with this buried inside.

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -131,6 +146,15 @@ public void killJob(
                 // Windows, etc. May support later
                 throw new java.lang.UnsupportedOperationException("Genie isn't currently supported on this OS");
             }
+
+            // Write additional file with kill reason
+            try {
+                this.objectMapper.writeValue(
+                    new File(this.baseWorkingDir + "/" + id + "/genie/kill-reason"),
```
tgianos: I would make the name of this file a constant probably in `JobConstants` that way we can at least change it and it will change all over the code where we reference it and we don't have to track down strings.

genie-core/src/test/java/com/netflix/genie/core/configs/ServicesConfigTest.java
```diff
@@ -408,4 +417,26 @@ public FactoryBean fileTransferFactory() {
...
+
+    /**
+     * Returns a temporary directory as the jobs resource.
+     *
+     * @return The job dir as a resource.
+     * @throws IOException If there is a problem.
+     */
+    @Bean
+    public Resource jobsDir() throws IOException {
+        final File jobsDir = Files.createTempDir();
```
tgianos: A) I think we already have this bean somewhere
B) please use the `Junit` rule that creates a temporary directory/ file for these things. That guarantees its cleaned up at the end of tests

genie-core/src/test/java/com/netflix/genie/core/jobs/JobKillReasonFileTest.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Test serialization and deserialization of JobKillReasonFile.
+     * @throws IOException in case of serialization error
+     */
+    @Test
+    public void serializeThenLoad() throws IOException {
+
+        final ObjectMapper objectMapper = new ObjectMapper();
+
+        final JobKillReasonFile orignalJobKillReasonFile = new JobKillReasonFile("Test");
```
tgianos: Break the actual reason out into a String and reference that variable in the assertion rather than using the getter in both places since this leads to possible test that is only succeeding due to some weird behavior in the getter. Minor and silly but still.

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobKillServiceImpl.java
mprimi: Also the bean is not present during tests, same as `jobDir` below. Let's take a look together. You can probably show me a trick or two to debug beans creation and injection.

genie-core/src/test/java/com/netflix/genie/core/configs/ServicesConfigTest.java
mprimi: Adding this as a `@Rule TemporaryFolder` in `GenieCoreTestApplication` is not going to work because the latter is not a JUnit class and so JUnit annotations such as `@Rule` are simply not detected and processed (i.e. like adding a `@Before` annotation to a method in `JobMonitor.java`, it's just having no effect).

mprimi: Just pushed a commit that uses `@PreDestroy` to cleanup the temporary folder.

PR #510, opened 2017-05-15
Adding docker image builds to PR's

tgianos:


mprimi:
Only a minor nitpick: in this PR a task is defined and then in different modules other tasks are marked as dependencies.

Wouldn't it be more readable if the dependencies were listed in the task definition? e.g.:
```
task dockerBuildAllImages(, ..., dependsOn: [ :projectA:taskX', ':projectB:taskY'] ) {
}
```

Alternatively, you could make it depend on all tasks without even listing them explicitly with something like:
```
dockerBuildAllImages.dependsOn {
    tasks.findAll { task -> task.name.startsWith('dockerBuild') }
}
```

tgianos:
@mprimi the original intent was allow the images to be "pluggable". It looks like tasks.findAll doesn't work for the subproject tasks so I added them explicitly to the top level task in the latest commit. It's a bit less magic this way so I'll leave it like this per your suggestion.

PR #512, opened 2017-05-16
Lock down npm dependencies

cabhishek:
- Adds `npm-shrinkwrap.json` to lock down all npm dependencies 
- Update code to use `PropTypes` from `prop-types` package
- Move bable config to `package.json`
- Remove `eslint` config (use prettier instead)


mprimi:
@cabhishek Does this need to be cherry-picked into `master`? 
(I think so, but not very familiar with these pieces, so if you could confirm)

mprimi:
@cabhishek nevermind, I missed it earlier but it's there.

PR #513, opened 2017-05-17
Test improvements

mprimi:
* Remove flaky test that relies on timing (JobClientIntegrationTests::submitAndKillJob)
* Fix failing test that relies on platform-specific output (JobClientIntegrationTests:: testCanGetJobStderr)
* Increase Gradle verbosity to see each test name and outcome as they are executed
* Re-enable JobClientIntegrationTests class
* Change Gradle settings for better test isolation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,6 +84,8 @@ configure(javaProjects) {
         testLogging {
             events "passed", "skipped", "failed"
         }
+        forkEvery 1
+        maxParallelForks 4
     }
 
     ext {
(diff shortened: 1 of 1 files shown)


mprimi:
This makes the test build run a lot longer, however it removes some unpredictability and some tests that were ignored can be re-enabled and pass.
We might discuss if it's worth doing.

mprimi:
Closing this and focusing on master. Will cherry-pick back as needed.

PR #517, opened 2017-05-25
Add simple user job limiting mechanism

mprimi:
Motivation:
Introduce a simple mechanism that allows administrators to set a upper limit to the number of jobs  users are allowed to concurrently run. The limit, if enabled, is applied to all users.
This mechanism is designed as a first line of defense against a user issuing an avalanche of job submissions (either maliciously (i.e. to cause denial of service), or unintentionally (i.e. client-side bug).

Change:
 * Implement database query that returns the number of active jobs for a given user (in `JPAJobSearchService`)
 * Introduce new properties to toggle and configure the per-user active jobs limit
 * Introduce new `GenieException` to model a user limit exceeded
 * Reject job submissions (in `JobCoordinationService`) if a user exceeded the limit

Note:
It is still possible for a user to issue more jobs than the limit.
For example, by issuing two jobs on two genie instances at exactly the same time, both of them will check the limit and let a job through, even if one one of them should have.

### DIFF
--- a/genie-common/src/main/java/com/netflix/genie/common/exceptions/GenieUserLimitExceededException.java
+++ b/genie-common/src/main/java/com/netflix/genie/common/exceptions/GenieUserLimitExceededException.java
@@ -1,0 +1,68 @@
+/*
+ *
+ *  Copyright 2015 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.common.exceptions;
+
+import lombok.Getter;
+
+/**
+ * Extension of a GenieException for a user exceeding some limit (e.g., submitting too many jobs).
+ *
+ * @author mprimi
+ * @since 3.1.0
+ */
+@Getter
+public final class GenieUserLimitExceededException extends GenieException {
+
+    private static final String ACTIVE_JOBS_LIMIT = "activeJobs";
+    private final String user;
+    private final String exceededLimitName;
+
+    private GenieUserLimitExceededException(
+        final String user,
+        final String limitName,
+        final String message
+    ) {
+        super(429, message);
+        this.user = user;
+        this.exceededLimitName = limitName;
+    }
+
+    /**
+     * Static factory method to produce a GenieUserLimitExceededException suitable for when the user exceeded the
+     * maximum number of active jobs and its trying to submit yet another.
+     * @param user the user name
+     * @param activeJobsCount the count of active jobs for this user
+     * @param activeJobsLimit the current limit on active jobs
+     * @return a new GenieUserLimitExceededException
+     */
+    public static GenieUserLimitExceededException createForActiveJobsLimit(
+        final String user,
+        final long activeJobsCount,
+        final long activeJobsLimit
+    ) {
+        return new GenieUserLimitExceededException(
+            user,
+            ACTIVE_JOBS_LIMIT,
+            "User exceeded active jobs limit ("
+                + activeJobsCount
+                + "/"
+                + activeJobsLimit
+                + ")"
+        );
+    }
+}
--- a/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
@@ -28,6 +28,7 @@ import com.netflix.genie.common.dto.JobStatus;
 import com.netflix.genie.common.dto.search.JobSearchResult;
 import com.netflix.genie.common.exceptions.GenieException;
 import com.netflix.genie.common.exceptions.GenieNotFoundException;
+import com.netflix.genie.common.exceptions.GenieServerException;
 import com.netflix.genie.core.jpa.entities.ApplicationEntity;
 import com.netflix.genie.core.jpa.entities.ClusterEntity;
 import com.netflix.genie.core.jpa.entities.CommandEntity;
@@ -49,6 +50,7 @@ import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.validation.annotation.Validated;
 
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
@@ -73,6 +75,7 @@ import java.util.stream.Collectors;
  */
 @Slf4j
 @Transactional(readOnly = true)
+@Validated
 public class JpaJobSearchServiceImpl implements JobSearchService {
 
     private final JpaJobRepository jobRepository;
@@ -361,5 +364,23 @@ public class JpaJobSearchServiceImpl implements JobSearchService {
         } else {
             throw new GenieNotFoundException("No job execution found for id " + jobId);
         }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public long getActiveJobCountForUser(@NotBlank final String user) throws GenieException {
+        log.debug("Called for jobs with user {}", user);
+        final Long count = this.jobRepository.countJobsByUserAndStatusIn(user, JobStatus.getActiveStatuses());
+        if (count == null || count < 0) {
+            throw new GenieServerException(
+                "Count query for user "
+                    + user
+                    + "produced an unexpected result: "
+                    + count
+            );
+        }
+        return count;
     }
 }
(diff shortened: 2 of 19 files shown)


mprimi:
@tgianos this is ready for review. I went with a simpler change than we originally discussed.
However I'd like to add a bit more testing. For example REST controller and client, but the integration test for those is currently ignored.

mprimi:
(p.s. you can look at commits individually and in order and that should make the review easier since it's piecemeal)

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -362,4 +363,43 @@ public String getJobHost(@NotBlank final String jobId) throws GenieException {
             throw new GenieNotFoundException("No job execution found for id " + jobId);
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Long getActiveJobCountByUserName(@NotBlank final String userName) {
```
tgianos: There are much simpler ways to do this with Spring Data JPA. The reason that in the job search we use this dynamic build up via JPA criteria builder is because there are some parameters that may or may not be there hence we use if statements to do it.

In this instance you know exactly what parameters you want to pass in so you can leverage their nice query builder methods that do it all for you.

See https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods for more.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -362,4 +363,43 @@ public String getJobHost(@NotBlank final String jobId) throws GenieException {
...
+        log.debug("Called with username {}", userName);
+
+        if (StringUtils.isBlank(userName)) {
```
tgianos: You shouldn't have to do this as you already have `@NotBlank` on the method parameter. Additionally what does `String.valueOf("    ")` or other blank get you

genie-web/src/main/java/com/netflix/genie/web/controllers/GenieExceptionMapper.java
```diff
@@ -105,6 +108,8 @@ public void handleGenieException(
             this.serverUnavailableRate.increment();
         } else if (e instanceof GenieTimeoutException) {
             this.timeoutRate.increment();
+        } else if (e instanceof GenieUserLimitExceededException) {
+            this.userLimitExceededRate.increment();
```
tgianos: we should tag this metric with the user name who is doing it.

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -100,6 +100,10 @@ rights for this to work.
 to work.
 |false
 
+|genie.jobs.users.activeJobsLimit
+|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. Setting this property to zero or a negative value means no limit is enforced.
+|0
```
tgianos: This is one way to do it another is a feature flag that enables or disables this. Do you have any reason you went with this over that? Setting to zero or negative is kind of counter intuitive as opposed to setting to true or false for enabled and then taking a setting...

genie-web/src/main/resources/application.yml
```diff
@@ -51,6 +51,7 @@ genie:
     users:
       creationEnabled: false
       runAsUserEnabled: false
+      concurrentJobLimit: 0
```
tgianos: This default variable does not match what you have in the documentation above

tgianos: This won't bind to your current Java class.

genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersProperties.java
```diff
@@ -31,4 +31,5 @@
 public class JobsUsersProperties {
     private boolean creationEnabled;
     private boolean runAsUserEnabled;
+    private int activeJobsLimit;
```
tgianos: Assuming you're using the java language default of 0 for integer here rather than saying it explicitly

genie-web/src/main/java/com/netflix/genie/web/controllers/GenieExceptionMapper.java
mprimi: Done (also tagged what kind of limit was exceeded, even if right now we have a single one)

genie-web/src/main/resources/application.yml
mprimi: Bug. Did a last minute rename in the code and missed this one. Fixed now (and became 2 variables, enable and value).

genie-common/src/main/java/com/netflix/genie/common/exceptions/GenieUserLimitExceededException.java
```diff
@@ -0,0 +1,73 @@
...
+        final String user,
+        final String limitName,
+        final String message
+    ) {
+        super(429, message);
+        this.user = user;
+        this.exceededLimitName = limitName;
+    }
+
+    public String getUser() {
```
tgianos: just add `@Getter` from lombok to the class so that you don't have to define this boilerplate yourself.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -362,4 +363,16 @@ public String getJobHost(@NotBlank final String jobId) throws GenieException {
             throw new GenieNotFoundException("No job execution found for id " + jobId);
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Long getActiveJobCountForUser(@NotBlank final String user) throws GenieException {
```
tgianos: So for this issue you should be able to add `@Validated` to the class to force the `@NotBlank` to fire. That way you don't need the redundant error checking within the method itself. It should be annotated on other classes.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
```diff
@@ -362,4 +363,16 @@ public String getJobHost(@NotBlank final String jobId) throws GenieException {
...
+        log.debug("Called for jobs with user {}", user);
+        if (StringUtils.isEmpty(StringUtils.trimToEmpty(user))) {
+            throw new IllegalArgumentException("User cannot be null or empty");
+        }
+        return this.jobRepository.countJobsByUserAndStatusIn(user, JobStatus.getActiveStatuses());
```
tgianos: Rather than returning the Nullable `Long` you might want to check for null here and dereference to the primitive  `long` or `0L` if null that way clients to this service don't have to worry about performing the null check themselves.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -134,6 +140,17 @@ public JobCoordinatorServiceImpl(
         this.selectCommandTimer = registry.timer("genie.jobs.submit.localRunner.selectCommand.timer");
         this.selectApplicationsTimer = registry.timer("genie.jobs.submit.localRunner.selectApplications.timer");
         this.setJobEnvironmentTimer = registry.timer("genie.jobs.submit.localRunner.setJobEnvironment.timer");
+
+        if (this.jobsProperties.getUsers().isActiveJobsLimitEnabled()) {
+            final int activeJobsLimit = this.jobsProperties.getUsers().getActiveJobsLimit();
```
tgianos: You definitely shouldn't have to check this every time. You should be able to put `@Size` or other validation in the properties class. You could write your own setter in the properties class as well that should force a quick fail when the property was set to invalid value or that resets the value to something allowed. This checking within the constructor here becomes distributed as you may have this check in multiple places (not that you do now but you could). It should be centralized in the data object itself.

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -100,6 +100,14 @@ rights for this to work.
 to work.
 |false
 
+|genie.jobs.users.activeJobsLimitEnabled
```
tgianos: you might want to change this to something like `genie.jobs.users.activeLimit.enabled` and then the other can be named `genie.jobs.users.activeLimit.count` that way it's easier to group them. Just a thought.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobSearchServiceImpl.java
mprimi: Whoa, i just deleted your comment on this line (i meant to delete mine).
You were saying how i shouldn't null/empty check myself.
As per side-conversation. It would be nice to test these annotations work, but it's out of scope for this.
Removed null/empty checking from this.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
mprimi: Removed this and added a `@Min(value=1)` annotation in on the property object.

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Done

PR #520, opened 2017-05-26
Display reason when job gets killed by monitor (#507)

mprimi:
Motivation:
If a job is killed by JobMonitor, the kill reason is lost and not propagated to the user through the UI, which is confusing.
The reason for this can be summarized as:
1. `JobMonitor` emits a `KillJobEvent` which containing a reason string.
2. `KillJobEvent` is consumed by `JobKillService`, which issues a `kill` command.
3. `JobCompletionServer` eventually detects process termination and updates the database, but the reason is lost.
4. The UI displays the generic "killed" message, which can confuse a user who didn't issue a kill.

Change overview:
Make JobKillService write an extra file `kill-reason` in the job folder, containing the reason for killing as present in the `KillJobEvent`.
`JobCompletionServer` looks for and reads the file before performing the final job database update.

Change content:

* Update interfaces Job{Coordinator,Kill}Service::killJob() to take an extra mandatory “reason” argument
* Introduce JobKillReasonFile to serialize and load job kill reason as JSON
* Pass base working directory as to LocalJobKillService during construction
* Add 'jobsDir' to test services configuration (ServicesConfigTest)
* Introduce new JobStatusMessage class with static constant strings used in job status messages
* Adding explicit Qualifier to autowired users of ‘jobDir’ resource


mprimi:
@tgianos cherry-picking from 3.0.x to master

PR #521, opened 2017-05-31
Add PropertiesConfig integration test

mprimi:


mprimi:
@tgianos Small change ready for review

mprimi:
How did I manage to *reduce* coverage by adding a test??

tgianos:
Cause coveralls is weird with rounding errors it seems.

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Integration test for PropertiesConfig values binding.
+ *
+ * @author mprimi
+ * @since 3.1.0
+ */
+@Category(IntegrationTest.class)
+@RunWith(SpringRunner.class)
+@TestPropertySource(locations = "classpath:/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.properties")
```
tgianos: You could likely also accomplish this (with a yml file as well) by just putting a properties file in source/test/resources and calling it like `application-propstest.yml` and then activating the `@Profile("propstest")` for this test. That should load that properties source. Also I wouldn't put this properties file explicitely in this location as it inconsistent with the rest of our locations.

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
```diff
@@ -0,0 +1,77 @@
...
+@SpringBootTest()
```
tgianos: Do you need to set anything here to make sure a full app server is not loaded?

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
mprimi: I like the idea of profile, it's more similar to how properties are normally loaded (and it's YAML). I'll switch to this and move this file with the other profiles.

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
mprimi: Without args defaults to a mock server, but i could make it explicit.

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
mprimi: Scratch that. For some reason, profile does not work. I tried to figure it out for an hour but in the interest of time, I'm leaving this as is (but moving it to the top level test resource directory)

mprimi:
@tgianos moved the file, but kept everything else the same (as i couldn't get loading via profile to work).

PR #523, opened 2017-06-02
Fix bug in JobCoordinatorService, recording metric with wrong unit

mprimi:


mprimi:
@tgianos looked at all other uses of `timer::record()` and they look correct.
Also checked on Altas, and this is the only one that seems way off.

tgianos:
@mprimi we probably want to fix this in 3.0.x as well.

PR #525, opened 2017-06-02
Switch to better dependency management

tgianos:


build.gradle
```diff
@@ -74,6 +74,28 @@ configure(javaProjects) {
...
+                entry "converter-jackson"
+            }
+            dependency("com.netflix.spectator:spectator-api:0.30.0")
+            dependencySet(group: "io.springfox", version: "2.7.0") {
+                entry "springfox-swagger2"
+                entry "springfox-swagger-ui"
+                entry "springfox-bean-validators"
+            }
+            dependency("commons-validator:commons-validator:1.5.1")
+            dependency("net.sf.jtidy:jtidy:r938")
```
mprimi: Doesn't adding dependencies like this add them to all configuration?
i.e. this `jtidy` thing was a `testCompile` dependency.
By moving it here, did it just become a `compile` dependency for all project modules?

tgianos: No. This only says what version to use if someone includes the dependency in the dependency section of the Java plugin. This is configuring the dependency management plugin so it's different. Just allows us to keep it all in one place and not have to remember what variable to use as a version

mprimi: Oh ok, I didn't notice the parent is `dependencyManagement`. I assumed this was a top-level `dependencies` which would have made these global project dependencies.

PR #526, opened 2017-06-12
Job failing initialization can get forever stuck in INIT state

mprimi:
A job failing during initialization can emit a JobFinishedEvent with a 'statusMsg' field exceeding the database 256 character limit for the same field.
This causes the job status change to fail to persist and results in the job being stuck in INIT state forever.

This change consists of the following:
 - Centralize "status messages" used over the project into a static set (in JobStatusMessage)
 - Where applicable, leave behind a file with the detailed trace of the error, for the user to inspect

### DIFF
--- a/genie-common/src/main/java/com/netflix/genie/common/dto/JobStatusMessage.java
+++ b/genie-common/src/main/java/com/netflix/genie/common/dto/JobStatusMessage.java
@@ -67,6 +67,39 @@ public final class JobStatusMessage {
     public static final String JOB_PROCESS_NOT_FOUND = "Couldn't check job process status.";
 
     /**
+     * Job PID terminated, exist status is yet to be verified.
+     */
+    public static final String PROCESS_DETECTED_TO_BE_COMPLETE = "Process detected to be complete";
+
+    /**
+     * Job is undefined intermediate state caused by a crash during setup.
+     */
+    public static final String SYSTEM_CRASHED_WHILE_JOB_STARTING = "System crashed while job starting";
+
+    /**
+     * Job was launched before Genie stopped, and it cannot be re-attached after restart.
+     */
+    public static final String UNABLE_TO_RE_ATTACH_ON_STARTUP = "Unable to re-attach on startup";
+
+    /**
+     * Job was killed by user before even starting.
+     */
+    public static final String USER_REQUESTED_JOB_BE_KILLED_DURING_INITIALIZATION =
+        "User requested job be killed during initialization";
+
+    /**
+     * Job precondition was not satisfied during initialization.
+     */
+    public static final String SUBMIT_PRECONDITION_FAILURE =
+        "Job validation failed, further details available in the job output directory";
+
+    /**
+     * Job failed with unexpected exception during initialization.
+     */
+    public static final String SUBMIT_INIT_FAILURE =
+        "Job initialization failed, further details available in the job output directory";
+
+    /**
      * Private constructor, this class is not meant to be instantiated.
      */
     private JobStatusMessage() {
--- a/genie-core/src/main/java/com/netflix/genie/core/events/JobFinishedEvent.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/events/JobFinishedEvent.java
@@ -21,6 +21,7 @@ import lombok.Getter;
 import org.hibernate.validator.constraints.NotEmpty;
 
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 
 /**
  * An event thrown when a job is completed.
@@ -45,7 +46,7 @@ public class JobFinishedEvent extends BaseJobEvent {
     public JobFinishedEvent(
         @NotEmpty final String id,
         @NotNull final JobFinishedReason reason,
-        @NotEmpty final String message,
+        @NotEmpty @Size(max = 255, min = 1) final String message,
         @NotNull final Object source
     ) {
         super(id, source);
--- a/genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
@@ -135,6 +135,11 @@ public final class JobConstants {
     public static final String GENIE_KILL_REASON_FILE_NAME = "genie/kill-reason";
 
     /**
+     * File created by Genie with details and trace for a job that failed to initialize.
+     **/
+    public static final String GENIE_INIT_FAILURE_MESSAGE_FILE_NAME = "init-failure-details";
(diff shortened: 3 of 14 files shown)


genie-common/src/main/java/com/netflix/genie/common/dto/JobStatusMessage.java
```diff
@@ -67,6 +67,39 @@
     public static final String JOB_PROCESS_NOT_FOUND = "Couldn't check job process status.";
 
     /**
+     * Job PID terminated, exist status is yet to be verified.
+     */
+    public static final String PROCESS_DETECTED_TO_BE_COMPLETE = "Process detected to be complete";
```
tgianos: Seems to me this class should be named `JobStatusMessages.java`. Just nitpicky but more accurate.

genie-core/src/main/java/com/netflix/genie/core/events/JobFinishedEvent.java
```diff
@@ -45,7 +46,7 @@
     public JobFinishedEvent(
         @NotEmpty final String id,
         @NotNull final JobFinishedReason reason,
-        @NotEmpty final String message,
+        @NotEmpty @Size(max = 255, min = 1) final String message,
```
tgianos: Do we ever validate these records?

genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
```diff
@@ -135,6 +135,11 @@
     public static final String GENIE_KILL_REASON_FILE_NAME = "genie/kill-reason";
 
     /**
+     * File created by Genie with details and trace for a job that failed to initialize.
+     **/
+    public static final String GENIE_INIT_FAILURE_MESSAGE_FILE_NAME = "init-failure-details";
```
tgianos: we might want to think about making this file a `.txt` or similar so that mime types can read it properly. Also dashes in file name do we think that's best?

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
```diff
@@ -203,6 +215,38 @@ public void submitJob(
...
+        final long start = System.nanoTime();
+        try {
+            final File jobDir = new File(this.baseWorkingDirPath.getFile(), id);
+            if (jobDir.exists()) {
+                final File detailsFile = new File(jobDir, JobConstants.GENIE_INIT_FAILURE_MESSAGE_FILE_NAME);
+                final boolean detailsFileExists = !detailsFile.createNewFile();
+                if (detailsFileExists) {
+                    log.warn("Init failure details file exists");
+                }
+                try (final Writer writer = new OutputStreamWriter(new FileOutputStream(detailsFile), "UTF-8")) {
```
tgianos: I think we have a constant for `UTF-8` lying around somewhere in JobConstants...

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
```diff
@@ -203,6 +215,38 @@ public void submitJob(
...
+
+                    writer.write("Job "  + id + " failed initialization due to: " + e.getMessage());
```
tgianos: Is there a way to templatize this string and then just replace the pieces we need?

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
```diff
@@ -203,6 +215,38 @@ public void submitJob(
...
+                    writer.write(System.lineSeparator());
+                    e.printStackTrace(new PrintWriter(writer));
+                    writer.write(System.lineSeparator());
+                }
+                log.info("Created init failure details file {}", detailsFile);
+            } else {
+                log.error("Could not create init failure details file, job directory does not exist");
+            }
+        } catch (Throwable t) {
+            log.error("Failed to create init failure details file", t);
```
tgianos: Seems to me there should be a counter here. Also why catch throwable and not Exception?

genie-common/src/main/java/com/netflix/genie/common/dto/JobStatusMessage.java
mprimi: Agree that's a better name. 
Are we going to cherry-pick this in `master`? (I'd say yes, but not completely sure what's the difference between the two right now)
Because if not, I don't think we should rename, as it would make cherry-picks a pain.

genie-core/src/main/java/com/netflix/genie/core/events/JobFinishedEvent.java
mprimi: I'm still not sure about the validation magic. I'm sure it's not active in test. In the actual app configuration there is a Validator bean, however I'm not certain in would validate this class.

genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
mprimi: I can add the extension (most other files don't have one, so i was just going with convention).
Same for dashes in the name. (What's the concern with those?)

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
mprimi: Ok, will change. That string is hardcoded in other parts of this file so I just copy-pasted honestly.

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
mprimi: Will do.

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
mprimi: I assumed timer also includes a counter. And things like createJobDir also use a timer rather than a counter so just went for the same.

As for `Throwable` I really wanted to make sure that this function does not throw.

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
mprimi: The only constant I'm seeing is: https://github.com/Netflix/genie/blob/3.0.x/genie-web/src/main/java/com/netflix/genie/web/resources/handlers/GenieResourceHttpRequestHandler.java#L55

PR #527, opened 2017-06-13
Increase Gradle test verbosity

mprimi:


build.gradle
```diff
@@ -80,6 +80,12 @@ configure(javaProjects) {
         }
     }
 
+    test {
+        testLogging {
+            events "passed", "skipped", "failed"
```
mprimi: This just prints {PASSED,SKIPPED,FAILED} next to each test as they complete.

mprimi: (Useful mostly on Travis CI, where a failure does not say anything about what test failed and there's no easy way to get the unittest reports)

tgianos: If it's really only useful on travis can't we conditionally enable it by checking the existence system variable `CI`? So that we don't have to see a lot of stuff locally we don't really care about as we can always see the test reports locally.

PR #533, opened 2017-06-16
Upgrade to Gradle 4

tgianos:


mprimi:
Tried this. Seems to work just like before, but faster 👍

PR #534, opened 2017-06-16
Improvements to job launch and run script

mprimi:
Change: swap the order of `setsid` and `sudo` at task launch. This ensure `sudo` is in a different process group than the Genie app [1], as such it will respond to a signal by forwarding it to its children (the run script).
Motivation: a recent `sudo` patch [1] makes sudo not forward signals to its children if the signal comes from a process in the same group.

Change: write genie.done into a different file, and atomically swap it if one does not exist already.
Motivation: if the command completes during handing of a kill signal, it may overwrite the genie.done file created by the kill handler.

Change: use absolute paths for output files created in the `run` script.
Motivation: if a command performs a `pushd` or changes working directory, the `run` script may fail to write output files.

Change: use `kill -0` to check for termination of the child process of the `run` script.
Motivation: the existing solution with `pgrep` was not working on Linux.

Change: increase verbosity of the `run` script.
Motivation: better visibility into what happened at runtime.

Change: disable 'failure' trap once inside the 'kill' trap.
Motivation: the 'failure' has no reason to execute and can override output files.

Change: remove hardcoded 'genie/genie.done' from strings
Motivation: use the existing constant string for easier code navigation

[1] https://www.sudo.ws/repos/sudo/rev/7ffa2eefd3c0 see also https://stackoverflow.com/questions/34337840/cant-terminate-a-sudo-process-created-with-python-in-ubuntu-15-10/34376188#34376188


genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
```diff
@@ -130,6 +130,12 @@
     public static final String GENIE_DONE_FILE_NAME = "genie/genie.done";
 
     /**
+     * Temporary done filename generated by Genie after running a job as not to conflict/overwrite the
+     * one that 'trap' handler might have created.
+     **/
+    public static final String GENIE_TEMPORARY_DONE_FILE_NAME = "genie/temp.genie.done";
```
tgianos: not sure but I thought most temporary files in *nix were desired to be {filename}.tmp or similar rather than prepending? Not a huge deal either way

genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
```diff
@@ -269,12 +280,14 @@
...
         .append("    ERROR_CODE=$?\n")
+        .append("    echo \"Handling exit signal (code: ${ERROR_CODE})\"\n")
+        .append("\n")
         .append("    # Good exit\n")
         .append("    if [[ ${ERROR_CODE} -eq 0 ]]; then\n")
         .append("        exit 0\n")
         .append("    fi\n")
         .append("    # Bad exit\n")
-        .append("    printf '{\"exitCode\": \"%s\"}\\n' \"${ERROR_CODE}\" > ./genie/genie.done\n")
+        .append("    printf '{\"exitCode\": \"%s\"}\\n' \"${ERROR_CODE}\" > ${GENIE_JOB_DIR}/genie/genie.done\n")
```
tgianos: Since we use `/genie` a lot perhaps we should make it its own variable. Like `${GENIE_LOG_DIR}` or something? Just a convenience thought.

Also I think there is a String variable for the name of the done file and we probably should use that here instead of the genie.done string. That way in case we change it it changes here too.

genie-core/src/main/java/com/netflix/genie/core/jobs/JobConstants.java
```diff
@@ -283,50 +296,40 @@
...
+        .append("    trap \"\" SIGTERM ERR EXIT\n")
         .append("\n")
         .append("    echo \"Kill signal received\"\n")
         .append("\n")
+        .append("    KILL_EXIT_CODE=").append(JobExecution.KILLED_EXIT_CODE).append("\n")
+        .append("\n")
         .append("    ### Write the kill exit code to genie.done file as exit code before doing anything else\n")
         .append("    echo \"Generate done file with exit code ${KILL_EXIT_CODE}\"\n")
-        .append("    printf '{\"exitCode\": \"%s\"}\\n' \"${KILL_EXIT_CODE}\" > ./genie/genie.done\n")
+        .append("    printf '{\"exitCode\": \"%s\"}\\n' \"${KILL_EXIT_CODE}\" > ${GENIE_JOB_DIR}/genie/genie.done\n")
```
tgianos: Same comment about the genie.done file.

genie-web/src/test/resources/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests/linux-runsh.txt
```diff
@@ -7,12 +7,14 @@ trap "handle_failure" ERR EXIT
...
     ERROR_CODE=$?
+    echo "Handling exit signal (code: ${ERROR_CODE})"
+
     # Good exit
     if [[ ${ERROR_CODE} -eq 0 ]]; then
         exit 0
     fi
     # Bad exit
-    printf '{"exitCode": "%s"}\n' "${ERROR_CODE}" > ./genie/genie.done
+    printf '{"exitCode": "%s"}\n' "${ERROR_CODE}" > ${GENIE_JOB_DIR}/genie/genie.done
```
tgianos: Done file hardcoding.

genie-web/src/test/resources/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests/linux-runsh.txt
```diff
@@ -21,44 +23,39 @@ trap "handle_kill_request" SIGTERM
...
+    trap "" SIGTERM ERR EXIT
 
     echo "Kill signal received"
 
+    KILL_EXIT_CODE=999
+
     ### Write the kill exit code to genie.done file as exit code before doing anything else
     echo "Generate done file with exit code ${KILL_EXIT_CODE}"
-    printf '{"exitCode": "%s"}\n' "${KILL_EXIT_CODE}" > ./genie/genie.done
+    printf '{"exitCode": "%s"}\n' "${KILL_EXIT_CODE}" > ${GENIE_JOB_DIR}/genie/genie.done
```
tgianos: Done file hardcoding.

genie-web/src/test/resources/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests/linux-runsh.txt
```diff
@@ -109,9 +106,12 @@ source ${GENIE_JOB_DIR}/jobsetupfile
...
-wait $!
+/bin/bash -c 'echo hello world' > ${GENIE_JOB_DIR}/stdout 2> ${GENIE_JOB_DIR}/stderr &
+export CHILDREN_PID=$!
+wait ${CHILDREN_PID}
 
 # Write the return code from the command in the done file.
-printf '{"exitCode": "%s"}\n' "$?" > ./genie/genie.done
+printf '{"exitCode": "%s"}\n' "$?" > ${GENIE_JOB_DIR}/genie/temp.genie.done
+# Swapping done file, unless one exist created by trap handler.
+mv -n ${GENIE_JOB_DIR}/genie/temp.genie.done ${GENIE_JOB_DIR}/genie/genie.done
```
tgianos: Done file hardcoding.

genie-web/src/test/resources/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests/linux-runsh.txt
mprimi: This (and the next two hardcoded comments) are on expected outputs that we compare against in test. I don't think we should remove this string (if we do, we need to export $GENIE_DONE_FILE as environment variable in the run script.

PR #536, opened 2017-06-19
Test with upcoming Travis image upgrade

mprimi:
https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2


mprimi:
Tests passed, no need to merge as this new image will become default.

PR #543, opened 2017-06-28
Expose selected cluster and command tags in run script environment

mprimi:
In order to allow tasks to make decisions based on the cluster that was selected by Genie, expose two new environment variables in the `run` script: GENIE_{COMMAND,CLUSTER}_TAGS


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
```diff
@@ -329,4 +358,22 @@ private void createJobEnvironmentVariables(
         // Append new line
         writer.write(LINE_SEPARATOR);
     }
+
+    private String tagsToString(final Set<String> tags) {
```
tgianos: Get rid of this whole method.

Use [this](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#join-java.lang.Iterable-char-)

Also should we try to bash escape any existing commas?

mprimi:
@tgianos this is ready for review. Easier to digest if you look at the two commits individually.

genie-core/src/test/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTaskUnitTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * Setup to run before each test.
+     * @throws IOException if a temporary folder cannot be created
+     */
+    @Before
+    public void setUp() throws IOException {
+        final Registry registry = Mockito.mock(Registry.class);
+        this.initialSetupTask = new InitialSetupTask(registry);
+        this.temporaryTestDirectory = Files.createTempDirectory(InitialSetupTaskUnitTest.class.getSimpleName() + "-");
```
tgianos: Is there a reason you didn't use [this](http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html)?

mprimi: Should have.

tgianos: are you going to change this or do you want me to approve as is since it works?

mprimi: No I'll change it, everything else looks ok?

tgianos: Yeah I'm not really sure why you bothered marking package protected methods `@VisibleForTesting` but I guess it doesn't really hurt anything

mprimi: Findbugs (and possibly other static analysis tools) may cry "this method should be private!".
Adding @VisibleForTesting tells them the reason it's not private:
```
/**
 * Annotates a program element that exists, or is more widely visible than
 * otherwise necessary, only for use in test code.
 */
public @interface VisibleForTesting {
}
```

mprimi: (Sorry I just noticed my response is not here. Maybe I never hit the submit button)

I had to make a bunch of private methods package local for the sake of testing them, and this is exactly what [this annotation is for ](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html), static analysis tools will take note of it and not complain.

mprimi:
@tgianos Addressed review comments

genie-core/src/test/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTaskUnitTest.java
```diff
@@ -37,23 +48,165 @@
...
      * Setup to run before each test.
+     * @throws IOException if a temporary folder cannot be created
      */
     @Before
-    public void setUp() {
-        this.registry = Mockito.mock(Registry.class);
+    public void setUp() throws IOException {
+        final Registry registry = Mockito.mock(Registry.class);
         this.initialSetupTask = new InitialSetupTask(registry);
+        this.temporaryTestDirectory = Files.createTempDirectory(this.getClass().getSimpleName() + "-");
```
tgianos: I thought you were going to switch this to use the `@Rule` from JUnit for temp directories?

mprimi: Changed to use @Rule. Not sure what I was doing wrong before, but copying from a different test worked without problems.

PR #547, opened 2017-06-29
Add dependencies to cluster entity

mprimi:


mprimi:
Opening for early feedback, but I still have **a lot** of tests to update/write and review the docs.

mprimi:
@tgianos Ready for review

genie-client/src/test/java/com/netflix/genie/client/GenieClientsIntegrationTestsBase.java
```diff
@@ -84,12 +84,17 @@ Cluster constructClusterDTO(final String id) {
         configList.add("config1");
         configList.add("configs2");
 
+        final Set<String> dependenciesList = new HashSet<>();
```
tgianos: Just a FYI but you can use the Guava `Sets.newHashSet(item1, item2, ...)` method to shortcut this kind of thing

genie-common/src/main/java/com/netflix/genie/common/dto/Cluster.java
```diff
@@ -37,6 +40,7 @@
 
     @NotNull(message = "A valid cluster status is required")
     private final ClusterStatus status;
+    private final Set<String> dependencies = new HashSet<>();
```
tgianos: As a general design comment for improving the code base I'm pretty sure now all the major dto's (JobRequest, Job, Cluster, Command, Application) now have dependencies as a field. As such this member variable should be pushed up into a super class that is shared so we don't have repetitive work. I thought this had already been done for some of them but dependencies should be pushed up as well.

genie-common/src/main/java/com/netflix/genie/common/dto/Cluster.java
```diff
@@ -85,6 +100,19 @@ public Builder(
         }
 
         /**
+         * Set the dependencies for the cluster if desired.
+         *
+         * @param dependencies The dependencies
+         * @return The builder
+         */
+        public Cluster.Builder withDependencies(final Set<String> dependencies) {
```
tgianos: Push into higher level builder.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ClusterTask.java
```diff
@@ -95,6 +98,13 @@ public void executeTask(@NotNull final Map<String, Object> context) throws Genie
                 AdminResources.CLUSTER
             );
 
+            // Create the dependencies directory for this id
+            createEntityInstanceDependenciesDirectory(
```
tgianos: At this point you're just creating a directory. If you want some reusable method that's fine but you can just say create a `dependencies` directory within this folder rather than all these arguments

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaClusterServiceImpl.java
```diff
@@ -318,6 +318,68 @@ public void removeAllConfigsForCluster(
      * {@inheritDoc}
      */
     @Override
+    public void addDependenciesForCluster(
```
tgianos: Not saying you need to do this as part of this PR but we have so many of these repeatable crud methods it would be nice to find a way to share implementations across clusters, commands, applications...

genie-ddl/src/main/sql/mysql/schema.mysql.sql
```diff
@@ -1,8 +1,8 @@
--- MySQL dump 10.13  Distrib 5.7.17, for osx10.12 (x86_64)
+-- MySQL dump 10.13  Distrib 5.7.17, for macos10.12 (x86_64)
 --
 -- Host: localhost    Database: genie
 -- ------------------------------------------------------
--- Server version	5.7.17
+-- Server version	5.6.35
```
tgianos: Probably not a huge deal but looks like you're using an older MySQL

genie-docs/src/docs/asciidoc/concepts/_dataModel.adoc
```diff
@@ -48,8 +48,8 @@ An application resource represents pretty much what you'd expect. It is a reusab
 and setup files that can be used to install and configure (surprise!) an application. Generally these resources are
 used when an application isn't already installed and on the `PATH` on a Genie node.
 
-When a job is run and applications are involved in running the job Genie will download all the dependencies,
-configuration files and setup files of each application and store it all in the job working directory. It will then
+When a job is run the job Genie will download all the dependencies, configuration files and
+setup files of each application and cluster and store it all in the job working directory. It will then
```
tgianos: I think commands as well no?

genie-docs/src/docs/asciidoc/concepts/_netflixExample.adoc
```diff
@@ -65,6 +65,7 @@ downloaded into the job working directory at runtime.
     "s3://bucket/users/bdp/h2prod/20161217/205111/genie/hdfs-site.xml",
     "s3://bucket/users/bdp/h2prod/20161217/205111/genie/core-site.xml"
   ],
+  "dependencies": [],
```
tgianos: We may want to update these examples with our "new" way of doing things with HoS

genie-client/src/test/java/com/netflix/genie/client/GenieClientsIntegrationTestsBase.java
mprimi: Ok, went with what the rest of these tests were doing.
In the first commit ("test improvements") of the latest version of this PR, I went through **all** tests and replaced this style with `Sets.newHashSet`

genie-common/src/main/java/com/netflix/genie/common/dto/Cluster.java
mprimi: As discussed offline, postponing this refactoring since it requires changes to Command

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ClusterTask.java
mprimi: Not sure what you are asking to do differently.
I am calling an existing method in BaseTask that has (admittedly very simple) logic to create dependencies folders.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaClusterServiceImpl.java
mprimi: Commands don't have dependencies as of now (although there are hints around the code that seem to indicate this was the case previously, or maybe it's a partially implemented feature that was never fully realized).
For example `createEntityInstanceDependenciesDirectory` knows how to create a folder for 'command' dependencies, but it's never used for that purpose.

If we add dependencies to commands (now in the list of things to do) we can indeed push some this and more up to some common parent and drop a bunch of duplicate code in JPA/DTO/Service/REST/...

genie-ddl/src/main/sql/mysql/schema.mysql.sql
mprimi: Yes, I used MAMP because it's a breeze to install, but it's not the newest.

However since it created a dump pretty much identical to the previous one, and the new table is identical in schema to application_dependencies, makes me think this is ok

genie-docs/src/docs/asciidoc/concepts/_dataModel.adoc
mprimi: Not right now. Will add Command dependencies in a different PR.

mprimi:
@tgianos addressed all review comments (other than the ones we agreed to postpone)

mprimi:
(just tested PostgreSQL schema migration)

mprimi:
Closing this PR and pulling in the changes into https://github.com/Netflix/genie/pull/558

PR #550, opened 2017-07-07
Import missing jquery dependency

cabhishek:


mprimi:
Cherry-picked this commit into a different PR: https://github.com/Netflix/genie/pull/553

PR #551, opened 2017-07-07
Handle empty columns commandCriteria and clusterCriterias

mprimi:
In an upcoming upgrade to the database schema, the columns will be migrated to TEXT, which does not support default value (currently '[]').
This change ensure the code is forward compatible with the new schema, by providing the expected default value via setter


genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
```diff
@@ -159,8 +159,12 @@ protected String getClusterCriterias() {
      *
      * @param clusterCriterias The cluster criterias.
      */
-    protected void setClusterCriterias(@NotBlank final String clusterCriterias) {
-        this.clusterCriterias = clusterCriterias;
+    protected void setClusterCriterias(final String clusterCriterias) {
+        if (StringUtils.isEmpty(clusterCriterias)) {
```
tgianos: Ternary and use `isBlank()`

PR #552, opened 2017-07-10
Add ScriptLoadBalancer

tgianos:
This PR includes a new implementation of the `ClusterLoadBalancer` interface which allows system administrators to provide a location for a script which can be downloaded and evaluated within the JVM for selecting a cluster. Currently it supports `JavaScript` and `Groovy` as script languages. More can be added by adding jars to the classpath.

The documentation has been updated with more details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,13 @@ configure(javaProjects) {
             dependency("org.apache.commons:commons-exec:1.3")
             dependency("org.bitbucket.b_c:jose4j:0.5.4")
             dependency("org.dbunit:dbunit:2.5.3")
+            dependencySet(group: "org.jruby", version:"9.1.12.0") {
+                entry "jruby-complete"
+                entry "jruby-core"
+                entry "jruby-stdlib"
+            }
             dependency("org.opensaml:opensaml:2.6.4")
+            dependency("org.python:jython-standalone:2.7.1")
             dependency("org.springframework.security.extensions:spring-security-saml2-core:1.0.2.RELEASE")
         }
     }
--- a/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
@@ -44,10 +44,12 @@ import com.netflix.genie.core.services.JobKillService;
 import com.netflix.genie.core.services.JobPersistenceService;
 import com.netflix.genie.core.services.JobSearchService;
 import com.netflix.genie.core.services.JobStateService;
+import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.Timer;
 import lombok.extern.slf4j.Slf4j;
 import org.hibernate.validator.constraints.NotBlank;
+import org.hibernate.validator.constraints.NotEmpty;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
@@ -70,6 +72,13 @@ import java.util.stream.Collectors;
 @Slf4j
 public class JobCoordinatorServiceImpl implements JobCoordinatorService {
 
+    private static final String NO_ID_FOUND = "No id found";
+    private static final String LOAD_BALANCER_CLASS_TAG = "class";
+    private static final String LOAD_BALANCER_STATUS_TAG = "status";
+    private static final String LOAD_BALANCER_STATUS_SUCCESS = "success";
+    private static final String LOAD_BALANCER_STATUS_FAILURE = "failure";
+    private static final String LOAD_BALANCER_STATUS_EXCEPTION = "exception";
+
     private final JobPersistenceService jobPersistenceService;
     private final JobKillService jobKillService;
     private final JobStateService jobStateService;
@@ -77,7 +86,7 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
     private final JobSearchService jobSearchService;
     private final ClusterService clusterService;
     private final CommandService commandService;
-    private final ClusterLoadBalancer clusterLoadBalancer;
+    private final List<ClusterLoadBalancer> clusterLoadBalancers;
     private final JobsProperties jobsProperties;
     private final String hostName;
 
@@ -85,11 +94,13 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
     private final Set<CommandStatus> commandStatuses;
 
     // Metrics
+    private final Registry registry;
     private final Timer coordinationTimer;
     private final Timer selectClusterTimer;
     private final Timer selectCommandTimer;
     private final Timer selectApplicationsTimer;
     private final Timer setJobEnvironmentTimer;
+    private final Id loadBalancerId;
 
     /**
      * Constructor.
@@ -103,7 +114,7 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
      * @param jobSearchService      Implementation of job search service
      * @param clusterService        Implementation of cluster service interface
      * @param commandService        Implementation of command service interface
-     * @param clusterLoadBalancer   Implementation of the cluster load balancer interface
+     * @param clusterLoadBalancers  Implementations of the cluster load balancer interface in invocation order
      * @param registry              The registry
      * @param hostName              The name of the host this Genie instance is running on
      */
@@ -116,7 +127,7 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
         @NotNull final JobSearchService jobSearchService,
         @NotNull final ClusterService clusterService,
         @NotNull final CommandService commandService,
-        @NotNull final ClusterLoadBalancer clusterLoadBalancer,
+        @NotNull @NotEmpty final List<ClusterLoadBalancer> clusterLoadBalancers,
         @NotNull final Registry registry,
         @NotBlank final String hostName
     ) {
@@ -127,7 +138,7 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
         this.jobSearchService = jobSearchService;
         this.clusterService = clusterService;
         this.commandService = commandService;
-        this.clusterLoadBalancer = clusterLoadBalancer;
+        this.clusterLoadBalancers = clusterLoadBalancers;
         this.jobsProperties = jobsProperties;
         this.hostName = hostName;
 
@@ -136,11 +147,13 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
         this.commandStatuses.add(CommandStatus.ACTIVE);
 
         // Metrics
+        this.registry = registry;
         this.coordinationTimer = registry.timer("genie.jobs.coordination.timer");
         this.selectClusterTimer = registry.timer("genie.jobs.submit.localRunner.selectCluster.timer");
         this.selectCommandTimer = registry.timer("genie.jobs.submit.localRunner.selectCommand.timer");
         this.selectApplicationsTimer = registry.timer("genie.jobs.submit.localRunner.selectApplications.timer");
         this.setJobEnvironmentTimer = registry.timer("genie.jobs.submit.localRunner.setJobEnvironment.timer");
+        this.loadBalancerId = registry.createId("genie.jobs.submit.selectCluster.loadBalancer");
     }
 
     /**
@@ -149,11 +162,9 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
     @Override
     public String coordinateJob(
         @Valid
-        @NotNull(message = "No job request provided. Unable to submit job for execution.")
-        final JobRequest jobRequest,
+        @NotNull(message = "No job request provided. Unable to execute.") final JobRequest jobRequest,
         @Valid
-        @NotNull(message = "No job metadata provided. Unable to submit job for execution.")
-        final JobMetadata jobMetadata
+        @NotNull(message = "No job metadata provided. Unable to execute.") final JobMetadata jobMetadata
     ) throws GenieException {
         final long coordinationStart = System.nanoTime();
         final String jobId = jobRequest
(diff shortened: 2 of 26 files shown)


genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -331,10 +342,81 @@ private void setRuntimeEnvironment(
...
+                        } else {
+                            this.registry.counter(
+                                this.loadBalancerId
+                                    .withTag(
+                                        LOAD_BALANCER_CLASS_TAG,
+                                        loadBalancerClass
+                                    )
+                                    .withTag(
+                                        LOAD_BALANCER_STATUS_TAG,
+                                        LOAD_BALANCER_STATUS_FAILURE
```
mprimi: Here returning null is considered FAILURE in terms of metrics, but maybe something like "NO_PREFERENCE" is more accurate?

genie-core/src/main/java/com/netflix/genie/core/services/impl/RandomizedClusterLoadBalancerImpl.java
```diff
@@ -52,4 +54,12 @@ public Cluster selectCluster(final List<Cluster> clusters) throws GenieException
         final Random rand = new Random();
         return clusters.get(Math.abs(rand.nextInt(clusters.size())));
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int getOrder() {
+        return Ordered.LOWEST_PRECEDENCE;
```
mprimi: We want this to always go last. What happens if another LB implementation also uses LOWEST_PRECEDENCE?
PriorityOrdered only works the other way around (giving higher precedence if two objects have the same precedence value).

genie-core/src/test/java/com/netflix/genie/core/configs/ServicesConfigTest.java
```diff
@@ -263,8 +263,7 @@ public JobSubmitterService jobSubmitterService(
         final ApplicationEventPublisher eventPublisher,
         final ApplicationEventMulticaster eventMulticaster,
         final List<WorkflowTask> workflowTasks,
-        @Qualifier("jobsDir")
-        final Resource genieWorkingDir,
+        @Qualifier("jobsDir") final Resource genieWorkingDir,
```
mprimi: Ok this is _my_ personal OCD, but this this line change (and a couple others) is unrelated to this feature/patch.
Sneaking it in can make it harder to track down changes and their reason via `git blame`.
Beside, isn't it generally recommended to have each annotation on a new line?
e.g.: 
>[each annotation is listed on a line of its own (that is, one annotation per line)](https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations)

> [When used on a declaration, each annotation often appears, by convention, on its own line](https://docs.oracle.com/javase/tutorial/java/annotations/basics.htm)

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -331,10 +342,81 @@ private void setRuntimeEnvironment(
...
         try {
-            log.info("Selecting cluster for job {}", jobRequest.getId());
-            final Cluster cluster
-                = this.clusterLoadBalancer.selectCluster(this.clusterService.chooseClusterForJobRequest(jobRequest));
-            log.info("Selected cluster {} for job {}", cluster.getId(), jobRequest.getId());
+            log.info("Selecting cluster for job {}", jobRequest.getId().orElse(NO_ID_FOUND));
+            final List<Cluster> clusters = this.clusterService.chooseClusterForJobRequest(jobRequest);
+            Cluster cluster = null;
+            if (clusters.isEmpty()) {
+                throw new GeniePreconditionException(
```
mprimi: How about a metric/tag for this outcome?

genie-docs/src/docs/asciidoc/features/_clusterLoadBalancing.adoc
```diff
@@ -0,0 +1,192 @@
...
+===== Script Contract
+
+The contract between the script and the Java code is as follows:
+
+.Script Inputs
+|===
+|Parameter |Description
+
+|clusters
+|JSON array of cluster objects (serialized as a string) of the clusters to be evaluated
```
mprimi: As of now, this is guaranteed to be non-empty. Should we make it explicit here and in Javadoc?

genie-web/src/main/java/com/netflix/genie/GenieWeb.java
```diff
@@ -41,8 +41,8 @@
 @EnableAspectJAutoProxy
 public class GenieWeb {
 
-    protected static final String SPRING_CONFIG_LOCATION = "spring.config.location";
-    protected static final String USER_HOME_GENIE = "${user.home}/.genie/";
+    static final String SPRING_CONFIG_LOCATION = "spring.config.location";
+    static final String USER_HOME_GENIE = "${user.home}/.genie/";
```
mprimi: Changes in this file are also not strictly related to this commit.
Splitting them into a different commit (and then not squashing on merge) would make the history of these lines more clear.

genie-web/src/main/java/com/netflix/genie/web/configs/ServicesConfig.java
```diff
@@ -235,8 +235,7 @@ public JobKillService jobKillService(
         final Executor executor,
         final JobsProperties jobsProperties,
         final ApplicationEventPublisher eventPublisher,
-        @Qualifier("jobsDir")
-        final Resource genieWorkingDir,
+        @Qualifier("jobsDir") final Resource genieWorkingDir,
```
mprimi: Same as above for style/history.

genie-web/src/main/java/com/netflix/genie/web/configs/ServicesConfig.java
```diff
@@ -286,10 +285,8 @@ public GenieFileTransferService genieFileTransferService(
     @Bean
     public GenieFileTransferService cacheGenieFileTransferService(
         final FileTransferFactory fileTransferFactory,
-        @Value("${genie.file.cache.location}")
```
mprimi: Same as above for style/history.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -331,10 +342,81 @@ private void setRuntimeEnvironment(
     private Cluster getCluster(final JobRequest jobRequest) throws GenieException {
         final long start = System.nanoTime();
         try {
-            log.info("Selecting cluster for job {}", jobRequest.getId());
-            final Cluster cluster
-                = this.clusterLoadBalancer.selectCluster(this.clusterService.chooseClusterForJobRequest(jobRequest));
-            log.info("Selected cluster {} for job {}", cluster.getId(), jobRequest.getId());
+            log.info("Selecting cluster for job {}", jobRequest.getId().orElse(NO_ID_FOUND));
+            final List<Cluster> clusters = this.clusterService.chooseClusterForJobRequest(jobRequest);
```
mprimi: Should this list be immutable to avoid plugins messing with it (as of now, I think LBs can add and remove elements).

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -331,10 +342,81 @@ private void setRuntimeEnvironment(
...
+                                    .withTag(
+                                        LOAD_BALANCER_CLASS_TAG,
+                                        loadBalancerClass
+                                    )
+                                    .withTag(
+                                        LOAD_BALANCER_STATUS_TAG,
+                                        LOAD_BALANCER_STATUS_SUCCESS
+                                    )
+                            ).increment();
+                            cluster = selectedCluster;
```
mprimi: Should we check that cluster is in the original list of clusters?

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -0,0 +1,279 @@
...
+    static final String STATUS_TAG_NOT_CONFIGURED = "not configured";
+    static final String STATUS_TAG_FOUND = "found";
+    static final String STATUS_TAG_FAILED = "failed";
+    static final String EXCEPTION_TAG_KEY = "exception";
+
+    private static final long DEFAULT_TIMEOUT_LENGTH = 5_000L;
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
+    private static final String SLASH = "/";
+    private static final String PERIOD = ".";
+    private static final int ORDER = Ordered.LOWEST_PRECEDENCE - 1;
```
mprimi: Should this be a property?
If someone wants to use this, they have no chance to layer something in between Script and Random

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -0,0 +1,279 @@
...
+    private final AtomicBoolean isConfigured = new AtomicBoolean();
+    private final ScriptEngineManager scriptEngineManager = new ScriptEngineManager();
+
+    private final AsyncTaskExecutor taskExecutor;
+    private final GenieFileTransferService fileTransferService;
+    private final Environment environment;
+    private final ObjectMapper mapper;
+    private final Registry registry;
+
+    private CompiledScript script;
```
mprimi: Should this be an AtomicReference?

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    @Override
+    public int getOrder() {
+        return ORDER;
+    }
+
+    /**
+     * Check if the script file needs to be refreshed.
+     */
+    public void refresh() {
```
mprimi: Maybe just paranoia, but would it be worth wrapping the entire refresh() body into a catch all?
Just to make sure we don't end up killing the executor thread.
(I think the current code is already not supposed to throw, but it wouldn't hurt to add an extra layer)

genie-web/src/test/resources/application-ci.yml
```diff
@@ -30,6 +30,10 @@ genie:
   health:
     maxCpuLoadPercent: 100
   jobs:
+    clusters:
```
mprimi: `application.yml` above does not rely on default property values, whereas this and the next conf do. Is this intentional.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
tgianos: Yeah seems a better defensive way to do this. Will change

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
tgianos: Sure.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
tgianos: Sure. And added counter.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
tgianos: agreed

genie-docs/src/docs/asciidoc/features/_clusterLoadBalancing.adoc
tgianos: done

genie-web/src/main/java/com/netflix/genie/GenieWeb.java
tgianos: Will try to keep this in mind for the future. For now this is already squashed so too late

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
tgianos: I agree with this. Lets do it as another feature in a different PR rather than bundling it in this one in the interest of getting this out.

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
tgianos: Done

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
tgianos: I moved the try up as high as possible while still giving the finally block access to the start time and map of strings. The try block catches every checked exception and `RuntimeException` so nothing can get out of the block.

genie-web/src/test/resources/application-ci.yml
tgianos: It is cause I don't want someone starting up Genie without knowledge of the `ScriptLoadBalancer` and seeing a bunch of Stack Traces periodically cause the system couldn't find the script. But for the integration test I wanted to make sure that the system could still run even if the `ScriptLoadBalancer` was enabled and not configured correctly. I could take it back out but it doesn't really hurt anything for the testing. Ideally this would pull a script to actually test running but the UnitTest already does that.

genie-web/src/test/groovy/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancerSpec.groovy
```diff
@@ -254,5 +254,6 @@ class ScriptLoadBalancerSpec extends Specification {
         type         | file
         "JavaScript" | this.class.getResource("loadBalance.js").file
         "Groovy"     | this.class.getResource("loadBalance.groovy").file
+//        "Python"     | this.class.getResource("loadBalance.py").file
```
mprimi: Maybe forgot to delete this line?

build.gradle
```diff
@@ -94,7 +94,13 @@ configure(javaProjects) {
             dependency("org.apache.commons:commons-exec:1.3")
             dependency("org.bitbucket.b_c:jose4j:0.5.4")
             dependency("org.dbunit:dbunit:2.5.3")
+            dependencySet(group: "org.jruby", version:"9.1.12.0") {
```
mprimi: Are JRuby and Jython really necessary?

tgianos: This is the dependency management section. They're not being added to the project in any way right now other than the versions are locked if we decide to ever add them. In this case I was testing the `ScriptEngine` implementations they had and decided to leave the dependency lock in just in case.

PR #553, opened 2017-07-11
UI Bug + More test for HAL API links 

mprimi:


genie-web/src/test/java/com/netflix/genie/web/controllers/RestControllerIntegrationTestsBase.java
```diff
@@ -382,4 +393,118 @@ void canDeleteTagForResource(
...
+            .append(entityApi)
+            .append("/")
+            .append(entityId)
+            .append("/")
+            .append(linkedEntityType)
+            .toString();
+        final String uriString = new URI(URI_SCHEME, URI_HOST, uriPath, null).toString();
+
+        // Append HAL parameters separately to avoid URI encoding
+        String halParams = "";
```
tgianos: here it seems to me you can just make halParams a new `StringBuilder` and below you can do toString and it will have the same effect. Saves you from allocating effectively two strings

mprimi: Fixed, I messed around a few times with this so the end result was not as clean as it could be.

PR #558, opened 2017-07-12
Cluster and Command dependencies, and miscellaneous database improvements

mprimi:
* Cluster dependencies
 * Command dependencies
 * SQL schema files renaming
 * Upgrade Spectator
 * Add index to jobs::name
 * Migrate job_requests::{cluster,command}_criteria to type TEXT
 * Increase 'tags' column sizes
 * Minor test improvements


genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
```diff
@@ -159,8 +159,12 @@ protected String getClusterCriterias() {
      *
      * @param clusterCriterias The cluster criterias.
      */
-    protected void setClusterCriterias(@NotBlank final String clusterCriterias) {
-        this.clusterCriterias = clusterCriterias;
+    protected void setClusterCriterias(final String clusterCriterias) {
+        if (StringUtils.isEmpty(clusterCriterias)) {
```
tgianos: You could simplify this using the ternary: `this.clusterCriterias = StringUtils.isBlank(clusterCriterias) ? EMPTY_JSON_ARAY : clusterCriterias;`

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
```diff
@@ -246,7 +250,11 @@ protected String getCommandCriteria() {
      * @param commandCriteria A set of command criteria tags as a JSON array
      */
     protected void setCommandCriteria(final String commandCriteria) {
-        this.commandCriteria = commandCriteria;
+        if (StringUtils.isEmpty(commandCriteria)) {
```
tgianos: Ternary

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaClusterServiceImpl.java
```diff
@@ -318,6 +318,68 @@ public void removeAllConfigsForCluster(
      * {@inheritDoc}
      */
     @Override
+    public void addDependenciesForCluster(
+        @NotBlank(message = "No cluster id entered. Unable to add dependencies.")
+        final String id,
+        @NotEmpty(message = "No dependencies entered. Unable to add dependencies.")
+        final Set<String> dependencies
+    ) throws GenieException {
+        this.findCluster(id).getDependencies().addAll(dependencies);
```
tgianos: One thing to be careful about here is that transactions at the Spring level are lost once you make a method call within the same class. By that I mean you do NOT get a new transaction when you call `this.findCluster` since it doesn't go back through the class proxy. In this case I don't think it matters since you're not expecting the other method to commit any intermediate work just something I thought you'd want to know cause we've been burned by it before

genie-core/src/main/java/com/netflix/genie/core/services/ClusterService.java
```diff
@@ -205,6 +205,77 @@ void removeAllConfigsForCluster(
...
+     *
+     * @param id         The id of the cluster to remove the dependency file from. Not
+     *                   null/empty/blank.
+     * @param dependency The dependency file to remove. Not null/empty/blank.
+     * @throws GenieException if there is an error
+     */
+    void removeDependencyForCluster(
+        @NotBlank(message = "No cluster id entered. Unable to remove dependency.")
+        final String id,
+        @NotBlank(message = "No dependency entered. Unable to remove dependency.")
```
tgianos: Project level comment: We probably have a growing technical debt to make sure we handle null inputs and such consistently throughout the project. We should at some point take a look at that

genie-core/src/test/java/com/netflix/genie/core/jpa/entities/ClusterEntityUnitTests.java
```diff
@@ -81,6 +79,8 @@ public void testDefaultConstructor() {
         Assert.assertNull(entity.getVersion());
```
tgianos: As you can probably tell looking at the code a lot of these tests were developed at various points of knowledge for the best way to test things. Basically this is kind of the worst where you have Junit + Mockito + raw Assert statements. Next best would be how it is in some other places where it uses hamcrest Matchers cause the error messages are better and it gives us more flexibility. The best (imo) is probably the Spock stuff as it removes a lot of the boiler plate and can be more readable. More technical debt we might want to address at some point. Outside the scope of this PR just pointing it out as general comment going forward you might want to develop new tests using Spock.

genie-core/src/test/java/com/netflix/genie/core/jpa/entities/ClusterEntityUnitTests.java
```diff
@@ -185,6 +183,21 @@ public void testSetConfigs() {
     }
 
     /**
+     * Test setting the dependencies.
+     */
+    @Test
+    public void testSetDependencies() {
+        Assert.assertNotNull(this.c.getDependencies());
+        final Set<String> dependencies = Sets.newHashSet();
```
tgianos: why not use `Sets.newHashSet({value})` here like you are everywhere else?

genie-ddl/src/main/sql/postgresql/schema.postgresql.sql
```diff
@@ -209,8 +221,8 @@ CREATE TABLE job_requests (
     command_args character varying(10000) NOT NULL,
     group_name character varying(255) DEFAULT NULL::character varying,
     setup_file character varying(1024) DEFAULT NULL::character varying,
-    cluster_criterias character varying(2048) DEFAULT '[]'::character varying NOT NULL,
-    command_criteria character varying(1024) DEFAULT '[]'::character varying NOT NULL,
+    cluster_criterias text DEFAULT ''::character varying NOT NULL,
```
tgianos: It seems here that Postgres allows you to have defaults for their text field so why not just keep it at `[]`?

genie-docs/src/docs/asciidoc/concepts/_howItWorks.adoc
```diff
@@ -191,6 +191,7 @@ directory
 . Application configuration and dependency files are downloaded and stored in the job directory if any applications are
 needed
 . Application related variables are written into the run script
+. Cluster configuration and dependency files are downloaded and store in the job directory
```
tgianos: store -> stored

genie-docs/src/docs/asciidoc/concepts/_howItWorks.adoc
```diff
@@ -191,6 +191,7 @@ directory
...
 . Command configuration and dependency files are downloaded and store in the job directory
```
tgianos: store -> stored

genie-docs/src/docs/asciidoc/concepts/_howItWorks.adoc
```diff
@@ -215,7 +216,7 @@ killed depending on that code.
 
 ===== Genie Performs Job Clean-Up
 
-To save disk space Genie will delete application dependencies from the job working directory after a job is completed.
+To save disk space Genie will delete application, cluster, job dependencies from the job working directory after a job is completed.
```
tgianos: we don't delete job dependencies.

genie-ddl/src/main/sql/postgresql/schema.postgresql.sql
mprimi: Just wanted to be consistent between MySQL and PostgreSQL

mprimi:
Re: entity tag max size annotation.
Uh, i missed that (I did a search for '2048' but somehow that didn't come up).
What do you think we should do? New intermediate entity for large tags, parent of Command, Application, Cluster?
An added bonus is that I could pull up 'dependencies' into it.

genie-core/src/main/java/com/netflix/genie/core/services/ClusterService.java
mprimi: I did notice a bit of this. 
Maybe one way is to have a layer that strictly checks what comes in through the API by just looking at the request (i.e. null field that cannot be null, or input too large, etc).
Then a second layer of validation that checks internal state (i.e. if performing a cluster operation, check that the cluster exists, etc).
Catching all of these errors as early as possible would allow for better error messages and would simplify the inner logic since we can avoid checking certain things.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
mprimi: Fixed

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
mprimi: Fixed

genie-core/src/test/java/com/netflix/genie/core/jpa/entities/ClusterEntityUnitTests.java
mprimi: Fixed

genie-docs/src/docs/asciidoc/concepts/_howItWorks.adoc
mprimi: Fixed

genie-docs/src/docs/asciidoc/concepts/_howItWorks.adoc
mprimi: Fixed

genie-docs/src/docs/asciidoc/concepts/_howItWorks.adoc
mprimi: Fixed

mprimi:
@tgianos added Command dependencies. This PR is now pretty huge, but commits can be looked at individually for easier review.
I couldn't find a nice way to remove some of the code for dependencies duplicated across App/Cmd/Cluster due to how some of it it's generated by annotations, I'll talk to you tomorrow about doing some refactoring to reduce it.

genie-common/src/main/java/com/netflix/genie/common/dto/ExecutionEnvironmentDTO.java
```diff
@@ -32,9 +32,10 @@
  * @since 3.0.0
  */
 @Getter
-public abstract class SetupFileAndDependenciesDTO extends CommonDTO {
+public abstract class ExecutionEnvironmentDTO extends CommonDTO {
 
     private static final long serialVersionUID = 2116254045303538065L;
+    protected final Set<String> configs = new HashSet<>();
```
tgianos: I don't think these should be `protected` they should be private and the super classes should use the getters.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
```diff
@@ -84,6 +84,11 @@
 
     @Lob
     @Basic
```
tgianos: If you're marking the database field as `TEXT NOT NULL` i think you need to have `@Basic(optional=false)` here to make it clear it's not nullable.  Same with `@Column(nullable=false)`

genie-common/src/main/java/com/netflix/genie/common/dto/ExecutionEnvironmentDTO.java
mprimi: Done

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobRequestEntity.java
mprimi: Done

PR #559, opened 2017-07-12
Precondition failed when no valid (non-blank) Cluster Criteria or Command Criteria submitted

tgianos:
This PR modifies the bean validation on `JobRequest` payloads to ensure that there is at least one non-blank tag in both one of the `ClusterCriteria` and the `CommandCriteria`. This prevents the job request from getting through and causing non-deterministic behavior in the DB query as `""` tags will match every cluster or command in DB.

Now when a job request is made with this precondition violated the entire request fails and returns a 412 HTTP status code to the user. The job request is NOT saved to the database since it was invalid anyway.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/controllers/GenieExceptionMapper.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/controllers/GenieExceptionMapper.java
@@ -31,6 +31,7 @@ import com.netflix.spectator.api.Registry;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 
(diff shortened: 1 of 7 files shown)


tgianos:
@skwslide fyi this is to address the issue you brought up yesterday

genie-web/src/main/java/com/netflix/genie/web/controllers/GenieExceptionMapper.java
```diff
@@ -137,4 +141,21 @@ public void handleConstraintViolation(
...
+     * @throws IOException on error in sending error
+     */
+    @ExceptionHandler(MethodArgumentNotValidException.class)
+    public void handleMethodArgumentNotValidException(
+        final HttpServletResponse response,
+        final MethodArgumentNotValidException e
+    ) throws IOException {
+        this.methodArgumentNotValidRate.increment();
+        log.error(e.getMessage(), e);
+        response.sendError(HttpStatus.PRECONDITION_FAILED.value(), e.getMessage());
```
mprimi: Isn't a 400 / Bad request more appropriate for this?
Precondition failed has to do with request headers asking for some condition that cannot be met according to https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

tgianos: This is a good point. I wasn't aware the spec was just based on Header validation. Originally used 412 because just considered it a precondition failed like a method call internally.

I won't change this as part of this PR or in 3.0.x cause it would change way too much by way of API across the board and for clients. We can explore this in 3.1.x

PR #566, opened 2017-07-19
Fix the creation of directories to use File.mkdirs() for consistency across application

tgianos:


genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -245,22 +244,22 @@ public void refresh() {
...
                 this.environment.getProperty(SCRIPT_FILE_DESTINATION_PROPERTY_KEY);
-            if (StringUtils.isBlank(scriptFileSourceValue)) {
+            if (StringUtils.isBlank(scriptFileDestinationValue)) {
                 throw new IllegalStateException(
                     "Invalid empty value for script destination directory property: "
                         + SCRIPT_FILE_DESTINATION_PROPERTY_KEY
                 );
             }
-            final String scriptDestinationDir = new URI(scriptFileDestinationValue).getPath();
+            final String scriptDestinationDir = new URI(scriptFileDestinationValue).toString();
```
mprimi: I think`toString()` here will return `file:///foo/bar/`.
At line 260, this is parsed as `File`. Will the `file://` part get dropped?
Will it fail if someone for some reason tries to pass `s3://...`?

tgianos: It should fail if someone sets it to `s3://` because the destination directory must be a local directory on disk.

mprimi: Right, this is good/expected behavior.
I was asking if that's what happens at line 260 when creating a `File` or we need an extra check.

PR #567, opened 2017-07-19
Fix script load balancer and local file transfer to use java.nio

tgianos:
Instead of old java.io.

Fix consistency of script handling for directories and files to be downloaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,8 @@ Thumbs.db
 
 # IntelliJ specific files/directories #
 #######################################
-out
+/out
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -121,6 +121,7 @@ configure(javaProjects) {
     clean {
         delete "${projectDir}/genie-db"
         delete "${projectDir}/data"
+        delete "${projectDir}/out"
```
mprimi: Why is this necessary?

PR #569, opened 2017-07-21
SLF4J-related fixes

dtrebbien:

PR #570, opened 2017-07-21
Utilize parameterized logging and remove calls to toString()

dtrebbien:
As suggested in the SLF4J FAQ: https://www.slf4j.org/faq.html#logging_performance
parameterized logging can improve the efficiency of logging calls when
logging at the specified level is disabled.

In addition, per the FAQ: https://www.slf4j.org/faq.html#string_contents
toString() is called automatically.

These changes were suggested and automatically made by SLF4J Helper:
http://plugins.netbeans.org/plugin/72557/


dtrebbien:
See #569

PR #571, opened 2017-07-21
Also log the IOException

dtrebbien:
SLF4J has special support for the last argument to a logger call being
an instance of Throwable: https://www.slf4j.org/faq.html#string_or_object


dtrebbien:
See #569

PR #572, opened 2017-07-21
Fix two SLF4J logger calls

dtrebbien:
There were an inconsistent number of formatting anchors ('{}') which
can cause information not to be logged as expected.

These issues were identified by SLF4J Helper for NetBeans:
http://plugins.netbeans.org/plugin/72557/


dtrebbien:
See #569

PR #574, opened 2017-07-24
URL-encode job ID when forwarding from UIController

mprimi:
Motivation:
Download of files from the web UI fails if the job ID contains certain characters (such as '+').
This is due to URL decoding performed on the "forward URL" string returned by UIController (before passing the URL to JobRestController).

Change:
URL-encode the ID contained in the forward string.
This results in the ID being decoded back by Spring MVC to it's original format, and passed intact to the downstream JobRestController.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/controllers/UIController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/controllers/UIController.java
@@ -17,6 +17,7 @@
  */
 package com.netflix.genie.web.controllers;
 
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.security.core.Authentication;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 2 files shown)


genie-web/src/main/java/com/netflix/genie/web/controllers/UIController.java
```diff
@@ -74,6 +78,18 @@ public String getIndex(@NotNull final HttpServletResponse response, @Nullable fi
...
+        String encodedId;
+        try {
+            // When forwarding, the downstream ApplicationContext will perform URL decoding.
+            // If the job ID contains special characters (such as '+'), they will be interpreted as url-encoded and
+            // decoded, resulting in an invalid job ID that cannot be found.
+            // To prevent this, always perform URL encoding on the ID before forwarding.
+            encodedId = URLEncoder.encode(id, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            log.warn("Failed to encode id: {}", id, e);
+            encodedId = id;
```
tgianos: At this point any reason we wouldn't throw the exception up and return error to client?

mprimi: In practice this will never throw since "UTF-8" is constant and a valid encoding.

PR #575, opened 2017-07-24
URL-encode job ID when forwarding from UIController

mprimi:
Motivation:
Download of files from the web UI fails if the job ID contains certain characters (such as '+').
This is due to URL decoding performed on the "forward URL" string returned by UIController (before passing the URL to JobRestController).

Change:
URL-encode the ID contained in the forward string.
This results in the ID being decoded back by Spring MVC to it's original format, and passed intact to the downstream JobRestController.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/controllers/UIController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/controllers/UIController.java
@@ -27,6 +27,8 @@ import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.validation.constraints.NotNull;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 
 /**
  * Controller for forwarding UI requests.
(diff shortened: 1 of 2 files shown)


genie-web/src/main/java/com/netflix/genie/web/controllers/UIController.java
```diff
@@ -69,9 +71,19 @@ public String getIndex(@NotNull final HttpServletResponse response, @Nullable fi
...
      * @return The forward address to go to at the API endpoint
+     * @throws UnsupportedEncodingException if URL-encoding of the job id fails
      */
     @GetMapping(value = "/file/{id}/**")
-    public String getFile(@PathVariable("id") final String id, final HttpServletRequest request) {
-        return "forward:/api/v3/jobs/" + id + "/" + ControllerUtils.getRemainingPath(request);
+    public String getFile(
+        @PathVariable("id") final String id,
+        final HttpServletRequest request
+    ) throws UnsupportedEncodingException {
```
tgianos: We don't have an exception mapper for this exception do you know what error code it maps to by default? 500 I assume?

PR #580, opened 2017-08-02
Metrics review and documentation

mprimi:


mprimi:
@tgianos Before I update the tests (build is going to fail), what do you think of this?

tgianos:
@mprimi I like it! It's the right thing to do for simplicity and to save clutter in the code.

Once you merge this though we'll need to update the internal dashboards

PR #581, opened 2017-08-03
Metrics review, improvement, documentation

mprimi:


tgianos:
@mprimi if I understand you correctly we should still keep the timer in the finally block and instead tag it with success or failure and in the case of failure put another tag of the exception class. We can always break counters out of timers so an additional counter is redundant.

mprimi:
@tgianos I still need to fix a bunch of tests to check for proper metrics tagging (see last commit), however would you mind taking a look if so far the refactoring/tagging/documentation makes sense?

In particular, do you think it would make sense to tag **all** timers with tag `status` and `exceptionClass` like I did for some?
Or is there any metric where you'd like more tags added (the easiest way for this is looking at the new documentation file I added.

genie-docs/src/docs/asciidoc/_metrics.adoc
```diff
@@ -6,13 +6,7 @@
...
 |nanoseconds
 |JobLauncher
-|-
-
-|genie.jobs.submit.exception
-|Counts the number of jobs that failed to initialize and launch
-|count
-|JobLauncher
-|exceptionClass
+|status, exceptionClass
```
tgianos: Do we want the tag to be `exceptionClass` or would `exception` be sufficient. Just to save storage space in Atlas those few characters over and over can save a bunch of space.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ApplicationTask.java
```diff
@@ -155,12 +166,26 @@ public void executeTask(
                         );
                         fts.getFile(configFile, localPath);
                     }
+
+                    final long applicationFinish = System.nanoTime();
+                    this.registry.timer(
+                        applicationTimerId
+                            .withTag(MetricsConstants.TagKeys.APPLICATION_ID, applicationId)
```
tgianos: Do you think it's redundant to do the name and the id here? Are you just trying to provide a way to group by easier? Any worry about cardinality?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/GenieBaseTask.java
```diff
@@ -41,6 +42,13 @@
 @Slf4j
 public abstract class GenieBaseTask implements WorkflowTask {
 
+    protected static final String NO_ID_FOUND = "<no id>";
```
tgianos: Don't we already have something like this somewhere? Seems like it would be common across the whole app

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/GenieBaseTask.java
```diff
@@ -41,6 +42,13 @@
...
+    protected final Registry registry;
```
tgianos: I'd recommend using a getter rather than Protected here. If you extend a task twice you lose access to protected. Get is cleaner.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/GenieBaseTask.java
mprimi: This was duped in a few places, now it's here and used by various tasks subclasses

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/GenieBaseTask.java
mprimi: Fixed

mprimi:
@tgianos I still need to review this myself, but it should be pretty close to its final version

genie-core/src/main/java/com/netflix/genie/core/jobs/JobLauncher.java
```diff
@@ -85,45 +80,35 @@ public JobLauncher(
...
+                this.jobRequest,
+                this.cluster,
+                this.command,
+                this.applications,
+                this.memory
+            );
+        } catch (final GenieException e) {
+            log.error("Unable to submit job due to exception: {}", e.getMessage(), e);
+            MetricsConstants.addFailureTagsWithException(tags, e);
+        } catch (Throwable t) {
```
tgianos: For this should you just re-wrap it in a GenieServerException before throwing it again? Any reason to propagate a random runtime exception? Also should it be final?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ApplicationTask.java
```diff
@@ -70,6 +75,7 @@ public void executeTask(
         @NotNull
         final Map<String, Object> context
     ) throws GenieException, IOException {
+        final Map<String, String> tags = MetricsConstants.newSuccessTagsMap();
```
tgianos: This doesn't look like a "constant" to me. More like a utility method? Perhaps we need a MetricsUtils class.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/GenieBaseTask.java
```diff
@@ -41,6 +42,13 @@
 @Slf4j
 public abstract class GenieBaseTask implements WorkflowTask {
 
+    protected static final String NO_ID_FOUND = "<no id>";
+    private final Registry registry;
+
+    protected GenieBaseTask(final Registry registry) {
```
tgianos: Any reason for protected vs. public for these methods?

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
```diff
@@ -53,15 +54,16 @@
...
 
     /**
      * Constructor.
      *
      * @param registry The metrics registry to use
      */
     public InitialSetupTask(@NotNull final Registry registry) {
-        this.timer = registry.timer("genie.jobs.tasks.initialSetupTask.timer");
+        super(registry);
+        this.timerId = registry.createId("genie.jobs.tasks.initialSetupTask.timer");
```
tgianos: should use the getter here.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobFailureAndKillHandlerLogicTask.java
```diff
@@ -39,15 +40,16 @@
...
 
     /**
      * Constructor.
      *
      * @param registry The metrics registry to use
      */
     public JobFailureAndKillHandlerLogicTask(@NotNull final Registry registry) {
-        this.timer = registry.timer("genie.jobs.tasks.jobFailureAndKillHandlerLogicTask.timer");
+        super(registry);
+        this.timerId = registry.createId("genie.jobs.tasks.jobFailureAndKillHandlerLogicTask.timer");
```
tgianos: getter.

tgianos: Or you might just want to add Utility methods to the base class or metrics utils? I mean the Registry is one global registry so perhaps something like a RegistryUtils we pass around? Dunno.

genie-core/src/main/java/com/netflix/genie/core/jobs/JobLauncher.java
mprimi: Wrapping would change the behavior, e.g. if it's a `RuntimeException` it would have propagated before, and it should be propagated now, I'm just catching to tag the timer and then letting it go as it was before.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
mprimi: Done

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ApplicationTask.java
mprimi: Done

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/GenieBaseTask.java
mprimi: No particular reason, but it's only meant for subclasses

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/JobFailureAndKillHandlerLogicTask.java
mprimi: Done

PR #585, opened 2017-08-10
Modify database cleanup task to delete jobs in configurable batches

tgianos:
This PR adds the ability for system admins to configure the number of jobs deleted by the `DatabaseCleanupTask` at a time. The system will loop deleting that many jobs per iteration until all the jobs that were created before the configured retention time are deleted.

This will prevent the system from being overwhelmed when large numbers of jobs are present for given dates. In Netflix when we would have 250k+ jobs per day sometimes this task would fail.

Effectively pages the deletion process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,14 @@ configure(javaProjects) {
                 entry "retrofit"
                 entry "converter-jackson"
             }
-            dependency("com.netflix.spectator:spectator-api:0.55.0")
+            dependencySet(group: "com.netflix.servo", version: "0.12.17") {
+                entry "servo-core"
+                entry "servo-internal"
+            }
+            dependencySet(group: "com.netflix.spectator", version: "0.56.0") {
+                entry "spectator-api"
+                entry "spectator-reg-servo"
+            }
             dependencySet(group: "io.springfox", version: "2.7.0") {
                 entry "springfox-swagger2"
                 entry "springfox-swagger-ui"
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -82,7 +82,14 @@ configure(javaProjects) {
                 entry "retrofit"
                 entry "converter-jackson"
             }
-            dependency("com.netflix.spectator:spectator-api:0.55.0")
+            dependencySet(group: "com.netflix.servo", version: "0.12.17") {
```
mprimi: Thank you for splitting this off 😍

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaJobPersistenceServiceImpl.java
```diff
@@ -294,12 +299,86 @@ public void setJobCompletionInformation(
...
-        this.jobMetadataRepository.deleteByIdIn(ids);
-        this.jobRepo.deleteByIdIn(ids);
-        return jobRequestRepo.deleteByIdIn(ids);
+    public long deleteAllJobsCreatedBeforeDate(@NotNull final Date date, @Min(1) final int batchSize) {
+        log.info("Attempting to delete all jobs created before {} ms from epoch", date.getTime());
+        long jobExecutionsDeleted = 0;
+        long jobMetadatasDeleted = 0;
+        long jobsDeleted = 0;
+        long jobRequestsDeleted = 0;
+        // Grab the first 10,000 jobs
```
mprimi: Comment is slightly out of date

PR #586, opened 2017-08-10
Improvements to S3FileTransfer and tests

mprimi:
Refactor S3FileTransferImpl to use AmazonS3URI to parse S3 paths.
Plus perform additional path validation as per:
 - http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys
 - http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules
Revamp the tests to cover the new code and logic.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/services/impl/S3FileTransferImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/impl/S3FileTransferImpl.java
@@ -18,22 +18,23 @@
 package com.netflix.genie.core.services.impl;
 
 import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3URI;
 import com.amazonaws.services.s3.model.AmazonS3Exception;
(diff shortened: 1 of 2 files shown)


genie-core/src/main/java/com/netflix/genie/core/services/impl/S3FileTransferImpl.java
```diff
@@ -71,8 +73,13 @@ public S3FileTransferImpl(@NotNull final AmazonS3 amazonS3Client, @NotNull final
...
         log.debug("Called with file name {}", fileName);
-        final Matcher matcher = this.s3PrefixPattern.matcher(fileName);
-        return matcher.matches();
+        try {
+            getS3Uri(fileName);
+            return true;
+        } catch (GenieServerException e) {
+            log.error("Invalid S3 path {} ({})", fileName, e.getMessage());
+        }
+        return false;
```
tgianos: Seems like you should just put the false inside the catch as there is no other way to out of it

PR #589, opened 2017-08-15
Don't save requests against API resources when OAuth2 security enabled

tgianos:
Previously requests against resources (API's) which were supposed to be stateless were creating (and saving) sessions when authentication failed in case the user had to be redirected after authentication. This shouldn't happen for stateless API access. Only SAML authentication should deal with sessions and if UI is accessing the API's the session will be restated by `Spring Session` before reaching OAuth2 resource filter and bypass OAuth authentication.


tgianos:
@twicksell fyi

PR #590, opened 2017-08-17
Make S3 strict URL checking optional and disabled by default

mprimi:
By default, turn off strict S3 URL checking and let arbitrary buckets/keys through.
Strict checking can be enabled via newly introduced property.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/properties/S3FileTransferProperties.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/properties/S3FileTransferProperties.java
@@ -1,0 +1,34 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


mprimi:
@tgianos you mentioned a counter, I can add it, but what do you think we should count?
The number of URLs that we let through that we let through and fail strict validation?

genie-core/src/main/java/com/netflix/genie/core/properties/S3FileTransferProperties.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Properties for S3FileTransfer.
+ *
+ * @author mprimi
+ * @since 3.1.0
+ */
+@Getter
```
tgianos: Could just implement `@Data`

mprimi: None of the other properties class in this package is doing this, so I think i should change all of them or none. Worth doing?

PR #591, opened 2017-08-21
Split database cleanup of jobs into multiple transactions

mprimi:
The strategy of deleting a very large number of jobs with a single paginated transaction can result in timeouts if the number of jobs to delete is large.
Split the deletion in multiple transactions (with a configurable max per transaction). Each transaction still paginates (with a configurable page size).

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/services/JobPersistenceService.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/JobPersistenceService.java
@@ -126,11 +126,15 @@ public interface JobPersistenceService {
     ) throws GenieException;
 
     /**
-     * This method will delete all jobs whose created time is less than date.
+     * This method will delete a chunk of jobs whose creation time is earlier than the given date.
      *
-     * @param date      The date before which all jobs should be deleted
-     * @param batchSize The maximum number of jobs that should be deleted per iteration
+     * @param date       The date before which all jobs should be deleted
+     * @param maxDeleted The maximum number of jobs that should be deleted
+     *                   (soft limit, can be rounded up to multiple of page size)
+     * @param pageSize   Page size used to iterate through jobs
      * @return the number of deleted jobs
      */
-    long deleteAllJobsCreatedBeforeDate(@NotNull final Date date, @Min(1) final int batchSize);
+    long deleteBatchOfJobsCreatedBeforeDate(@NotNull final Date date,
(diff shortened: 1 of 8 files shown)


genie-core/src/test/java/com/netflix/genie/core/jpa/services/JpaJobPersistenceImplIntegrationTests.java
```diff
@@ -101,7 +101,7 @@ public void canDeleteJobsCreatedBeforeDateWithLargeBatchSize() {
         cal.set(2016, Calendar.JANUARY, 1, 0, 0, 0);
         cal.set(Calendar.MILLISECOND, 0);
 
-        final long deleted = this.jobPersistenceService.deleteAllJobsCreatedBeforeDate(cal.getTime(), 10_000);
+        final long deleted = this.jobPersistenceService.deleteBatchOfJobsCreatedBeforeDate(cal.getTime(), 10_000, 1);
```
tgianos: Looks like you may want a case where the second argument isn't 1?

genie-core/src/main/java/com/netflix/genie/core/services/JobPersistenceService.java
```diff
@@ -126,11 +126,15 @@ void setJobCompletionInformation(
...
-     * @param date      The date before which all jobs should be deleted
-     * @param batchSize The maximum number of jobs that should be deleted per iteration
+     * @param date       The date before which all jobs should be deleted
+     * @param maxDeleted The maximum number of jobs that should be deleted
+     *                   (soft limit, can be rounded up to multiple of page size)
+     * @param pageSize   Page size used to iterate through jobs
      * @return the number of deleted jobs
      */
-    long deleteAllJobsCreatedBeforeDate(@NotNull final Date date, @Min(1) final int batchSize);
+    long deleteBatchOfJobsCreatedBeforeDate(@NotNull final Date date,
```
tgianos: Another thing we could add is a count query we could invoke before this to get an idea of how many records we actually need to delete

genie-core/src/test/java/com/netflix/genie/core/jpa/services/JpaJobPersistenceImplIntegrationTests.java
mprimi: Added 2 more tests

PR #592, opened 2017-08-23
Add missing tag reporting final job status

mprimi:
Restore a metric tag that reports final job status as part of JobCompletionService handling of a completed job.


tgianos:
Done as part of #593

PR #593, opened 2017-08-24
Fix tagging for final job status on metrics

tgianos:
- Improve NPM Build
- Add additional caching to travis to hopefully speed up build

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
@@ -75,6 +75,7 @@ import java.util.stream.Collectors;
 public class JobCompletionService {
 
     private static final String ERROR_SOURCE_TAG = "error";
+    private static final String JOB_FINAL_STATE = "jobFinalState";
     private final JobPersistenceService jobPersistenceService;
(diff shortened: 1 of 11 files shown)


tgianos:
@cabhishek have a look at the `genie-web/build.gradle` and let me know if you see any potential issues with the changes I made regarding the UI npm build procees

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -200,8 +193,11 @@ private Job getJob(final String jobId) throws GenieException {
         return this.jobSearchService.getJob(jobId);
     }
 
-    private Void updateJob(final Job job, final JobFinishedEvent event)
-        throws GenieException {
+    private Void updateJob(
+        final Job job,
+        final JobFinishedEvent event,
+        final Map<String, String> tags
```
ajoymajumdar: Isnt it better to return the status of the job rather than sending in the map?

tgianos: That might be but I suspect he (@mprimi) put it in this way in case he needed to provide more tags within the method.

PR #602, opened 2017-09-08
Refactor job dependencies cleanup and include Command dependencies

mprimi:


genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -376,79 +376,108 @@ private JobStatus updateFinalStatusForJob(final String id) throws GenieException
...
-                }
+            final Collection<File> dependencyDirectories = Sets.newHashSet();
 
-            } catch (Exception e) {
-                log.error("Could not delete job dependencies after completion for job: {} due to error {}",
-                    jobId, e);
+            // Collect application dependencies
+            try {
+                for (Application application : this.jobSearchService.getJobApplications(jobId)) {
+                    final Optional<String> appId = application.getId();
```
tgianos: I think you can condense this to `application.getId().ifPresent(lambda)`

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -376,79 +376,108 @@ private JobStatus updateFinalStatusForJob(final String id) throws GenieException
...
-            } catch (Exception e) {
-                log.error("Could not delete job dependencies after completion for job: {} due to error {}",
-                    jobId, e);
-                incrementErrorCounter("DELETE_CLUSTER_DEPENDENCIES_FAILURE", e);
+            // Delete all dependencies
+            for (File dependencyDirectory : dependencyDirectories) {
+                try {
+                    deleteDependenciesDirectory(dependencyDirectory);
+                } catch (IOException e) {
+                    log.error(
```
tgianos: So you have a counter above for getting the files but you don't have an error counter if we actually get an error here?

mprimi: The counter is inside `deleteDependenciesDirectory()`, however since this is now the only place that this is called from, I'm going to inline it so all counters and logic are in one self contained method

PR #610, opened 2017-09-24
DB Standardization

tgianos:
This pull request focusses on making all possible databases use the same schema and get rid of any inconsistencies. This will help make upgrades far simpler.

This PR includes:
* Upgrade Spring Boot to 1.5.7
* Upgrade Gradle to 4.2
* Change Database for consistency by providing upgrade scripts and new schema files
* Add [Flyway](https://flywaydb.org/) to support future database migrations
* Add schema for HSQLDB so not rely on Hibernate auto schema generation
* Update release documentation

Commits are separate for each of these changes and should be fast-forward merged not squashed to keep isolation.


tgianos:
@ajoymajumdar @mprimi This PR is ready for review and once merged should provide the final commits for [3.2.0](https://github.com/Netflix/genie/milestone/27) which could then be released.

genie-ddl/src/main/sql/mysql/schema.mysql.sql
```diff
@@ -120,16 +121,16 @@ CREATE TABLE `clusters` (
...
   `setup_file` varchar(1024) DEFAULT NULL,
   `status` varchar(20) NOT NULL DEFAULT 'OUT_OF_SERVICE',
   `entity_version` int(11) DEFAULT '0',
   PRIMARY KEY (`id`),
   KEY `CLUSTERS_NAME_INDEX` (`name`),
-  KEY `CLUSTERS_TAG_INDEX` (`tags`),
-  KEY `CLUSTERS_STATUS_INDEX` (`status`)
-) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+  KEY `CLUSTERS_STATUS_INDEX` (`status`),
+  KEY `CLUSTERS_TAGS_INDEX` (`tags`(3072))
```
mprimi: This tags index specifies `KEY_BLOCK_SIZE`, unlike all other indices in this schema. Seems a bit strange, could this be left to default?

genie-docs/src/docs/asciidoc/_installation.adoc
```diff
@@ -97,80 +97,70 @@ Genie also ships with database schema scripts for MySQL and PostgreSQL. You will
...
 [source,bash]
 ----
-mysql -u {username} -p{password} -h {host} genie < upgrade-3.0.1-to-3.1.0.mysql.sql
+mysql -u {username} -p{password} -h {host} genie < upgrade-3.1.x-to-3.2.0.mysql.sql
 ----
 
 ===== PostgreSQL
 
-NOTE: This assumes the PSQL binaries are installed
+NOTE: This assumes the PSQL binaries are installed]
```
mprimi: Stray `]`

genie-ddl/src/main/sql/mysql/schema.mysql.sql
tgianos: I left it explicit here because it will flag when anyone running mysql 5.6 doesn't have the proper parameters setup. Which reminds me I need to add that to the documentation. In other words if they're running 5.6 with defaults and we say `INDEX(tags)` their index would only be 767 bytes long or whatever the default is without large_prefix_enabled set.

genie-docs/src/docs/asciidoc/_installation.adoc
tgianos: Good catch will fix.

PR #613, opened 2017-09-26
Explicitly set the lang

rgbkrk:
This likely won't fix the current `laquo` issue, wise to do in the meantime.


rgbkrk:
The issue that is happening is that the character set for genie on test is somehow being set to `windows-1252`.

PR #614, opened 2017-09-27
Globally force UTF-8 encoding of server responses

mprimi:
To avoid Tomcat falling back to the default 'ISO-8859-1' charset [1], force content-type to be UTF-8 from within Spring

[1] https://wiki.apache.org/tomcat/FAQ/CharacterEncoding


mprimi:
@tgianos this sets UTF-8 **globally**. 
It is not strictly necessary since from what I'm seeing most, if not all, requests already detect and correctly set utf-8. The exception is the static resource `index.html`.
So if we didn't want to do this, we could instead force content-type just for that GetMapping.
WDYT?

tgianos:
I'm certainly not an expert. I doubt there's a problem with our REST API returns other than the data returns from the `output/**` endpoints. If someone were to write non-utf encoded data we may screw that up?

mprimi:
@tgianos ready for review

PR #618, opened 2017-10-03
Revert to ThymeLeaf template for serving index.html in Genie UI

mprimi:
Motivation: Removing ThymeLeaf and serving index.html as a static web resource caused a loop in servlet dispatching which made it impossible to browse job files from the web UI.
To properly fix this, some refactoring around controllers, views and their configuration may be necessary.
For the sake of unblocking the current release, revert to templates and come back to properly fix this later.

 * Re-introduce ThymeLeaf dependency
 * Move back index.html into resource templates directory
 * Update UIController to reflect the above
 * Update tests


mprimi:
@tgianos if we do master first we can see this working in `dev` (already tested on my stack, but just in case). Then we can cherry-pick

PR #620, opened 2017-10-04
Make command and cluster criteria tags available to job as environmen…

mprimi:
…t variables

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
@@ -19,7 +19,9 @@ package com.netflix.genie.core.jobs.workflow.impl;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.netflix.genie.common.dto.Cluster;
+import com.netflix.genie.common.dto.ClusterCriteria;
 import com.netflix.genie.common.dto.Command;
+import com.netflix.genie.common.dto.JobRequest;
 import com.netflix.genie.common.exceptions.GenieException;
 import com.netflix.genie.common.exceptions.GeniePreconditionException;
 import com.netflix.genie.common.exceptions.GenieServerException;
@@ -38,6 +40,7 @@ import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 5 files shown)


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
```diff
@@ -376,6 +382,60 @@ void createJobEnvironmentVariables(
...
+        for (ClusterCriteria clusterCriteria : clusterCriterias) {
+            clusterCriteriasStrings.add("[" + tagsToString(clusterCriteria.getTags()) + "]");
+        }
+
+        // Append new line
+        writer.write(System.lineSeparator());
+
+        // create environment variable for the list of cluster tags/criteria in the job request as a single
+        // value in the form "[[x,y,z],[a,b,c]]"
+        writer.write(JobConstants.EXPORT
```
tgianos: would it make more sense to use a format that's already known like JSON?

Additionally perhaps for scale (and I think we have a JIRA for this already) a better feature would be instead of continually dumping environment variables we just write out JSON dumps of the various components (job request, cluster, command, applications) to disk as files and then users can do whatever they want with them. Then we don't have to change all the time or worry about this formatting. Cause there is every chance this could eventually have some bash issues with quotes etc.

mprimi: This is JSON-ish (it's missing the quotes around strings because we'd need to escape them, making it harder to use this variable from bash.

Are you suggesting removing these 2 variables and instead dumping the entire job request as JSON, in this patch?

tgianos: I'm wondering out loud if rather than doing this and then having @rdblue and others write scripts to this specification and then they decide they want something else and we go to another if it wouldn't just be worth it to short circuit all this and give as much info as possible up front. Not saying we have to do it but to me that seems to be the most scalable / flexible.

tgianos: Plus if we dump all the info into files as JSON every possible language would know how to parse it whereas this format needs to be parsed specifically

rdblue: Right now, I don't consider this a specification. My code to match is:

```
if echo $COMMAND_TAGS | grep 'data:test'; then
  # test metastore
else
  # prod metastore
fi
```

Whatever you guys want to define as the convention is fine with me. I think the original request was to just dump the request JSON to a file. That would work.

PR #626, opened 2017-10-12
Develop

mprimi:
PoC and examples for Protobuf, gRPC


mprimi:
@tgianos this is reviewable (no rush since it's standalone PoCs). Each commit can be looked at independently.

build.gradle
```diff
@@ -105,18 +106,33 @@ configure(javaProjects) {
             dependency("org.opensaml:opensaml:2.6.4")
             dependency("org.python:jython-standalone:2.7.1")
             dependency("org.springframework.security.extensions:spring-security-saml2-core:1.0.2.RELEASE")
+            dependencySet(group: "com.google.protobuf", version: "3.4.0") {
```
tgianos: (OCD) Can you put these in alphabetical order with all the rest?

tgianos: Also it seems there's already a version of `protobuf` managed by the [Spring platform](https://docs.spring.io/platform/docs/Brussels-SR5/reference/htmlsingle/#appendix-dependency-versions) should we just go with that or is there a reason you bumped it to 3.4.0. I see that version is 2.6.1. There is a standard way to override Spring platform versions and that's to put a property probably something like `protobuf.version` to override it.

build.gradle
```diff
@@ -105,18 +106,33 @@ configure(javaProjects) {
...
+              entry "protoc-gen-grpc-java"
+              entry "grpc-testing"
+            }
         }
     }
 
     ext {
         generatedSourcesDir = new File("${projectDir}/src/generated")
         generatedSourcesJavaDir = new File(generatedSourcesDir, "/java")
+        // Necessary due to limitations of proto plugin (${source-set-name}/${variant} is always appended)
```
tgianos: Why does this have to be in `src/generated/main/java`?  why not just `src/generated/java` like the other one? I might be missing the point of this comment. Seems to conflict with what you have down in the plugin configuration below

genie-proto/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+        // Download the protoc compiler
+        artifact = 'com.google.protobuf:protoc'
+        // Use project convention for generated files
+        generatedFilesBaseDir = "${projectDir}/src/generated"
+    }
+
+    plugins {
+        grpc {
+            // Download the grpc compiler
+            artifact = 'io.grpc:protoc-gen-grpc-java'
```
tgianos: Recommend consistency in `"` vs `'`

genie-web/src/main/java/com/netflix/genie/grpc/services/impl/GreeterServiceImpl.java
```diff
@@ -0,0 +1,42 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: Wrong year.

You can configure intellij to use a template for these that have the year as a variable.

genie-web/src/main/java/com/netflix/genie/grpc/services/impl/package-info.java
```diff
@@ -0,0 +1,25 @@
+/*
+ *
+ *  Copyright 2015 Netflix, Inc.
```
tgianos: Really wrong year :-)

genie-web/src/test/java/com/netflix/genie/grpc/services/impl/GreeterServiceImplUnitTest.java
```diff
@@ -0,0 +1,71 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: year

genie-web/src/test/java/com/netflix/genie/grpc/services/impl/package-info.java
```diff
@@ -0,0 +1,25 @@
+/*
+ *
+ *  Copyright 2015 Netflix, Inc.
```
tgianos: year.

genie-core/src/test/java/com/netflix/genie/core/grpc/poc/GreeterClient.java
```diff
@@ -0,0 +1,103 @@
...
+        final HelloReply response = blockingStub.sayHello(request);
+        return response.getMessage();
+    }
+
+    /**
+     * Create a client and receive a greeting from the server.
+     * @param args CLI arguments: host name, port, user name
+     * @throws Exception in case of error
+     */
+    public static void main(final String[] args) throws Exception {
```
tgianos: A main in test classes?

genie-test/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+ -->
+ <configuration>
```
tgianos: Was something wrong with how boot was doing it by default? Can't we change the logging levels via the standard spring properties `logging.level.{package}`

build.gradle
```diff
@@ -118,6 +118,10 @@ configure(javaProjects) {
               entry "protoc-gen-grpc-java"
               entry "grpc-testing"
             }
+            dependencySet(group: "net.devh", version: "1.3.0-RELEASE") {
```
tgianos: This is the one @twicksell recommended right?

genie-core/src/main/java/com/netflix/genie/core/grpc/poc/SimpleLoggingInterceptor.java
```diff
@@ -0,0 +1,67 @@
+/*
+ *
+ *  Copyright 2015 Netflix, Inc.
```
tgianos: Need to check all your years in the headers.

genie-core/src/test/java/com/netflix/genie/core/grpc/poc/HeartbeatExample.java
```diff
@@ -0,0 +1,344 @@
...
+/**
+ * Heartbeat protocol example.
+ * Server randomly shuts down.
+ * Client enqueues heartbeats at fixed rate and handles disconnections/retries.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public final class HeartbeatExample {
```
tgianos: cool little PoC 👍

build.gradle
twicksell: The upcoming Spring 5 and Boot 2 supports proto 3.x, and recommends 3.4.0. https://github.com/spring-io/platform/blob/d5f74380022d67d03b02feebf30963bc52913f40/platform-bom/pom.xml#L136

If you're a ways out from releasing this, it may be worth coding to the Boot 2.x milestone releases.

tgianos: Agreed that we will be targeting boot 2.x. We're going to have to follow the upgrade steps though not part of this: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Release-Notes

genie-test/src/test/resources/logback-test.xml
mprimi: Oh, didn't think of that. 
However this applies more broadly to **all** tests.
Netty and gRPC are extremely verbose so i wanted to shut them down.

PR #627, opened 2017-10-13
Fix area of possible NPE in response mapping interceptor for the java client

tgianos:
It was reported that sometimes if the server or proxy (e.g. ELB) returned an error response without a body that a NPE could be thrown out of the response mapper instead of the expected `GenieClientException`. This PR should fix that issue.

### DIFF
--- a/genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
+++ b/genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
@@ -23,17 +23,20 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.genie.client.exceptions.GenieClientException;
 import okhttp3.Interceptor;
 import okhttp3.Response;
+import okhttp3.ResponseBody;
 
 import java.io.IOException;
+import java.io.Reader;
 
 /**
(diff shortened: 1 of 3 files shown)


genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
```diff
@@ -53,17 +56,24 @@ public Response intercept(final Chain chain) throws IOException {
...
-                final JsonNode responseBody = this.mapper.readTree(bodyContent);
-                throw new GenieClientException(
-                    response.code(),
-                    response.message() + " : " + responseBody.get("message")
-                );
-            } catch (final JsonProcessingException jpe) {
-                throw new GenieClientException(response.code(), response.message() + bodyContent);
+                if (bodyReader != null) {
+                    try {
+                        final JsonNode responseBody = this.mapper.readTree(bodyReader);
```
skwslide: Shouldn't it check for the possibility that `responseBody` is null?

tgianos: Yeah. The documentation on the `readTree` method was a little vague but better safe than sorry. Will fix.

PR #631, opened 2017-10-21
[WIP] Aspect to monitor turnaround time of healthchecks

mprimi:


genie-web/src/main/java/com/netflix/genie/web/controllers/JobSearchAspect.java
```diff
@@ -0,0 +1,122 @@
...
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+@Aspect
+@Component
+@Slf4j
+public class JobSearchAspect {
```
tgianos: Class name seems wrong.

genie-web/src/main/java/com/netflix/genie/web/controllers/JobSearchAspect.java
```diff
@@ -0,0 +1,122 @@
...
+        );
+        final Map<String, String> tags;
+        if (throwable == null) {
+             tags = MetricsUtils.newSuccessTagsMap();
+        } else {
+            tags = MetricsUtils.newFailureTagsMapForException(throwable);
+        }
+        tags.put(HEALTH_INDICATOR_NAME, joinPoint.getTarget().getClass().getSimpleName());
+        registry
+            .timer(registry.createId("health-check.indicator.timer", tags))
```
tgianos: don't like this name. Haven't we followed `genie.{metricClass}.{name}.{type}` kind of naming syntax for all our other ones? I like at least having `genie` at the front to indicate that it's something we've created rather than something coming from Boot or other system

genie-web/src/main/java/com/netflix/genie/web/controllers/JobSearchAspect.java
```diff
@@ -0,0 +1,122 @@
...
+            turnaround,
+            throwable != null ? throwable.getClass().getSimpleName() : "none"
+        );
+        final Map<String, String> tags;
+        if (throwable == null) {
+             tags = MetricsUtils.newSuccessTagsMap();
+        } else {
+            tags = MetricsUtils.newFailureTagsMapForException(throwable);
+        }
+        tags.put(HEALTH_INDICATOR_NAME, joinPoint.getTarget().getClass().getSimpleName());
```
tgianos: For this make sure we're not getting back one of the Spring wrapper classes as the names. Sometimes you need to cast it to the true class the spring aspect class is wrapping.

genie-web/src/main/java/com/netflix/genie/web/controllers/JobSearchAspect.java
mprimi: Yeah started as POC, definitely need to rename and relocate

genie-web/src/main/java/com/netflix/genie/web/controllers/JobSearchAspect.java
mprimi: From local testing it looks right (i.e. no need for ClassAware)

genie-web/src/main/java/com/netflix/genie/web/controllers/JobSearchAspect.java
mprimi: Yes this was a placeholder, going to change to follow convention

mprimi:
See https://github.com/Netflix/genie/pull/632

PR #632, opened 2017-10-24
Health metrics and disable non-essentials checks

mprimi:
Publish timing and failure metrics related to healtcheck. 
Disable two non-essential healtchchecks.

### DIFF
--- a/genie-docs/src/docs/asciidoc/_metrics.adoc
+++ b/genie-docs/src/docs/asciidoc/_metrics.adoc
@@ -410,6 +410,30 @@ and system metrics and statistics.
 |DiskCleanupTask
 |-
 
+|genie.health.endpoint.timer
+|Time taken for the Health endpoint to collect and aggregate state from health indicators
+|nanoseconds
+|HealthCheckMetricsAspect
+|status, exceptionClass
+
+|genie.health.indicator.timer
+|Time taken for each health indicator to report its status
+|nanoseconds
+|HealthCheckMetricsAspect
+|status, exceptionClass, healthIndicatorClass
+
+|genie.health.indicator.counter
+|Counter for of calls each health indicator tagged by reported status
+|count
+|HealthCheckMetricsAspect
+|status, exceptionClass, healthIndicatorName
+
+|genie.health.failure.counter
+|Counter for calls each health indicator that reports status other than UP
+|count
+|HealthCheckMetricsAspect
+|status, exceptionClass, healthIndicatorName
+
 |===
 
 (*) Source may add additional tags on a case-by-case basis
--- a/genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
@@ -1,0 +1,248 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.web.aspect;
+
+import com.google.common.collect.Maps;
+import com.netflix.genie.core.util.MetricsConstants;
+import com.netflix.genie.core.util.MetricsUtils;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
+import lombok.extern.slf4j.Slf4j;
+import org.aspectj.lang.JoinPoint;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Before;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.health.Health;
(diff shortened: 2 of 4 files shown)


genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * Autowired constructor.
+     * @param registry metrics registry
+     */
+    @Autowired
+    public HealthCheckMetricsAspect(final Registry registry) {
+        this.registry = registry;
+        this.healthEndpointTimerId = registry.createId(HEALTH_ENDPOINT_TIMER_NAME);
+        this.healthIndicatorTimerId = registry.createId(HEALTH_INDICATOR_TIMER_METRIC_NAME);
+        this.healthIndicatorCounterId = registry.createId(HEALTH_INDICATOR_COUNTER_METRIC_NAME);
```
tgianos: Timers automatically include a counter metric. Creating a separate one is redundant.

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
```diff
@@ -0,0 +1,248 @@
...
+        this.healthIndicatorFailureCounterId = registry.createId(HEALTH_FAILURES_COUNTER_METRIC_NAME);
```
tgianos: Any reason not to just tag the timer with success or failure or similar?

mprimi: Your two questions are related.
The join point that is able to capture the indicator time to complete, and publish the timer metric, does not have access to the status reported by the healtcheck. So it can only publish how long it took to evaluate.

The other counter metric has access to status (UP/DOWN/OUT_OF_SERVICE), etc, but does not have access to timing (since the join point is aggregation of the final status, after all indicators have been evaluated already).
Because of this, I cannnot publish a single metric that has both timing, count and status.

I have to publish timing with a timer and separately status with a counter.

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
```diff
@@ -0,0 +1,248 @@
...
+                    .counter(healthIndicatorCounterId.withTags(tags))
+                    .increment();
+
+                //Count failures (if a healthcheck passees, this counter is incremented by 0), publish tagged with
+                // name and status.
+                // Good for alerting, any aggregate that is greater than zero for some period of time signals a
+                // check that is consistently failing.
+                registry
+                    .counter(healthIndicatorFailureCounterId.withTags(tags))
+                    .increment(isUp ? 0 : 1);
```
tgianos: Increment 0 just makes it so that the value doesn't increase. So are you saying that your rate becomes 0 not your absolute value?

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
tgianos: So you want the timer as well so that we can alert on long running checks?

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
```diff
@@ -0,0 +1,248 @@
...
+        } catch (final Throwable t) {
+            log.warn(
+                "Failed to cast AbstractHealthAggregator health details argument: {}",
+                joinPoint.getArgs(),
+                t
+            );
+            return;
+        }
+
+        healthDetailsMap.forEach(
```
tgianos: Don't we already have this via the lower level intercepts you have? (I'm forgetting how we discussed this before I left) I remember having a reason to do this at the global level but not sure why you're also doing it at the lower level. Seems like the top level one would be enough?

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
mprimi: Yes if HCs start timing out this should give us all the info to see at glance what is happening.

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
mprimi: Not sure what you mean by top level vs. lower level

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
mprimi: This metric counts the number of failed healtchecks, I publish zeroes so that the metric is present even if there are no failures, making it easy to see that queries/graphs/alerts are working.

genie-web/src/main/java/com/netflix/genie/web/aspect/HealthCheckMetricsAspect.java
tgianos: I believe we discussed this in person but I meant the aggregator (top) vs. the individual (lower) levels. I think we're good now

genie-docs/src/docs/asciidoc/_metrics.adoc
```diff
@@ -410,6 +410,30 @@ and system metrics and statistics.
...
+|Time taken for each health indicator to report its status
+|nanoseconds
+|HealthCheckMetricsAspect
+|status, exceptionClass, healthIndicatorClass
+
+|genie.health.indicator.counter
+|Counter for of calls each health indicator tagged by reported status
+|count
+|HealthCheckMetricsAspect
+|status, exceptionClass, healthIndicatorName
```
tgianos: Just minor that you're using `healthIndicatorName` as a tag in these latter two and `healthIndicatorClass` in the previous one. Is there an explicit reason this is different?

mprimi: I wish we could be consistent, but in one case (the timer) all I see is the class.
In the other (the counter) i get a map with name->status.

PR #633, opened 2017-10-25
Fix a few places where exception traces were not being propagated

mprimi:
To aid human debugging and machine monitoring of logs, find and fix all the places where exception are being swallowed, rather than being wrapped into new exceptions thrown by Genie.


mprimi:
Build keeps getting stuck in the docker build image step, but it passed before and it passes locally. Suspecting a Travis issue. Giving it some time.

PR #634, opened 2017-10-30
Cherry-picks from master

mprimi:
* https://github.com/Netflix/genie/pull/622 SLF4J 
* https://github.com/Netflix/genie/pull/633 Missing exception traces
* https://github.com/Netflix/genie/pull/632 Healtchecks


tgianos:
Is this a cherry pick of #622? If so we should probably document that in the title

PR #635, opened 2017-10-30
Disable Redis and Mail health checks

mprimi:
Disable non-essential health checks that can make Genie unavailable (due to removal from load balancer).
Redis is only used in the path of the web UI. Genie instances should not stop accepting jobs if the latter is temporarily unreachable.
Similarly, the mail notification server does not warrant genie reporting unhealthy status.

### DIFF
--- a/genie-web/src/main/resources/application.yml
+++ b/genie-web/src/main/resources/application.yml
@@ -121,6 +121,11 @@ management:
   context-path: /actuator
   security:
     roles: ADMIN
+  health:
+    redis:
+      enabled: false
+    mail:
+      enabled: false
 
 security:
   basic:


genie-web/src/main/resources/application.yml
```diff
@@ -121,6 +121,11 @@ management:
   context-path: /actuator
   security:
     roles: ADMIN
+  health:
+    redis:
+      enabled: false
+    mail:
+      enabled: false
```
tgianos: I'm going back and forth on whether we should disable the mail check. I understand the thought process wherein it is not critical to the job execution itself and is merely something that happens after however if the mail system is down and users are expecting a mail notification and don't get it that could actually be a problem for user workflow. I could be convinced either way but right now I'm kind of leaning on it is in the "critical" path of full job lifecycle as something that occurs during cleanup.

mprimi: Leaving the mail healtcheck active is probably the more cautious thing to do. 
I.e., just in case don't accept new jobs as we may not be able to notify.

However we cannot offer the same guarantee for jobs already in flight. So if we really wanted to make the mail notification robust and reliable, there is more work to be done. Leaving the healtcheck on doesn't really do that much.

On the other side, imagine a case where the mail server has a serious outage. I'd imagine most Genie users would be fine with going ahead without mail, rather than stopping all job processing until mail is back (i.e. imagine saying to an angry user whose jobs are not running: "yes we're still down because there is a mail server outage").

mprimi:
Closing and merging with https://github.com/Netflix/genie/pull/632

PR #639, opened 2017-11-08
Run integration tests against PostgreSQL and MySQL in CI

mprimi:


build.gradle
```diff
@@ -169,6 +169,9 @@ configure(javaProjects) {
             excludeCategories "com.netflix.genie.test.categories.IntegrationTest"
             excludeCategories "com.netflix.genie.test.categories.DocumentationTest"
         }
+        outputs.upToDateWhen {
```
tgianos: doesn't this make the outputs ALWAYS up to date no matter what? Is there a way to only do this when it's actually on the ci server so we can force it? Seems annoying to have local builds always recompile test stuff no matter what

mprimi: This says this task is never up to date: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskOutputs.html#upToDateWhen(groovy.lang.Closure)

These tasks are not part of our builds, I thought they were created for developers to invoke on the command-line. As such, it makes sense to me that they re-run the tests no matter what.

Issue #636, opened 2017-11-06
Errors after upgrading jar from 3.0.8 to 3.2.3


kirubel:
Getting the following errors after upgrading .jar from 3.0.8 to 3.2.3.  The DB was upgraded to 3.2.0 Postgres.  The UI /jobs is populated but not /clusters or /applications.
Below is entry from the error logs

" WARN 45466 --- [nio-8080-exec-8] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 0, SQLState: 22003
2017-11-02 19:22:56.199 ERROR 45466 --- [nio-8080-exec-8] o.h.engine.jdbc.spi.SqlExceptionHelper   : Bad value for type long : Analytics Runtime Framework SIT/Andromeda Submit Command
2017-11-02 19:22:56.206 ERROR 45466 --- [nio-8080-exec-8] c.n.g.w.c.GenieExceptionMapper           : Failed to execute data service method
"
 looks like some data type mismatch between existing data and Spring Hibernate.

Any advice would be appreciated!

tgianos:
Hey @kirubel To verify your upgrade process you were on the postgres schema from 3.0.1?

And then you ran the following files to upgrade:
[3.0.1 to 3.1.0](https://github.com/Netflix/genie/blob/3.2.x/genie-ddl/src/main/sql/postgresql/upgrade-3.0.1-to-3.1.0.postgresql.sql)
[3.1.x to 3.2.0](https://github.com/Netflix/genie/blob/3.2.x/genie-ddl/src/main/sql/postgresql/upgrade-3.1.x-to-3.2.0.postgresql.sql)

Is this true?

tgianos:
can you run `pg_dump` on your database with the following flags:
`pg_dump --no-tablespaces --no-owner --no-privileges --schema-only` and `diff` it to the [expected schema](https://raw.githubusercontent.com/Netflix/genie/v3.2.3/genie-ddl/src/main/sql/postgresql/schema.postgresql.sql) and post any results?

I believe @mprimi tested all this as he was doing the 3.2.x changes but we may have missed something as Postgresql isn't our current production DB

kirubel:
Here are some difference
1) - Dumped from database version 9.6.2
6.	-- Dumped by pg_dump version 9.6.5	6.	-- Dumped by pg_dump version 9.6.4
(output from my env was 9.6.2 for server and 9.6.4 for client node

The following lines were present in my env but missing from the expected schema.
[...]

PR #640, opened 2017-11-13
Annotate '@Lob' entity fields as TextType to aid Hibernate/PostgreSQL…

mprimi:
… driver

The PostgreSQL hibernate driver is failing to load values for fields marked @Lob.
The annotation was introduced when the fields were migrated from varchar to text.
PostgreSQL expect a value reference in this column (a integer-type pointer to the actual column value stored externally), and it fails when inline text is found instead.

The resulting error is:
```
  org.postgresql.util.PSQLException: Bad value for type long : <application description>
```

See also: https://github.com/Netflix/genie/issues/636


mprimi:
Deprecated, see https://github.com/Netflix/genie/pull/641

PR #651, opened 2017-11-20
Merge jobs tables and other database improvements

tgianos:
This PR is quite large due to the fact it needs to be done as one large commit due to changing database structure necessitates changes to many classes particularly in the JPA packages (any other trickle should be the result of exposing new functionality)

This PR:
- Merges the 4 jobs tables (`job_requests`, `jobs`, `job_metadata `and `job_executions`) from previous genie releases into one large `jobs `table
- Normalizes Tag and File references into their own tables `tags` and `files` so they can be reused via mapping tables to other resources like `clusters` or `jobs`
- Makes `command_args` for a job optional and breaks it out into a list of arguments rather than one large string to better support serialization to command line going forward
- Breaks cluster and command criterion into their own table for better insight
- Improves query performance by leveraging several projections so as not to return entire row of data when not necessary
- Completely re-writes how clusters and commands are found for given criteria. Before it was an iterative approach which was inefficient and wasteful. Now it is done via a query which returns a lookup map of cluster `id` to command `id` which can then save iterations and database queries later on
- Added `grouping` and `groupingInstance` fields to a job to allow them to be searched with other jobs they're related to. Also exposed in the find API
- Added tests
- Code structure improvement
- Removed unnecessary classes and files
- Upgraded to Spring Cloud Dalston SR4
- Others

This PR DOES NOT:
- Include support for removal of orphan `tag` and `file` references in the database. Need to come up with a good way to do that
- Still relies on an unreleased version of Spring Data JPA while waiting for [this](https://jira.spring.io/browse/DATAJPA-1185) bug to be released as part of the [Ingalls SR9](https://jira.spring.io/browse/DATAJPA/fixforversion/16386/?selectedTab=com.atlassian.jira.plugins.jira-development-integration-plugin:release-report-tabpanel) Spring data release
- Need to upgrade to Boot 1.5.9 once [Brussels SR6](https://github.com/spring-io/platform/milestone/41) of Spring Platform is released


tgianos:
@ajoymajumdar I'd appreciate you taking a look at some of this as it involves a lot of database schemas and interactions and seeing if you have any feedback

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaTagServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public JpaTagServiceImpl(final JpaTagRepository tagRepository) {
+        this.tagRepository = tagRepository;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+//    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
mprimi: Need to change this right?

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/ApplicationEntity.java
```diff
@@ -50,9 +47,10 @@
  */
 @Getter
 @Setter
+@ToString(callSuper = true)
```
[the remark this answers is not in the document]
tgianos: A `toString()` or the call to super? I added the `toString()` because we tend to log things out a lot so having a nice toString is better than the default one from Java.

Calling super to make sure the members of parent classes are included.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/AuditEntity.java
```diff
@@ -0,0 +1,115 @@
...
+    private Integer entityVersion;
+
+    /**
+     * Updates the created and updated timestamps to be creation time.
+     */
+    @PrePersist
+    protected void onCreateBaseEntity() {
+        final Date date = new Date();
+        this.updated = date;
+        this.created = date;
```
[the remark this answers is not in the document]
tgianos: No. The method is annotated with `@PrePersist` which is only called the first time the entity is saved to the database. Also the created field is annotated with `@Column(name = "created", nullable = false, updatable = false)` and the `updatable=false` makes it so that the value is never included in update queries.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/JobEntity.java
```diff
@@ -140,9 +256,73 @@
             @JoinColumn(name = "application_id", referencedColumnName = "id", nullable = false)
         }
     )
-    @OrderColumn(name = "application_order", nullable = false)
+    @OrderColumn(name = "application_order", nullable = false, updatable = false)
     private List<ApplicationEntity> applications = new ArrayList<>();
 
+    @ManyToMany(fetch = FetchType.LAZY, cascade = CascadeType.ALL)
```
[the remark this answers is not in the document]
tgianos: Because there is no repository for the `Criterion` entities we're only creating and deleting them as part of a given job. The cascade enables the JPA manager to delete all the associated entities. Though our database ddl's do this as well it is good to have this in case some entity state is stuck in the Entity Manager we want to make sure it's flushed out. Better safe than sorry.

genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaClusterRepository.java
```diff
@@ -16,13 +16,87 @@
...
+     *                         e.g. 'tag1','tag2'
+     * @param commandTagsSize  The number of command tags (size of commandTags)
+     * @return The id pairs found or empty list if no matches
+     */
+    @Query(value = CLUSTER_COMMAND_QUERY, nativeQuery = true)
+    List<Object[]> findClustersAndCommandsForCriterion(
+        @Param("clusterTags") final Set<String> clusterTags,
+        @Param("clusterTagsCount") final int clusterTagsCount,
+        @Param("commandTags") final Set<String> commandTags,
+        @Param("commandTagsCount") final int commandTagsSize
```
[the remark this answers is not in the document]
tgianos: That would be nice however I didn't really find a quick way to do that with the parameter filled methods. Could probably look around for something but didn't think it was a big enough deal to go into [SpEL](https://docs.spring.io/spring/docs/3.0.x/spring-framework-reference/html/expressions.html) or something. That may be smart enough to pull out the sizes of the collections but it would really lose compile time checks.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaClusterServiceImpl.java
```diff
@@ -98,19 +117,19 @@ public JpaClusterServiceImpl(
...
-        if (clusterId.isPresent() && this.clusterRepo.exists(clusterId.get())) {
+        if (clusterId.isPresent() && this.clusterRepository.existsByUniqueId(clusterId.get())) {
             throw new GenieConflictException("A cluster with id " + clusterId.get() + " already exists");
         }
 
         final ClusterEntity clusterEntity = new ClusterEntity();
-        clusterEntity.setId(cluster.getId().orElse(UUID.randomUUID().toString()));
-        this.updateAndSaveClusterEntity(clusterEntity, cluster);
-        return clusterEntity.getId();
+        clusterEntity.setUniqueId(cluster.getId().orElse(UUID.randomUUID().toString()));
```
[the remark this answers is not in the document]
tgianos: Hmm... you really want to make an extra DB query just on the minuscule chance? As it is this would throw a 409.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaClusterServiceImpl.java
```diff
@@ -398,69 +423,79 @@ public void addTagsForCluster(
...
+        final Set<TagEntity> tags = this.findCluster(id).getTags();
+        // Remove all the tags except the ones that start with "genie."
+        tags.removeAll(
+            tags
+                .stream()
+                .filter(
+                    tagEntity -> !tagEntity.getTag().startsWith(JpaBaseService.GENIE_TAG_NAMESPACE)
+                )
+                .collect(Collectors.toSet())
+        );
```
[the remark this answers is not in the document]
tgianos: Yes. Because the collection after exiting the method (and thus initiating a transaction commit) will only contain references to the ones it needs it will do a set differential and issue the proper queries.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaTagServiceImpl.java
tgianos: No. We decided to leave it as part of the same transaction to avoid risk of cascading connection requests. The save and flush will push the commit however it will include any changes done up to that point in the method stack (since the transaction was opened by entering the initial aspect wrapped method). I believe I went back through the call stack and ensured that the `Tag` and `File` entities were created before any other changes were made but it's something you could help me double check

genie-common/src/main/java/com/netflix/genie/common/dto/JobRequest.java
```diff
@@ -93,6 +97,17 @@
...
+        this.grouping = builder.bGrouping;
+        this.groupingInstance = builder.bGroupingInstance;
+    }
+
+    /**
+     * Get the arguments to be put on the command line along with the command executable.
+     *
+     * @return The command arguments
+     */
+    public Optional<String> getCommandArgs() {
```
ajoymajumdar: We can abstract out this command args logic into a separate base class.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/ApplicationEntity.java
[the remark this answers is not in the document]
ajoymajumdar: Do you want all the members to be included? It might end up querying all the lazy associations too just for logging.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/ApplicationEntity.java
```diff
@@ -113,60 +120,66 @@ protected void onCreateOrUpdateApplication() throws GenieException {
...
 
     /**
-     * Sets the configurations for this application.
+     * Set all the files associated as configuration files for this application.
      *
-     * @param configs The configuration files that this application needs
+     * @param configs The configuration files to set
      */
-    public void setConfigs(final Set<String> configs) {
+    public void setConfigs(@Nullable final Set<FileEntity> configs) {
```
ajoymajumdar: If updating an attached application entity, then the logic of clear and adding will end up deleting all records and inserting the new ones. Should be fine?

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaApplicationServiceImpl.java
```diff
@@ -96,19 +113,19 @@ public JpaApplicationServiceImpl(
...
     public String createApplication(
         @NotNull(message = "No application entered to create.")
-        @Valid
-        final Application app
+        @Valid final Application app
     ) throws GenieException {
         log.debug("Called with application: {}", app);
         final Optional<String> appId = app.getId();
-        if (appId.isPresent() && this.applicationRepo.exists(appId.get())) {
+        if (appId.isPresent() && this.applicationRepository.existsByUniqueId(appId.get())) {
```
ajoymajumdar: Do we need to check the exists? On save, when you get a DuplicateKeyException, you can propogate the GenieConflictException

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Create a file reference in the database and return the attached entity.
+     *
+     * @param file The file to create
+     * @return The file entity that has been persisted in the database
+     * @throws GenieException on error
+     */
+    FileEntity createAndGetFileEntity(final String file) throws GenieException {
+        this.fileService.createFileIfNotExists(file);
```
ajoymajumdar: Isn't it better to return the FileEntity if exists or return the attached one after creation

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/ApplicationEntity.java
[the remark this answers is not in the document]
tgianos: Good point @ajoymajumdar. I'll add exclusions to the collections

genie-common/src/main/java/com/netflix/genie/common/dto/JobRequest.java
tgianos: ugh. I hate that you're right about this.

genie-core/src/main/java/com/netflix/genie/core/jpa/entities/ApplicationEntity.java
tgianos: I'm not clear on how it works. My assumption was that since this modification will be called inside the scope of a transaction at the end of the transaction only the final state will be translated into the SQL calls not the intermediate changes. I don't think it acts like a write ahead log. I could be wrong but this seemed the easiest way to do a set differential. Open to other ideas?

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaApplicationServiceImpl.java
tgianos: Fair. Saves a query. Will change.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
tgianos: This was written when the `createFileIfNotExists` method still required a new transaction. So I needed to do the get of the attached entity once it returned to the current transaction context since the entity manager would be different for both transactions. Now that it's one transaction I probably could do as you say however leaving it as is may be safer if we have to revert to multiple transactions. Additionally since this `find` would use the current entity manager in the same transaction the entity should already be attached in the scope so it wouldn't go to the database again correct @ajoymajumdar ?

ajoymajumdar: Yes.

genie-common/src/main/java/com/netflix/genie/common/dto/JobRequest.java
tgianos: After looking at it the way the inheritance structure for the DTO's are setup this isn't really possible since `Job` extends from `CommonDTO` and `JobRequest` from `ExecutionEnvironmentDTO`. If I make `ExecutionEnvironmentDTO` extend a `JobCommonDTO` it then screws up `Application`, `Cluster`, `Command` cause they all extend that `ExecutionEnvironmentDTO`. Since Java doesn't support multiple inheritance this probably isn't worth the hassle right now. For 4.0 we'll clean all this up anyway and deprecate or remove all these DTO's in favor of new ones that line up to 4.0 way of doing things whatever that ends up being.

PR #653, opened 2017-11-28
Fix javadoc for #651

tgianos:
Also add javadoc task to pull request build to prevent surprises once merge occurs

Build broke for #651 due to lack of javadoc target in PR builds


tgianos:
Fair enough. Split them up @mprimi

PR #654, opened 2017-11-28
Improve documentation and compiliation

tgianos:
Improve compilation times by combining meta-model generation into the main java build
Improve documentation by explicitly declaring JSON types

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,7 @@ configure(javaProjects) {
             dependency("org.apache.commons:commons-exec:1.3")
             dependency("org.bitbucket.b_c:jose4j:0.6.2")
             dependency("org.dbunit:dbunit:2.5.4")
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -203,9 +197,27 @@ configure(javaProjects) {
...
+    clean {
+        dependsOn(removeIdeaOutDirectory)
+    }
+
     compileJava {
+        doFirst {
+            generatedSourcesDir.mkdirs()
+            generatedSourcesJavaDir.mkdirs()
+        }
+        dependsOn(removeGeneratedSources)
```
mprimi: This will delete and regen every time we compile, so all downstream tasks of compile will never be up to date

tgianos: Seems to work better than before when the downstream worked every time. Since the hash of the generated files are always the same (provided no entity class changes) the outputs are marked as the same and the tasks don't run.

PR #658, opened 2017-12-06
Add metadata field to job, cluster, comnand and application resources

tgianos:
For adding JSON metadata on top of resources that don't fit into existing data model


genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
```diff
@@ -44,12 +49,19 @@
...
     static final String GENIE_TAG_NAMESPACE = "genie.";
     static final String GENIE_ID_TAG_NAMESPACE = GENIE_TAG_NAMESPACE + "id:";
     static final String GENIE_NAME_TAG_NAMESPACE = GENIE_TAG_NAMESPACE + "name:";
     private static final char COMMA = ',';
     private static final String EMPTY_STRING = "";
 
+    static {
+        final DateFormat iso8601 = new GenieDateFormat();
+        iso8601.setTimeZone(TimeZone.getTimeZone("UTC"));
+        MAPPER = new ObjectMapper().registerModule(new Jdk8Module()).setDateFormat(iso8601);
```
mprimi: Does this mean that if someone sends a JSON with a string field containing a date in a format other than ISO8601 when we serialize then deserialize, the string value for that field will change?
This might trip some users, if they rely on a specific format for setting and parsing back dates from the metadata they store.
Although that is probably the case anyway, wether you explicitly use ISO8601 or not.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaServiceUtils.java
```diff
@@ -263,4 +277,38 @@ static JobMetadata toJobMetadataDto(final JobMetadataProjection jobMetadataProje
...
+        final E entity
+    ) {
+        if (entity.getMetadata().isPresent()) {
+            try {
+                final String metadata = entity.getMetadata().get();
+                builder.withMetadata(metadata);
+            } catch (final GeniePreconditionException gpe) {
+                // Since the DTO can't be constructed on input with invalid JSON this should never even happen
+                log.error("Invalid JSON metadata. Should never happen.", gpe);
+                builder.withMetadata((JsonNode) null);
```
mprimi: Why catching this rather than letting it throw?

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
tgianos: This is consistently how we serialize dates from the database and is mentioned in the documentation. I'm not sure there's much else we can do there.

genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaServiceUtils.java
tgianos: I could throw a `GenieServerException` but that would basically fail the entire action. Since in reality it should be impossible cause the input has to be valid JSON I felt it wasn't worthwhile to have `throws` messages all over the place. If the Genie Exceptions had been moved to runtime exceptions by now I would have let it do that for sure. Otherwise I need to make it bubble up a checked exception and felt it wasn't worth it at this time given our intention to make the exceptions runtime exceptions anyway

PR #660, opened 2017-12-13
Miscellaneous build tweaks

mprimi:


travis/buildViaTravis.sh
```diff
@@ -6,14 +6,14 @@ set -e
 
 if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
   echo -e "Build Pull Request #$TRAVIS_PULL_REQUEST => Branch [$TRAVIS_BRANCH]"
-  ./gradlew --no-daemon build javadoc asciidoc coveralls dockerBuildAllImages
+  # Build and run all tests, create coverage report
+  ./gradlew --no-daemon build coveralls
```
tgianos: So do we run the integration tests here with the in memory database and then two more times with the other databases? Why leave coveralls here but move the others?

.travis.yml
```diff
@@ -1,5 +1,6 @@
 sudo: required
 dist: trusty
+group: travis_lts
```
tgianos: Do we really want to do this? It could lead to us having to tackle far more issues when they do upgrade the LTS images rather than dealing with them little at a time as they come

travis/buildViaTravis.sh
mprimi: My logic was to move the two artifact-producing steps (asciidoc, docker) past the test, so they run (and publish) only if all tests passed.

Not being sure of how coverall works, i left it where it was (together with `build`). I'm going to try to move it after to see if we get additional coverage stats from the 2 sets of integration tests.

.travis.yml
mprimi: `travis_lts` is the default value if we don't set this, so I was just making it explicit.
The only alternative is setting `travis_latest`. Do you think we should use that instead?

tgianos: I'm fine with leaving it out and just going with the default behavior so we're on the "paved path"

travis/buildViaTravis.sh
tgianos: I doubt the coveralls would change as the code paths should be the same but you can test it

mprimi: I agree it may not go up (we don't have SQL-specific code paths).
But it may go *down*, depending on how coverall works, because it may only publish results from the last execution (which would be integration, if i move this away from `build`), that is the reason I left it where it was. But going to try, maybe it is able to look at everything regardless of when it executed.

.travis.yml
mprimi: Removed this commit

PR #661, opened 2017-12-14
Switch from HSQLDB to H2

tgianos:
HSQLDB had to be pinned back to 2.3.3 and was causing builds to timeout when version was upgraded beyond that. H2 doesn't have that problem with version specified by platform BOMs.

Also the schema ddl language is closer to MySQL which is our default database.

This PR also includes turning on schema validation from hibernate as part of application boot process.


genie-web/src/main/resources/db/migration/h2/V3_3_0__Merge_Jobs_Tables.sql
```diff
@@ -0,0 +1,501 @@
+/*
```
mprimi: Is this a 1:1 copy of one of the existing ones?

tgianos: I took the mysql 3.3.0 one and modified it for syntax that wasn't supported in H2 primarily the indices need to be defined separately from table definition like how postgres does it. Also constraint definitions were slightly different syntax but that's pretty much it.

PR #665, opened 2017-12-15
Add protobuf, gRPC

mprimi:


mprimi:
Only the last 2 commits should be in this PR, the rest are in https://github.com/Netflix/genie/pull/660 and will be removed once that is merged and this is rebased.

build.gradle
```diff
@@ -98,22 +99,45 @@ configure(javaProjects) {
             dependency("org.opensaml:opensaml:2.6.4")
             dependency("org.python:jython-standalone:2.7.1")
             dependency("org.springframework.security.extensions:spring-security-saml2-core:1.0.2.RELEASE")
+            dependencySet(group: "com.google.protobuf", version: "3.4.0") {
```
tgianos: Horrible OCD nitpick on my part but I've tried to keep these in alphabetical order.

build.gradle
```diff
@@ -98,22 +99,45 @@ configure(javaProjects) {
...
+              entry "grpc-server-spring-boot-autoconfigure"
+              entry "grpc-server-spring-boot-starter"
+            }
         }
     }
 
     ext {
         generatedSourcesDir = file("${buildDir}/generated-sources/main")
         generatedSourcesJavaDir = file(new File(generatedSourcesDir, "/java"))
+        generatedSourcesGRPCDir = file("${buildDir}/generated-grpc/main/java")
```
tgianos: Should this not extend from the `generatedSourcesDir` the same way we do with Java? Or why not just dump it in to the same java generated sources dir?

build.gradle
```diff
@@ -98,22 +99,45 @@ configure(javaProjects) {
...
     sourceSets {
         main {
             java {
                 srcDir generatedSourcesJavaDir
+                srcDir generatedSourcesGRPCDir
             }
         }
     }
 
+    idea.module {
```
tgianos: I don't see why this is needed? Are you depending on gradle for all your IDEA stuff? I haven't seen any issues?

genie-proto/build.gradle
```diff
@@ -0,0 +1,81 @@
+apply plugin: 'com.google.protobuf'
```
tgianos: For consistency I've been going with `"` instead of `'` in the build.gradle files.

genie-proto/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+    protogenIgnorePattern = "**/com/netflix/genie/protogen/*"
+}
+
+dependencies {
+
+    /*******************************
+     * Compile Dependencies
+     *******************************/
+
+     compile ("com.google.protobuf:protobuf-java")
```
tgianos: *OCD* extra space

genie-proto/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+
+}
+
+// Configure protobuf classes and grpc stubs generation
+protobuf {
+    protoc {
+        // Download the protoc compiler
+        artifact = 'com.google.protobuf:protoc'
+        // Use project convention for generated files
+        generatedFilesBaseDir = file("${buildDir}/generated-grpc")
```
tgianos: Why not reference the same variable from the root build file?

genie-proto/src/main/proto/genie.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+
+package com.netflix.genie.proto;
```
tgianos: Do we want to call it `proto` or give it a name more representative of the generated classes purposes? e.g. we have our `DTO` classes we don't call them like `JSON` or whatever.

I'm not sure what the name would be but i'm wondering if packaging them under namespace more representative of their purpose would help with clarity long term but I can also understand just saying they're proto classes

genie-web/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
      *******************************/
 
     compile(project(":genie-core"))
+    compile project(":genie-proto")
```
tgianos: inconsistent with how above project dependency is defined

genie-web/src/main/java/com/netflix/genie/grpc/interceptors/SimpleLoggingInterceptor.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+        final ServerCall<ReqT, RespT> call,
+        final Metadata headers,
+        final ServerCallHandler<ReqT, RespT> next
+    ) {
+
+        return next.startCall(new ForwardingServerCall.SimpleForwardingServerCall<ReqT, RespT>(call) {
+
+            @Override
```
tgianos: Missing Javadocs here

genie-web/src/main/java/com/netflix/genie/grpc/services/impl/GreeterServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Example implementation of the Greeter service definition.
+ */
+@GrpcService(
+    value = GreeterServiceGrpc.class,
+    interceptors = {
+        SimpleLoggingInterceptor.class,
+    }
+)
+public class GreeterServiceImpl extends GreeterServiceGrpc.GreeterServiceImplBase {
```
tgianos: Is there any reason to actually commit these things into the repository? We'll never actually use this it's like checking in hello world right?

build.gradle
mprimi: IDEA will re-generate those models and then conflict on duplicate classes.
This config only affect the IDEA project generation gradle task (i.e. importing a gradle project the way you do it is not using this snippet)

build.gradle
mprimi: This is a limitation of the gradle protobuf plugin, I don't remember the details because it was before the break, but I spent a day or two making this work with generated files in the same folder and there's no way to.
It has to do with hardcoded path components that the gradle protobuf plugin appends. In the end it would end up in `main/main/java` or something like that

genie-proto/build.gradle
mprimi: Done

genie-proto/build.gradle
mprimi: Done

genie-proto/src/main/proto/genie.proto
mprimi: Since they will be used in various different components, I couldn't think of a name to group them all, other than `proto`.
And keep in mind these are stubs and raw messaged, we'll need wrappers and implementations of those stubs around the code, and those can have good package names.

genie-web/src/main/java/com/netflix/genie/grpc/interceptors/SimpleLoggingInterceptor.java
mprimi: Done

genie-web/build.gradle
mprimi: Done

genie-proto/build.gradle
mprimi: The value is not the same, but let me see if I can refactor this to avoid hardcoding the same path twice

genie-proto/src/main/proto/genie.proto
mprimi: Oh and the java package name is actually defined a few lines below: `com.netflix.genie.protogen`.
Any preference on one vs. the other or something else entirely?

mprimi: Changed java package name to `com.netflix.genie.proto`

mprimi:
@tgianos addressed all review comments

genie-web/src/main/java/com/netflix/genie/grpc/services/impl/GreeterServiceImpl.java
mprimi: Right now, the only value is testing that the build works (e.g., generates protobufs, Spring configures the service, etc).
I agree we should remove them as soon as we have some real services or messages.

PR #669, opened 2018-01-08
Add message to @NotBlank validation annotations of Tag and File

mprimi:
Without an explicitly message, the error propagated back to the user is a generic "cannot be empty".
This change makes it a little simpler to identify which field contains a blank string that is getting rejected.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaFileServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaFileServiceImpl.java
@@ -52,7 +52,9 @@ public class JpaFileServiceImpl implements FileService {
      */
     @Override
 //    @Transactional(propagation = Propagation.REQUIRES_NEW)
-    public void createFileIfNotExists(@NotBlank final String file) throws GenieException {
+    public void createFileIfNotExists(
+        @NotBlank(message = "File path cannot be blank") final String file
+    ) throws GenieException {
         if (this.fileRepository.existsByFile(file)) {
             return;
         }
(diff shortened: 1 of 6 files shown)


genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaFileServiceImpl.java
```diff
@@ -52,7 +52,9 @@ public JpaFileServiceImpl(final JpaFileRepository fileRepository) {
      */
     @Override
 //    @Transactional(propagation = Propagation.REQUIRES_NEW)
-    public void createFileIfNotExists(@NotBlank final String file) throws GenieException {
+    public void createFileIfNotExists(
+        @NotBlank(message = "File path cannot be blank") final String file
```
tgianos: The annotations must match exactly in the interface definitions and the implementations or you will get a runtime error.

mprimi: Yes, bumped into this when testing on my stack.

mprimi:
@tgianos Is it worth activating the validators in our integration tests, so we could check for this? Right now they are not, sending an empty `setupFile` fails, but later, these annotations are not stopping it from going through.

tgianos:
@mprimi You can activate the validators. I think there was a config for it. In the main branch #670 uses the default application context so I believe they're already validated there. I think you should make this change against 3.3.x first and then port it over to master as master will have more diverged once #670 is merged.

mprimi:
Targeting 3.3.0 branch first.

PR #671, opened 2018-01-09
Add message to @NotBlank validation annotations of Tag and File

mprimi:
Without an explicitly message, the error propagated back to the user is a generic "cannot be empty".
This change makes it a little simpler to identify which field contains a blank string that is getting rejected.

Example server response for an empty tag:

```
{
  "timestamp" : "2018-01-09T00:26:24.700Z",
  "status" : 412,
  "error" : "Precondition Failed",
  "exception" : "javax.validation.ConstraintViolationException",
  "message" : "Tag name cannot be blank",
  "path" : "/api/v3/clusters"
}
```

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
@@ -33,6 +33,7 @@ import com.netflix.genie.core.services.TagService;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.hibernate.validator.constraints.NotBlank;
 
 import java.text.DateFormat;
 import java.util.Set;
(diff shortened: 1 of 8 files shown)


genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaBaseService.java
```diff
@@ -123,7 +126,9 @@ FileEntity createAndGetFileEntity(final String file) throws GenieException {
      * @return The tag entity that has been persisted in the database
      * @throws GenieException on error
      */
-    TagEntity createAndGetTagEntity(final String tag) throws GenieException {
+    TagEntity createAndGetTagEntity(
+        @NotBlank(message = "Tag name cannot be blank") final String tag
```
tgianos: is it really a "name"? Or is it a value? I'm never sure on this myself just wondering whether it's confusing cause the tag is what it is... it's a tag

mprimi: Yeah, i guess i could just say `Tag cannot be blank`

PR #674, opened 2018-01-16
Upgrade Hibernate to 5.2 and move to Java 8 JSR 310 time types

tgianos:
In preparation for moving to Spring Boot 2.x as the primary platform for Genie 4 development it will upgrade us to Hibernate 5.2.x. This PR addresses that upgrade to prepare us for when that switch happens. It simultaneously addresses a long time technical debt item of moving off the `java.util.Date` class for instant savings in favor of the `java.time.Instant` and associated classes. 
This provides immutable, thread safe implementations for our times and far improved helper methods and utilities.


tgianos:
Still WIP while testing against existing data sets and seeing if old serializer/deserializers are needed or not.

genie-common/src/main/java/com/netflix/genie/common/dto/ClusterCriteria.java
```diff
@@ -51,19 +51,16 @@
...
+        tags.forEach(
+            tag -> {
+                if (StringUtils.isNotBlank(tag)) {
+                    builder.add(tag);
                 }
-            );
-        }
+            }
+        );
         this.tags = builder.build();
```
danielbwatson: Alternatively:

```java
this.tags = tags.stream().filter(StringUtils::isNotBlank).collect(ImmutableSet.toImmutableSet());
```

tgianos: Would love to but this is only in Guava 21+ and I have concerns about upgrading such a core library out of band with what all the other libraries work with it

genie-client/build.gradle
```diff
@@ -35,38 +35,3 @@ dependencies {
     testCompile("org.springframework.boot:spring-boot-starter-jetty")
     testCompile("org.springframework.cloud:spring-cloud-contract-wiremock")
 }
-
```
mprimi: This doesn't belong with either commits...

genie-common/src/main/java/com/netflix/genie/common/dto/ClusterCriteria.java
mprimi: `tags.forEach` will throw NPE if tags is null. Am I missing something that guarantees this is not null?

genie-common/src/main/java/com/netflix/genie/common/dto/JobRequest.java
```diff
@@ -216,18 +216,14 @@ public Builder(
...
-                this.bClusterCriterias.addAll(clusterCriterias);
-            }
-            if (commandCriteria != null) {
-                commandCriteria.forEach(
-                    criteria -> {
-                        if (StringUtils.isNotBlank(criteria)) {
-                            this.bCommandCriteria.add(criteria);
-                        }
+            this.bClusterCriterias.addAll(clusterCriterias);
+            commandCriteria.forEach(
```
mprimi: Same here, possible NPE, unless I'm missing something

genie-common/src/main/java/com/netflix/genie/common/dto/JobRequest.java
```diff
@@ -256,18 +252,14 @@ public Builder(
...
```
mprimi: And here

genie-common/src/test/java/com/netflix/genie/common/dto/ApplicationStatusUnitTests.java
```diff
@@ -64,6 +64,6 @@ public void testInvalidApplicationStatus() throws GeniePreconditionException {
      */
     @Test(expected = GeniePreconditionException.class)
     public void testBlankApplicationStatus() throws GeniePreconditionException {
-        ApplicationStatus.parse(null);
```
mprimi: Isn't it worth testing both?

genie-common/src/test/java/com/netflix/genie/common/dto/ClusterCriteriaUnitTests.java
```diff
@@ -46,15 +46,6 @@ public void canConstruct() {
         Assert.assertThat(cc.getTags(), Matchers.is(tags));
     }
 
-    /**
```
mprimi: Whatever the expected behavior is (including NPE), isn't worth leaving this here and testing (so that when things change, a flag is raised)

genie-common/src/test/java/com/netflix/genie/common/dto/JobUnitTests.java
```diff
@@ -290,7 +292,7 @@ public void canBuildJobWithNulls() {
         builder.withCommandName(null);
         builder.withFinished(null);
         builder.withStarted(null);
-        builder.withStatus(null);
```
mprimi: Status is no longer optional?

genie-client/build.gradle
tgianos: That's cause it's completely worthless left over dangling code from when @irontable removed the python code from the repository. But sure I can break it into its own commit.

genie-common/src/main/java/com/netflix/genie/common/dto/ClusterCriteria.java
tgianos: `@ParametersAreNonnullByDefault` was added to the package header for `com.netflix.genie.common.dto`. As tags are not optional in this case it will ensure something is passed in at compile time.

genie-common/src/test/java/com/netflix/genie/common/dto/ApplicationStatusUnitTests.java
tgianos: The IDE/compiler checks the `@Nonnull` for us after `@ParametersAreNonnullByDefault` was added. it is now an opt out via `@Nullable`. Leaving the check in with `null` just produced warnings.

genie-common/src/test/java/com/netflix/genie/common/dto/ClusterCriteriaUnitTests.java
tgianos: You can't pass in null the compiler will fail the build.

genie-common/src/test/java/com/netflix/genie/common/dto/JobUnitTests.java
tgianos: It should never have been. It's even marked required in the docs: https://netflix.github.io/genie/docs/3.3.2/rest/#_job

tgianos:
@mprimi this should be ready to merge. Testing on live data looks good.

PR #680, opened 2018-01-22
Split unused file and tag entity deletion into a SELECT ... FOR UPDATE and a DELETE to prevent database locking

tgianos:
The previous `DELETE FROM...` query on internal sample data took 9 minutes to complete. The select takes of the same query form only took 100ms. This implementation selects the `id`'s to delete from the `FILES` and `TAGS` tables and locks the rows as `FOR UPDATE`. It then explicitely calls `DELETE` on set of ids preventing all other tables from having to be opened.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaFileServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaFileServiceImpl.java
@@ -28,6 +28,7 @@ import org.springframework.transaction.annotation.Transactional;
 import javax.validation.constraints.NotNull;
 import java.time.Instant;
 import java.util.Date;
+import java.util.stream.Collectors;
 
 /**
  * JPA based implementation of the FileService interface.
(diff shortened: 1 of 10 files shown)


genie-core/src/main/java/com/netflix/genie/core/jpa/services/JpaFileServiceImpl.java
```diff
@@ -74,7 +75,13 @@ public void createFileIfNotExists(@NotBlank(message = "File path cannot be blank
      * {@inheritDoc}
      */
     @Override
-    public int deleteUnusedFiles(@NotNull final Instant createdThreshold) {
-        return this.fileRepository.deleteUnusedFiles(Date.from(createdThreshold));
+    public long deleteUnusedFiles(@NotNull final Instant createdThreshold) {
+        return this.fileRepository.deleteByIdIn(
```
ajoymajumdar: Do you want to batch it or does JPA handle that?

tgianos: given the scale of this data I was trying to avoid batching it

PR #681, opened 2018-01-23
Cherry-picks from 3.3.x

mprimi:
Includes #678 #679


mprimi:
@tgianos the last commit had conflicts, does it look all right to you?

PR #693, opened 2018-02-07
Upgrade to Spring Boot 2.0 RC1

tgianos:
This PR includes modifications to get Genie 4.0 building and running with Spring Boot 2.0 RC1.

All integration tests and local testing against data sets and real services (security in particular) pass but expect to find on-going issues as we move through 4.0 development.

This PR also includes a modification of the testing framework used for API integration tests. In the past `MockMvc` was used. This PR switches to using [RestAssured](http://rest-assured.io/) to ensure full HTTP requests are sent to a full Tomcat instance to better simulate real clients as well as decouple us from concern that tests may just pass cause we're using Spring components all over.


tgianos:
@danielbwatson @ajoymajumdar @zhljen FYI on possible lessons learned for other project migrations

twicksell:
@tcellucci fyi as well

tgianos:
@mprimi I broke up the middle commits a lot for easier review but all the 2.0 boot commits I can redo and squash into one atomic commit after review is complete

PR #695, opened 2018-02-14
Switch metrics collection to Micrometer

tgianos:
Allow systems to plugin their own metrics aggregation backend without code changes. Internally this will be an `AtlasRegistry`

Upgrade to Spring Cloud Finchley M6


tgianos:
@ajoymajumdar @zhljen @danielbwatson fyi

PR #697, opened 2018-02-14
Agent registration

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/AgentConfig.java
```diff
@@ -0,0 +1,61 @@
...
+class AgentConfig {
+
+    /**
+     * Setup bean validation.
+     *
+     * @return The bean validator
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    public String hostName() {
```
tgianos: Maybe instead of creating beans of type `String` which is annoying and almost always needs qualifiers we should combine these two things into a Bean that is more of a properties container similar to how AWS exposes Instance metadata. That way it's a single bean of type `AgentInstanceMetadata` or something which we can pass around and modify as we go.

genie-agent/src/main/java/com/netflix/genie/agent/cli/ServerArgumentsImpl.java
```diff
@@ -0,0 +1,53 @@
...
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.validators.PositiveInteger;
+import lombok.Getter;
+import org.springframework.stereotype.Component;
+
+/**
+ * Implementation of ServerArguments delegate.
+ */
+@Component
+@Getter
```
tgianos: With this only having a getter and no `AllArgsConstructor` how are they set?

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
```diff
@@ -0,0 +1,48 @@
...
+@ThreadSafe
+public interface ExecutionContext {
+
+    /**
+     * Set the unique agent identifier obtained by the server.
+     *
+     * @param agentId a non-blank string
+     * @throws RuntimeException if the agent id is already set
+     */
+    void setAgentId(@NotBlank final String agentId);
```
tgianos: Should it be possible to create one of these WITHOUT having this agentId? If so I would argue the interface should only have the getter and it should be immutable. Also why have this as an interface instead of a concrete or abstract class? Are we planning multiple implementations of this interface?

genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/AgentRegistrationException.java
```diff
@@ -0,0 +1,46 @@
...
+     * Construct with a message.
+     *
+     * @param message message
+     */
+    public AgentRegistrationException(final String message) {
+        super(message);
+    }
+
+    /**
+     * Construct with a message.
```
tgianos: copy paste miss here.

genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/AgentRegistrationException.java
```diff
@@ -0,0 +1,46 @@
...
+
+package com.netflix.genie.agent.execution.exceptions;
+
+/**
+ * Error to register the agent with the server.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public class AgentRegistrationException extends Exception {
```
tgianos: Are you SURE you don't want to just go with RuntimeExceptions? I say this from experience of driving myself nuts putting `@throws GenieException` on every server side method just in case we ever through an exception and most of the time just propagating it all the way up to ExceptionMapper anyway cause they're unrecoverable. In this case is it not also unrecoverable? There's nothing stopping you from putting `@throws` in your JavaDoc for these runtime exceptions I Just think you're asking to add a lot of boilerplate with no real benefit

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/GrpcAgentRegistrationServiceImplSpec.groovy
```diff
@@ -0,0 +1,215 @@
...
+import io.grpc.stub.StreamObserver
+import io.grpc.testing.GrpcServerRule
+import org.junit.Rule
+import spock.lang.Ignore
+import spock.lang.Specification
+
+
+class GrpcAgentRegistrationServiceImplSpec extends Specification {
+
+    final String hostName = "agent.com"
```
tgianos: Any reason why you are using so many Java class stuff when you could just use `def`?

genie-proto/src/main/proto/genie.proto
```diff
@@ -31,3 +31,21 @@ message PongResponse {
   google.protobuf.Timestamp timestamp = 2;
   map<string, string> server_metadata = 3;
 }
+
+//
+service AgentRegistrationService {
```
tgianos: I thought we were going to go with a `GenieService`?

genie-agent/src/main/java/com/netflix/genie/agent/cli/ServerArgumentsImpl.java
mprimi: JCommander sets them even if they are private.

genie-agent/src/main/java/com/netflix/genie/agent/AgentConfig.java
mprimi: Sounds good, should I do the same on the server?

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
mprimi: I don't know what we're going to stuff in this, likely a bunch of stuff (e.g. stack of states that requires cleanup).
It's an interface so the implementation can be package local.

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/GrpcAgentRegistrationServiceImplSpec.groovy
mprimi: Better autocomplete when writing the tests...

genie-proto/src/main/proto/genie.proto
mprimi: There may be implications I'm not seeing and that are not coming out when a single service is present. Once we have 2 or 3 we can try to refactor into a single client and see what happens

genie-agent/src/main/java/com/netflix/genie/agent/AgentConfig.java
tgianos: I think we probably should do something similar. I was looking at the Spring Cloud AWS earlier and they provide an [abstraction](http://cloud.spring.io/spring-cloud-static/spring-cloud-aws/2.0.0.M3/multi/multi__cloud_environment.html#_cloud_environment) on instance metadata that we could wrap with our own common interface. I don't think it should be the same interface on the server and agent if that's what you're asking

mprimi:
@tgianos changed a few things and ready for review

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadataImpl.java
```diff
@@ -0,0 +1,86 @@
...
+            if (!StringUtils.isBlank(hostName)) {
+                return hostName;
+            }
+        } catch (final UnknownHostException e) {
+            log.warn("Failed to retrieve local host name", e);
+        }
+        return FALLBACK_STRING;
+    }
+
+    private static String getAgentPidOrFallback() {
```
tgianos: FYI https://docs.spring.io/spring-boot/docs/2.0.0.RC1/reference/htmlsingle/#production-ready-process-monitoring

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadataImpl.java
```diff
@@ -0,0 +1,86 @@
...
+    private final String agentHostName;
+    private final String agentPid;
+
+    AgentMetadataImpl() {
+        this.agentVersion = getAgentVersionOrFallback();
+        this.agentHostName = getAgentHostNameOrFallback();
+        this.agentPid = getAgentPidOrFallback();
+    }
+
+    private static String getAgentVersionOrFallback() {
```
tgianos: Why did you make all these static?

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadata.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.hibernate.validator.constraints.NotBlank;
+
+/**
+ * Container for Genie agent metadata and runtime information.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public interface AgentMetadata {
```
tgianos: Why an interface and not just a class?

genie-agent/src/main/java/com/netflix/genie/agent/rpc/GrpcConfig.java
```diff
@@ -0,0 +1,67 @@
...
+@Slf4j
+class GrpcConfig {
+
+    @Bean
+    @Lazy
+    ManagedChannel channel(
+        final ArgumentDelegates.ServerArguments serverArguments
+    ) {
+        return ManagedChannelBuilder.forAddress(serverArguments.getServerHost(), serverArguments.getServerPort())
+            .usePlaintext(true)
```
tgianos: Is this something we'd leave in in production? feels like something we should flip on a property/profile basis?

genie-agent/src/main/java/com/netflix/genie/agent/cli/InfoCommand.java
```diff
@@ -68,42 +90,54 @@ public void run() {
...
+                            envEntry.getKey(),
+                            envEntry.getValue()
+                        )
                     )
-                );
+                    .append(System.lineSeparator());
             }
         }
 
         if (infoCommandArguments.isIncludeProperties()) {
```
tgianos: You probably also want to Serialize the spring `Environment` here which would give you everything including the systme properties

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/InitializeAction.java
```diff
@@ -35,13 +39,35 @@
...
     @Override
     protected Events executeStateAction(final StateContext<States, Events> context) {
         log.info("Initializing...");
-        //TODO implement this action
+
+        final String agentId;
+        try {
+            agentId = agentRegistrationService.registerAgent();
+        } catch (final AgentRegistrationException e) {
+            throw new RuntimeException("Failed to obtain agent id", e);
```
tgianos: So this harkens back to my other comment about the exceptions. What is the value here of catching a checked exception only to rewrap in a runtime? You could use Guava Throwables.Propogate here too.

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadataImpl.java
mprimi: No particular reason, they enforces they're pure functions.

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadata.java
mprimi: So the implementation is package-local. I though you said you wanted to to more of this moving forward, rather than just making classes public willy-nilly because they're used in a different package.

genie-agent/src/main/java/com/netflix/genie/agent/rpc/GrpcConfig.java
mprimi: Indeed, SSL and the metatron stuff needs to plug in in here.

genie-agent/src/main/java/com/netflix/genie/agent/cli/InfoCommand.java
mprimi: Oh, didn't know about this, will add.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/InitializeAction.java
mprimi: So there's the service abstraction, the service does not know how it is going to be used.
So the checked exception is the way for the service to say "this failed" and give you some context (exception trace).
The service can also still throw runtime exceptions.
In this particular case, at the layer above, we wrap it in a Runtime, because that's the contract for state actions.
If you want, let's sit down and think through exception strategies, I'm not married to the idea that this is the best way, we can always change as we go.

PR #699, opened 2018-03-01
Server side implementation of resolveJobSpecification RPC API

tgianos:
Other miscellaneous improvements like upgrading to Bean Validation 2.0 universally and standardizing spacing in proto file.


genie-agent/src/main/java/com/netflix/genie/agent/configs/ValidationConfig.java
```diff
@@ -15,7 +15,7 @@
  *     limitations under the License.
  *
  */
-package com.netflix.genie.agent;
+package com.netflix.genie.agent.configs;
```
mprimi: I was keeping each config in the package where the components are.
I like it because I don't have to make implementation public just to make them visible from the config in a different package (`such as `....agent.configs`).
Because this config creates a component that is "global" to the whole agent, I placed it in `...agent`.

I'm not in favor of having a `config` package that only contains a subset of the configs.
And if we move all configs here, there's stuff we may be forced to to make `public` as a consequence.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobSpecificationService.java
```diff
@@ -0,0 +1,50 @@
...
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.agent.execution.services;
+
+import com.netflix.genie.common.dto.v4.JobRequest;
```
mprimi: Do you expect this DTO to work for both the `v4` API to submit jobs AND this internal job request resolution?

I haven't given much thought to wether that is the case or not, but in general seems strange to use a public API entity for an internal protocol.
For one thing we may want to evolve them separately.
Just a gut feeling that they're going to diverge more and more.

Example: Container information, it applies to the API but not here.
Maybe the API job request is a superset of the internal fields?

genie-common/src/main/java/com/netflix/genie/common/dto/v4/CommonUserMetadata.java
```diff
@@ -0,0 +1,193 @@
...
+        this.tags = builder.bTags == null ? ImmutableSet.of() : ImmutableSet.copyOf(builder.bTags);
+    }
+
+    /**
+     * Get the version.
+     *
+     * @return The version as an {@link Optional}
+     */
+    public Optional<String> getVersion() {
+        return Optional.ofNullable(this.version);
```
mprimi: Given that `Optional` are immutable, why not create them at construction time?
Would create fewer runtime objects and could generate getters via annotation.

genie-common/src/main/java/com/netflix/genie/common/dto/v4/ExecutionEnvironment.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Get the configuration files needed at runtime. The returned set will be immutable and any attempt to modify will
+     * throw an exception.
+     *
+     * @return The set of setup file locations.
+     */
+    public Set<String> getConfigs() {
+        return this.configs;
```
mprimi: Why are these getters necessary? (as opposed to `@Getter`)

genie-proto/src/main/proto/genie.proto
```diff
@@ -48,3 +48,86 @@ message AgentRegistrationResponse {
...
+
+//
+// Job Specification APIs
+//
+service JobSpecificationService {
+    rpc resolveJobSpecification (ResolveJobSpecificationRequest) returns (JobSpecificationResponse);
+    rpc getJobSpecification (GetJobSpecificationRequest) returns (JobSpecificationResponse);
+}
+
+message Criterion {
```
mprimi: `OneOf`?

genie-proto/src/main/proto/genie.proto
```diff
@@ -48,3 +48,86 @@ message AgentRegistrationResponse {
...
+    repeated string configs = 12;
+    repeated string dependencies = 13;
+    repeated string command_args = 14;
+    bool is_interactive = 15;
+}
+
+message ExecutionResourceCriteria {
+    repeated Criterion cluster_criteria = 1;
+    Criterion command_criterion = 2;
+    repeated string application_ids = 3;
```
mprimi: Mh, is this right? Application IDs in the specification request?

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaClusterServiceImpl.java
```diff
@@ -590,6 +581,48 @@ public long deleteTerminatedClusters() {
         );
     }
 
+    private Map<Cluster, String> findClustersAndCommandsForJob(
+        final List<Set<String>> clusterCriteria,
+        final Set<String> commandCriterion
+    ) throws GenieServerException {
+        final Map<Cluster, String> foundClusters = Maps.newHashMap();
+        for (final Set<String> clusterCriterion : clusterCriteria) {
+            final List<Object[]> clusterCommands = this.clusterRepository.findClustersAndCommandsForCriterion(
```
mprimi: I think using a pair/tuple for this would be better than an `Object[2]`.
Type safe and can get rid of the next dozen or so lines.

genie-agent/src/main/java/com/netflix/genie/agent/configs/ValidationConfig.java
tgianos: The flip side and why I don't favor moving the configs around is it makes it hard to find out where everything in the app is coming from. Even today I had to search for where various beans were defined which could easily lead to me redefining them. Spring already breaks the `public/private/protected` nature of the config files anyway by using annotations/reflection. Generally speaking in community jars I've looked at this is how the spring configs are packaged. Especially the boot auto configure configs they're all sitting in one package within a given jar.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobSpecificationService.java
tgianos: I do.

Firstly the Job Request is not the wire level protocol between the two components and it is translated on both sides by the adapter so it's already shielded there.

Second, the job request as it currently stands consists of only 4 required fields (`name`, `user`, `clusterCriteria`, `commandCriterion`). Everything else is completely optional.

Thirdly we can extend `JobRequest` or wrap it in a larger payload representation for an API request (which I believe is what you mean by your last line in your comment) e.g.
```
POST /api/v4/jobs
{
   "additionalInfo" : { ... },
   "jobRequest": { ... },
   "containerParameters": { ... }
}
```

I don't have much interest in creating yet another data class (server side dto, wire protocol proto file, agent side job metadata something) which will end up representing the same information. My feeling is that regardless of job launching path there are certain fields that are necessary for Genie to properly handle a job and I've tried to capture these in this `JobRequest` DTO.

genie-common/src/main/java/com/netflix/genie/common/dto/v4/CommonUserMetadata.java
tgianos: I was thinking about this too while I was coding it. Generally I've tried to avoid doing what you say because `Optional` best practice articles I've read from community say only really use them to return from methods because if you store them in internal state you have to unwrap them all the time. However in this instance since the classes are entirely immutable and just used to transfer data around I think what you're saying is the better thing to do. I'll make the change for the v4 DTOs in this PR

genie-common/src/main/java/com/netflix/genie/common/dto/v4/ExecutionEnvironment.java
tgianos: I did this because internally I was making the members types from Guava and I didn't want to expose that externally forcing a compile time dependency of jar consumers on Guava. It's probably an overkill thing but I was effectively down casting these `sets`/`lists` so as not to expose internal representations and instead focus on the important standard lib API's that should be used to work with them. Whether it's the best thing to do or not I'm not sure and we could talk about exposing the Guava classes.

Another option was just to have the members be `Set` or `List` instead of `ImmutableSet` and `ImmutableList` and just assign instances of the immutable versions in the constructor. I made the members the concrete types so that the code was more self documenting.

genie-proto/src/main/proto/genie.proto
tgianos: The proposed algorithm for handling criteria in v4 doesn't enforce that only one of these is present exclusively.

For example one of the primary requested use cases by @stephen-mw is to be able to override the `UP` status of a cluster while still using the same tags. For this case the `Criterion` needs to be able to provide both `tags` and `status`

genie-proto/src/main/proto/genie.proto
tgianos: It's a feature present in Genie 3 requested by Dan's team that allows them to override defaults if need be. I do not expect most users to use this feature but it is there for power users and thus carried over here as well.

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaClusterServiceImpl.java
tgianos: This might very well be true. I wrote it for `3.3` and since that native sql query is pretty complicated once I had it working I didn't really mess with it. If you look this method is just a copy of a code blocked moved to a helper method for re-use between the 4.0 API and the 3.x API. I'm open to improving it but I don't think we should do it as part of this PR particularly as once we change how the `Criterion` are handled for 4.x much of this may change anyway.

genie-common/src/main/java/com/netflix/genie/common/dto/v4/CommonUserMetadata.java
tgianos: So I went to change this and IntelliJ gave the following warning:
```
Reports any uses of java.util.Optional<T>, java.util.OptionalDouble, java.util.OptionalInt, java.util.OptionalLong or com.google.common.base.Optional as the type for a field or a parameter. Optional was designed to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result". Using a field with type java.util.Optional is also problematic if the class needs to be Serializable, which java.util.Optional is not. 
```

The first part of that Is what I was commenting on above but we both agree that's not really a big deal in this case. The last sentence though is interesting. What do you think? Force it or just leave it as is and deal with the small overhead?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobSpecificationService.java
tgianos: Additionally I definitely view both agent initiated and REST API initiated calls going to the `JobSpecificationServiceImpl` APIs on the server. I view the REST API path looking something like `API -> JobOrchestrator -> JobSpecificationServiceImpl` and inside whatever form that `JobOrchestrator` takes we'll be persisting things to the database (like the initial job metadata: `id`, `name`, `user`, etc). I also think the `resolveJobSpecification` API within this implementation will take care of persisting which cluster, command, etc was chosen for the job.

genie-agent/src/main/java/com/netflix/genie/agent/configs/ValidationConfig.java
mprimi: So should we remove `@Component`, `@Service` from all the beans instantiated that way and put *every* bean into a config in this config package?

genie-common/src/main/java/com/netflix/genie/common/dto/v4/CommonUserMetadata.java
mprimi: Mh, dunno. Sounds like they have a strong opinion against `Optional` here...

genie-common/src/main/java/com/netflix/genie/common/dto/v4/ExecutionEnvironment.java
mprimi: Makes sense.

genie-proto/src/main/proto/genie.proto
mprimi: 👍

genie-agent/src/main/java/com/netflix/genie/agent/configs/ValidationConfig.java
tgianos: You've hit on the conundrum I keep going back and forth on with this stuff. Originally with Genie 3 development I took a hard line and said we should put all `bean` definitions in config files to remove and dependency on frameworks from our class definitions leaving them as POJO's. You can see remnants of this with `ServicesConfig` in `genie-web`.

This is fine and well till you start trying to use more features of frameworks beyond just DI. For example you start using `@Transactional` or `@Validated` or `@EventListener`. Well now for these annotations to have any impact they need to be beans why have the bean definitions in a config file for separation of concerns when it's obvious you're relying on the framework anyway?

This is a weird topic that is bleeding through into this PR so I'm going to remove moving this file for now. We need to come to a convention for how we want to do things within our project. My initial opinion is to do something like:

If only an internal class just mark it with the proper `Stereotype` along with any conditionals if we envision it being overridden by someone. For anything we're overriding from the framework itself (validation, resource handlers, etc) that aren't just our classes we should put them in `Config` classes in a central locations. We can discuss though

PR #703, opened 2018-03-07
Agent job setup, launch and more

mprimi:


genie-common/src/main/java/com/netflix/genie/common/dto/v4/JobRequest.java
```diff
@@ -53,6 +55,8 @@
     private final JobMetadata metadata;
     @Valid
     private final ExecutionResourceCriteria criteria;
+    @Valid @NotNull
```
tgianos: Pretty sure `@Valid` already applies `@NonNull` but could double check. Also I thought we said `@Valid` on file did nothing?

genie-proto/src/test/groovy/com/netflix/genie/proto/v4/adapters/JobSpecificationServiceAdapterSpec.groovy
```diff
@@ -416,6 +420,7 @@ class JobSpecificationServiceAdapterSpec extends Specification {
         def jobSpecification = com.netflix.genie.proto.JobSpecification
                 .newBuilder()
                 .setIsInteractive(interactive)
+                .setJobDirectoryLocation(jobDirectoryLocation.getPath())
```
tgianos: `getPath` or `getAbsolutePath` I'm never quite sure

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
```diff
@@ -44,4 +48,60 @@
      * @throws RuntimeException if the agent id is already set
      */
     void setAgentId(@NotBlank final String agentId);
+
+    /**
+     * Set the job process once it has been launched.
+     *
+     * @param jobProcess a process handle for the children job
+     */
+    void setJobProcess(final Process jobProcess);
```
tgianos: Does this whole class need to be an interface? Seems to me that we're only ever going to have one implementation and it would save a lot of this boiler plate as you could just use Lombok `@Data`

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Process launchProcess() throws JobLaunchException {
+        final JobSpecification jobSpec = executionContext.getJobSpecification();
+
+        if (jobSpec == null) {
```
tgianos: So this assumes ExecutionContext is created before job spec is resolved and thus could be null. I think the execution context should either return `Optional<JobSpecification>` or we maybe add some sort of execution state flag into the context so we know it's already resolved?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        // Configure job running directory
+        processBuilder.directory(jobDirectory);
+
+        final Map<String, String> currentEnvironmentVariables = processBuilder.environment();
+        final Map<String, String> jobEnvironmentVariables = executionContext.getJobEnvironment();
+
+        // Validate additional environment variables
+        if (jobEnvironmentVariables == null) {
```
tgianos: Prefer Optional

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+        final ProcessBuilder processBuilder,
+        final JobSpecification jobSpec
+    ) throws JobLaunchException {
+        final List<String> commandLineArguments = jobSpec.getCommandArgs();
+
+        // Validate arguments
+        if (commandLineArguments == null) {
+            throw new JobLaunchException("Job command-line arguments not set in execution context");
+        } else if (commandLineArguments.isEmpty()) {
+            throw new JobLaunchException("Job command-line arguments are empty");
```
tgianos: I think this is fine and shouldn't throw an exception. The reason I made commandArgs as a non-required field in Genie 3.3 and soon 4.0 was cause I could imagine a command that is just like `/spark-shell` and no command line arguments are required to dump you into a default shell. In the past we had to get around this but putting a space in as the only command line argument which was just a waste.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+        // Configure arguments
+        processBuilder.command(commandLineArguments);
+    }
+
+    private void configureIO(
+        final ProcessBuilder processBuilder,
+        final JobSpecification jobSpec
+    ) {
+        if (jobSpec.isInteractive()) {
+            processBuilder.inheritIO();
```
tgianos: Have we double checked with @danielcweeks and @rdblue  that this is the behavior they want? If I'm not mistaken now there won't be any data in stdout and stderr when a job is interactive.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/LaunchJobAction.java
```diff
@@ -35,13 +40,55 @@
...
+        }
+
+        log.info("Job process started (pid: {})", getPid(jobProcess));
+
+        executionContext.setJobProcess(jobProcess);
+
         return Events.LAUNCH_JOB_COMPLETE;
     }
+
+    /* TODO: HACK, Process does not expose PID in Java 8 API */
```
tgianos: Since this isn't available in Java 8 what is the TODO here? Just to find a better way?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -35,13 +59,354 @@
...
+            } catch (final Exception e) {
+                throw new SetUpJobException("Failed to create directory: " + path, e);
+            }
+        }
+    }
+
+    private DownloadManifest createDownloadManifest(
+        final File jobDirectory,
+        final JobSpecification jobSpec,
+        final List<File> setupFiles
```
tgianos: Why pass this in separately when it's already part of the job spec?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/CleanupJobAction.java
```diff
@@ -44,7 +46,27 @@
...
         log.info("Cleaning up job...");
-        //TODO implement this action
+
+        final Deque<StateAction> deque = executionContext.getCleanupActions();
+        while (!deque.isEmpty()) {
+            final StateAction nextAction = deque.pop();
+
+            if (nextAction == this) {
+                // Skip self
+                continue;
```
tgianos: Is it even possible that there's a state after the current one? Should we break instead of continue?

genie-agent/src/test/java/com/netflix/genie/agent/execution/exceptions/ExecutionExceptionsTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+
+/**
+ * Parametrized test class to test all (trivial) exception classes in this package.
+ */
+@RunWith(Parameterized.class)
```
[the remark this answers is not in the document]
mprimi: Switched to it. Like it more than Parametrized.

genie-common/src/main/java/com/netflix/genie/common/jobs/JobConstants.java
```diff
@@ -157,6 +158,24 @@
     public static final String GENIE_ENV_PATH = "/genie/logs/env.log";
 
     /**
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
mprimi: Fixed

genie-common/src/main/java/com/netflix/genie/common/jobs/JobConstants.java
```diff
@@ -15,7 +15,7 @@
  *     limitations under the License.
  *
  */
-package com.netflix.genie.web.jobs;
+package com.netflix.genie.common.jobs;
```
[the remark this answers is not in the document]
mprimi: So something like `common-internal` that `web` and `agent` depend on, but not `client`.
I'm in favor of this, but it seems to be beyond the scope of the agent, it's something that affects more the client than anything else, no?

On the other side, if i ever need to move anything major into common that we don't want users to see, i'll create it.
(these are just constant strings, there's not much harm in having them visible no? I agree it would be better not to)

genie-agent/src/main/java/com/netflix/genie/agent/cli/CacheArgumentsImpl.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Getter
+@Component
+class CacheArgumentsImpl implements ArgumentDelegates.CacheArguments {
+
+    @VisibleForTesting
+    static final String DEFAULT_CACHE_PATH = "/tmp/genie-cache";
```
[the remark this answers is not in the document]
mprimi: 👍

genie-common/src/main/java/com/netflix/genie/common/jobs/JobConstants.java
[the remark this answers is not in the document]
tgianos: Yes something like that. Metacat had something like `server-common` and since we're so great at naming things I'm sure we could come up with something haha. 

We could reintroduce a `genie-core` like we had before.

genie-agent/src/main/java/com/netflix/genie/agent/utils/EnvUtils.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Utilities for shell environment variables.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public final class EnvUtils {
+
+    private static Pattern singleLineValuePattern = Pattern.compile("^(.+)='(.*)'$");
```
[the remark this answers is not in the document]
mprimi: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
mprimi: It depends, does args include the executable?
I assumed it did, since it's not sent back in the job spec in any other place.
(And I think it's good to have the entire command-line in one parameter in the spec object, as opposed to have the binary somewhere else hard to find)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
mprimi: Inherit means that it's inherited from the terminal, so if a command prints 'foo', the user sees 'foo' in the terminal. 
If the user writes 'bar', it will be forwarded to the process.

It is the simplest behavior for interactive commands.

If we do also want to intercept in/out/err and log it to a file, then we need to intercept this and it'll be more complex.
But in terms of M1, to reach parity with the Runner, I'm fairly sure this is the right setting.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/LaunchJobAction.java
mprimi: I have to do this hack because Java 8 does not expose PID.
Java 9 does, so in the future we may clean this up and just call `getPid()`

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
mprimi: `setupFiles` is an output parameter.
As I'm creating the manifest, I'm also noting down the setup files (since the latter are later sourced and used to build the process environment).

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/CleanupJobAction.java
mprimi: I wanted to avoid this state having knowledge of the rest of the machine topology.
It will execute cleanup for whatever is in the queue.
It knows that it can skip it's own cleanup, that's its prerogative.

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
mprimi: Just trying to make implementation classes package-local. If I remove the interface I need to make this public (and same for other classes with a single implementation for which I created an interface, for example all Services)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
tgianos: oh this is a good point. I totally missed doing this on the server side of the job spec. Let me fix that bug. I'll pass it back in this command args list then this check is accurate. The arguments should NEVER be null cause the DTO doesn't return a null though.

tgianos: #704 addresses this

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadManifest.java
```diff
@@ -0,0 +1,106 @@
...
+     * Get the set of URI s (source files) that were added to the manifest.
+     * @return an immutable set of URIs
+     */
+    Set<URI> getSourceFileUris();
+
+    /**
+     * Get the map entries: URI source -> File target.
+     * @return an immutable set of map entries
+     */
+    Set<Map.Entry<URI, File>> getEntries();
```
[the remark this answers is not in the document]
mprimi: I didn't want to expose that it's a map internally, because the implication of that is that you cannot have the same URL downloaded to two different locations.
I think right now it's ok to have that limitation, but if we ever need to change, exposing this set is more future-proof.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadManifestImpl.java
```diff
@@ -0,0 +1,130 @@
...
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Builder addFileWithTargetDirectory(
+            final URI sourceFileUri,
+            final File targetDirectory
+        ) {
+            final String uriPath = sourceFileUri.getPath();
```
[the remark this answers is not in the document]
mprimi: Here I'm extracting the filename from a URL to decide what to name the local file, so I don't think that class would help (since it works with files)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    DownloadServiceImpl(final FetchingCacheService fetchingCacheService) {
+        this.fetchingCacheService = fetchingCacheService;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public DownloadManifest.Builder newManifestBuilder() {
+        return new DownloadManifestImpl.Builder();
```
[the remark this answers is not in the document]
mprimi: I don't like this either. The idea was that a download service implementation provides and works with its own manifest implementation.

But better is that download manifest works with any manifest without knowing any implementation.
But in that case I don't know how to provide a manifest builder instance without exposing its implementation (i.e. making it public).
A prototype bean would work, but don't like it much. Maybe a manifest builder factory bean?
But it seems a bit over-engineered for returning a simple builder class.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void download(final DownloadManifest downloadsManifest) throws DownloadException {
+
+        // Validate all destination directories exist
+        for (final File targetDirectory : downloadsManifest.getTargetDirectories()) {
```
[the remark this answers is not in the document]
mprimi: The idea was that the download service does one thing: download.
If you tell it where to download stuff for you, it's your responsability to create those directories.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        // Validate all target files don't exist
+        for (final File targetFile : downloadsManifest.getTargetFiles()) {
+            if (targetFile.exists()) {
+                throw new DownloadException("Target file exists: " + targetFile.getAbsolutePath());
+            }
+        }
+
+        final Map<URI, File> cachedFiles = fetchingCacheService.getAll(downloadsManifest.getSourceFileUris());
```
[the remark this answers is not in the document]
mprimi: Indeed, in the current implementation there's no version (it won't detect and download a new version of a cached file) and it won't detect partial/corrupted downloads either.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        for (final Map.Entry<URI, File> uriFileEntry : downloadsManifest.getEntries()) {
+
+            final File cachedFile = cachedFiles.get(uriFileEntry.getKey());
+
+            if (cachedFile == null || !cachedFile.exists()) {
+                throw new DownloadException("Failed get file handle from cache: " + uriFileEntry.getKey());
+            }
+
+            try {
+                Files.copy(cachedFile, uriFileEntry.getValue());
```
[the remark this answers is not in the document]
mprimi: You're probably right.
The cache owns the files, giving out handles to them makes it hard to for example garbage-collect old versions because who knows what is using them.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
mprimi: Same as above, shouldn't be possible to get here with this not set.
(and if we ever do, this is a nicer error than an NPE)

genie-agent/src/test/java/com/netflix/genie/agent/utils/PathUtilsTest.java
```diff
@@ -0,0 +1,155 @@
...
+public class PathUtilsTest {
+
+    private static final String JOB_DIR_STRING = "/tmp/genie-job";
+    private static final File JOB_DIR_FILE = new File(JOB_DIR_STRING);
+
+    /**
+     * Test compose path.
+     */
+    @Test
+    @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
```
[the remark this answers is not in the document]
mprimi: Moved to Spock DDT (and FindBug does not complain)

genie-agent/src/test/resources/EnvUtilsTest1.txt
```diff
@@ -0,0 +1,259 @@
+AUTOSOURCEPLUS_BASEDIR='/Users/mprimi/Dropbox/sourceplus/Descartes'
+Apple_PubSub_Socket_Render='/private/tmp/com.apple.launchd.3zXrbJbcaT/Render'
+COLORFGBG='7;0'
+COLORTERM='truecolor'
+COLUMNS='232'
+COMMAND_MODE='unix2003'
+EDITOR='/usr/local/bin/mate -w'
+GENIE_FOO='foo'
+HOME='/Users/mprimi'
```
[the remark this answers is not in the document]
mprimi: Mhmm, yeah going to do some cleanup...

genie-common/src/main/java/com/netflix/genie/common/dto/v4/JobRequest.java
mprimi: From my understanding, `@Valid` enabled validation and `@NotNull` is what the validation checks for (since there is no file-specific validator.
I'll remove `@Valid`. It's _probably_ superfluous.

genie-proto/src/test/groovy/com/netflix/genie/proto/v4/adapters/JobSpecificationServiceAdapterSpec.groovy
mprimi: This should be `getAbsolutePath`, fixed.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
tgianos: Sounds good. We can address it if need be after M1

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/LaunchJobAction.java
tgianos: Ok so it will stay as is for now until we can move to Java 9+ (probably Java 11 as I believe it's the next LTS version)

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
tgianos: I'm fine. Was just trying to save on passing around more parameters if they were already present. It's a private method so not a big deal if it ever needs to be refactored.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/CleanupJobAction.java
tgianos: Makes sense I guess if you think cleanup could ever be called twice. No big deal I guess to keep it naive

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
tgianos: I think that services should have interfaces for sure but this is effectively a DTO I don't see the value of the added maintenance of an interface. Making these kinds of classes public I think is fine.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadManifest.java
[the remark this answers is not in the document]
tgianos: But it's a set of `Map.Entry` so you're still exposing internal representation. I don't think it's a huge deal either way just seemed like a useless API if you just have the getter for the map if someone wants the entry set they can get it directly.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadManifestImpl.java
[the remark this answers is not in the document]
tgianos: k

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadService.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public interface DownloadService {
+    /**
+     * Get a new manifest builder.
+     * @return a new builder
+     */
+    DownloadManifest.Builder newManifestBuilder();
```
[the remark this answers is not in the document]
tgianos: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
tgianos: I think if you're coupling them that tightly (1-1) you should just make the Implementation a private class of the download implementation. Again the download manifest I don't really see the need for Multiple implementations as it's only a DTO and should be immutable after it's built. Do we envision needing more than one implementation of this?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
tgianos: fair enough provided you make that expectation clear in the documentation for the service.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
tgianos: We should address this somehow going forward but I agree it doesn't need to be part of this PR

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
tgianos: So you're moving this?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+            } catch (final IOException e) {
+                throw new DownloadException("Failed to copy file from cache: " + uriFileEntry.getKey(), e);
```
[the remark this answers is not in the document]
tgianos: Sounds good

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
tgianos: I think my point was more about the contract (API) between the `ExecutionContext` and its consumers. In this case when I look at your API definition you don't have the return marked as `@Nullable` and you aren't returning `Optional` so there's no real signal outside the class that this value could be null without looking at hte source or understanding the entire flow of the system. I think one of those two above options should be present on APIs so that we can trust class behavior as we consume them around the system

genie-agent/src/main/java/com/netflix/genie/agent/utils/PathUtils.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param baseDirectory the base directory
+     * @param children path components
+     * @return a Path
+     */
+    public static Path composePath(
+        final File baseDirectory,
+        final String... children
+    ) {
+        File pathAccumulator = baseDirectory;
+        for (String child : children) {
```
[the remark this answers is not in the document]
tgianos: Fair enough though I still like the `Paths` option as it is maintained/tested by a larger community and probably tests cases we may miss

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
mprimi: Public vs package+interface is debatable, but this guy is not a DTO.
It does have behavior, it's mutable, it is not serializable, It is not sent to a remote interface, it's a singleton.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadManifest.java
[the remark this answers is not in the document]
mprimi: Doh! It should be a set of `Pair<URL, File>` or something to properly do the thing I was saying above.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
mprimi: No, like in other places, it's the only way to keep implementations package local. Which to me is worth the price of maintaining the extra interface (IDEA does all the work anyway, and when using the class, it's nicer to read)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
mprimi: Adding comment

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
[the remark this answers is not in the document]
mprimi: Yes

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/LaunchJobServiceImpl.java
mprimi: Oh, I see. I'll mark all fields `@Nullable`

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/NaiveFetchingCacheServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+
+        log.info("Preparing to fetch: {}", uriString);
+
+        final Resource resource = resourceLoader.getResource(uriString);
+
+        if (!resource.exists()) {
+            throw new DownloadException("Resource not found: " + uriString);
+        }
+
+        final String resourceCacheId = getResourceCacheId(sourceFileUri);
```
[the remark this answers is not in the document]
mprimi: Oh, so move this whole block up? I thought it was a comment on the line (which is required to check for existence)

genie-agent/src/main/java/com/netflix/genie/agent/utils/EnvUtils.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    public static Map<String, String> parseEnvFile(
+        final File environmentFile
+    ) throws IOException, ParseException {
+        return parseEnvStream(new FileInputStream(environmentFile));
+    }
+
+    /**
+     * Parse a file containing shell variable declarations.
+     * Expects all values to be single-quoted. Lines not conforming are silently skipped.
```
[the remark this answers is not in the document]
mprimi: I have not, i should crawl the genie api for `env.txt`.
Even if we don't find any tho, it's inconclusive because all the runner jobs cannot be looked at this way.

I'd go ahead without this for the moment, we can make a note to add "escaped single quote support" to the env parser later.

mprimi:
@tgianos addressed all minor changes brought up in review.
The only major change is in the commit 'Add DownloadService to agent' (not linking as SHA might change by the time you review).
It's now a single commit for manifest, download and cache service.
The cache is now responsible for copying files to the target location.
Also added one commit at the end to move all service implementations to the `'impl` package.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        try {
+            fetchingCacheService.get(downloadsManifest.getEntries());
+        } catch (final IOException e) {
+            throw new DownloadException("Failed to download", e);
+        }
+    }
+
+    @Getter
```
tgianos: Do you want `EqualsAndHasCode` for storing in collections and `ToString` for debugging?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/FetchingCacheService.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Download a given resource (if not already cached) and copy it to the specified destination.
+     *
+     * @param sourceFileUri the resource URI
+     * @param destinationFile the location on disk where to copy the resource for use
+     * @throws DownloadException if the resource is not found or fails to download
+     * @throws IOException if downloading or copying the file to destination fails
+     */
+    void get(final URI sourceFileUri, final File destinationFile) throws DownloadException, IOException;
```
tgianos: Any reason you don't only throw the `DownloadException` and catch the `IOException` internally? Are you going to do different things with them outside this service?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/FetchingCacheService.java
```diff
@@ -0,0 +1,56 @@
...
+import java.net.URI;
+import java.util.Set;
+
+/**
+ * Interface for a cache that downloads resources via URL and stores them on local disk for later reuse.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public interface FetchingCacheService {
```
tgianos: It seems to me like this is more of the `DownloadService` and the other thing is more of a `ExecutionEnvironmentSetup` service or similar. it's probably just naming semantics

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/NaiveFetchingCacheService.java
```diff
@@ -0,0 +1,150 @@
...
+        for (final Pair<URI, File> sourceDestinationPair : sourceDestinationPairs) {
+            get(sourceDestinationPair.getKey(), sourceDestinationPair.getValue());
+        }
+    }
+
+    private File lookupOrDownload(final URI sourceFileUri) throws DownloadException, IOException {
+
+        final String uriString = sourceFileUri.toASCIIString();
+
+        log.info("Lookup: {}", uriString);
```
tgianos: Debug?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/FetchingCacheService.java
mprimi: I just thought it could come in handy to distinguish problem downloading vs. problem writing the file for users of this class.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/FetchingCacheService.java
mprimi: Download service downloads file where you tell it to, does not know anything about environments or execution.
And this is just a very simple on-disk file cache, it does not only download but retrieve any resource (i.e. from classpath, other places on the filesystem)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/NaiveFetchingCacheService.java
mprimi: I guess. Need another pass at this to add metrics anyway, going to merge as-is.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/DownloadServiceImpl.java
mprimi: I don't expect to have collections of manifests, tho wouldn't hurt to add those.
`ToString` would be nice to print the map as JSON...

PR #704, opened 2018-03-07
Adding missing command executable to command args sent back with job specification

tgianos:


mprimi:
Nitpick: test does not cover the case of 'binary' getting split.

PR #710, opened 2018-03-16
Export the job tags into runtime environment

sumitnetflix:


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
```diff
@@ -380,6 +382,21 @@ void createJobEnvironmentVariables(
 
         // Append new line
         writer.write(LINE_SEPARATOR);
+
```
mprimi: Double newline

mprimi:
Also read this about commit message: https://chris.beams.io/posts/git-commit/

A bit subjective but please keep the format with short subject + newline + details (optional)

And definitely remove the JIRA link since this is an OSS project.

tgianos:
I would have expected required changes to have been made to these two files: 
[linux-runsh.txt](https://github.com/Netflix/genie/blob/3.3.x/genie-web/src/test/resources/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests/linux-runsh.txt)
[non-linux-runsh.txt](https://github.com/Netflix/genie/blob/3.3.x/genie-web/src/test/resources/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests/non-linux-runsh.txt)

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/InitialSetupTask.java
mprimi: This is still here. No big deal, but if you squash you could take care of this.

mprimi:
Nitpick on commit message, altough we are "exporting" in bash lingo, the thing we're really doing is "exposing" that variable to the job.
And the "runtime" could mean server runtime, but it's actually the job runtime.
So I think "Expose job tags in job runtime environment" could have conveyed more clearly the change.
Don't bother updating, just something to keep in mind so we're all on the same language/terms.

PR #711, opened 2018-03-19
Enable file downloads within Genie Agent

tgianos:


genie-agent/build.gradle
```diff
@@ -10,10 +10,13 @@ dependencies {
 
     // Spring
     compile("org.springframework.boot:spring-boot-starter")
+    compile("org.springframework.cloud:spring-cloud-aws-context")
+    compile("org.springframework.cloud:spring-cloud-starter-aws")
```
tgianos: @mprimi I think we only need to pull this in and it will pull your other AWS dependencies you've added here in transitively? Can check and fix in another commit.

mprimi: The STS one was not coming in transitively, but not even sure you need it. 
Not sure what's the minimum set you need, i added those to the patch just to save you a look up but we may need to revisit/change.

tgianos:
@mprimi I've conditionally enabled the AWS functionality whether the agent is run on a cloud instance or not. We should revisit this as there's not really a good reason this couldn't run on a non-cloud instance with some other way of getting instance credentials.

I've committed it like this to get us unblocked on the build/pylon for M1 but we should come back and refine this to conditionally enable what we want. Perhaps some of these issues @twicksell has already solved but if you look through the AWS cloud autoconfigurations there's no nice way to disable them all via property so we almost have to go one by one.

PR #713, opened 2018-03-21
Make job tags order deterministic in tests

sumitnetflix:


genie-web/src/test/java/com/netflix/genie/web/controllers/JobRestControllerIntegrationTests.java
```diff
@@ -1376,6 +1377,12 @@ private String getExpectedRunContents(
         final String jobWorkingDir,
         final String jobId
     ) {
+        /* Sort the job tags in the same order as they are created in the run script by the server.
+         * Else test might fail sometimes.
+         */
+        final List<String> sortedJobTags = Arrays.asList(JOB_TAG1, JOB_TAG2);
```
tgianos: You can also use `Lists.newArrayList(JOB_TAG1, JOB_TAG2)` from Guava if you want to skip the Array middleman here. Just FYi going forward

PR #715, opened 2018-03-26
Expose job tags in job runtime environment

sumitnetflix:


genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobSpecificationServiceImplSpec.groovy
```diff
@@ -224,7 +224,7 @@ class JobSpecificationServiceImplSpec extends Specification {
                 false,
                 null,
                 false,
-                new JobMetadata.Builder(jobName, UUID.randomUUID().toString()).build(),
+                new JobMetadata.Builder(jobName, UUID.randomUUID().toString()).withTags(Sets.newHashSet(UUID.randomUUID().toString(), UUID.randomUUID().toString())).build(),
```
tgianos: I'd just say to improve this layout as it's hard to read and tell that you're calling the `with` method

sumitnetflix: Layout fixed

PR #718, opened 2018-04-03
Bypass server resolution, perform shell variable expansion and more

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
```diff
@@ -156,4 +156,11 @@
         converter = ArgumentConverters.JSONConverter.class
     )
     private JsonNode jobMetadata = GenieObjectMapper.getMapper().createObjectNode();
+
+    @Parameter(
+        names = {"--jobSpecificationFile"},
+        description = "The local file where to load a job specification from, rather than resolving via server",
```
tgianos: How do you envision using this in production vs. the `getSpecification` one?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -310,12 +313,41 @@ private void addEntitiesFilesToManifest(
         }
     }
 
+
+    private Map<String, String> createAdditionalEnvironmentMap(
+        final File jobDirectory,
+        final JobSpecification jobSpec
+    ) {
+        final Map<String, String> envMap = Maps.newHashMap();
```
tgianos: You could also use `ImmutableMaps.Builder<String, String> envMap = ImmutableMaps.builder()` which might be nicer than wrapping your said map in the collections unmodifiable interface

genie-agent/src/main/java/com/netflix/genie/agent/utils/EnvUtils.java
```diff
@@ -172,6 +180,52 @@ private EnvUtils() {
...
+     * Performs shell environment variables expansion on the given string.
+     *
+     * @param inputString a string, possibly containing shell variables
+     * @param environmentVariablesMap a map of string environment variables
+     * @return a new string with variables expanded
+     * @throws VariableSubstitutionException if the value for a variable appearing in input is not found in the map
+     */
+    public static String expandShellVariables(
+        final String inputString,
+        final Map<String, String> environmentVariablesMap
```
tgianos: So the problem here I could see eventually is that the `environmentVariablesMap` is unordered. A lot of our variable usage is cascading (e.g. `GENIE_JOB_DIRECTORY` ends up being used a lot or some env could be set by someone else in a setup script and we need it to be defined before others can be evaluated. We may need to use a linked hash map or something instead of a generic map?

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
mprimi: When we submit jobs via API and start the agent inside a container we could use this.
It would ensure the spec is created by the server before we even attempt to fork a container, and once the agent comes up we avoid having to resolve again.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
mprimi: Didn't know about that builder, will change

mprimi:
@tgianos Switched to `ImmutableMap.Builder`, the rest is unchanged

PR #719, opened 2018-04-03
Tag job execution timers with cluster and command parameters

sumitnetflix:


genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ClusterTask.java
```diff
@@ -76,8 +75,9 @@ public void executeTask(@NotNull final Map<String, Object> context) throws Genie
             final JobExecutionEnvironment jobExecEnv =
                 (JobExecutionEnvironment) context.get(JobConstants.JOB_EXECUTION_ENV_KEY);
             final Cluster cluster = jobExecEnv.getCluster();
-            tags.put(MetricsConstants.TagKeys.CLUSTER_NAME, cluster.getName());
-            tags.put(MetricsConstants.TagKeys.CLUSTER_ID, cluster.getId().orElse(NO_ID_FOUND));
+
+            MetricsUtils.addCommonMetricTags(jobExecEnv, tags, NO_ID_FOUND);
```
tgianos: Since `MetricsUtils` is used in many places this method name leads me to believe it is a program wide set of `commonMetricTags` when it really is just `addCommonJobWorkflowMetricTags` or something like that

genie-core/src/main/java/com/netflix/genie/core/services/JobSubmitterService.java
```diff
@@ -40,6 +40,13 @@
 @Validated
 public interface JobSubmitterService {
 
+    /**
+     * Constant representing a no id found for atlas metrics.
+     * E.g this values is used for a missing cluster id when tagging
+     * atlas metrics by cluster id
+     */
+    String NO_ID_FOUND = "<no id>";
```
tgianos: Since this is related to metrics I'd put it in `MetricsConstants`

genie-core/src/main/java/com/netflix/genie/core/services/impl/LocalJobRunner.java
```diff
@@ -132,9 +138,13 @@ public void submitJob(
         @Min(value = 1, message = "Memory can't be less than 1 MB") final int memory
     ) throws GenieException {
         final long start = System.nanoTime();
+        final Map<String, String> tags = MetricsUtils.newSuccessTagsMap();
 
         try {
             log.info("Beginning local job submission for {}", jobRequest);
+
+            MetricsUtils.addCommonMetricTags(cluster, command, tags, NO_ID_FOUND);
```
tgianos: Same naming issue here

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
```diff
@@ -80,4 +83,44 @@ private MetricsUtils() {
...
+     * execution environment to the map of tags passed in.
+     * Populates defaultIdValue in case ids are not present in the job execution environment
+     * E.g defaultIdValue will be added to the tags maps for the cluster_id key in case cluster_id
+     * does not exist in jobExecEnv
+     *
+     * @param tags The map of tags
+     * @param jobExecEnv Genie job execution env
+     * @param defaultIdValue Default value for missing id in jobExecEnv
+     */
+    public static void addCommonMetricTags(final JobExecutionEnvironment jobExecEnv,
```
tgianos: For multiple arguments here that don't fit on one line within the character limit I like to put each parameter on a new line. It makes it easier (at least for me) to read it so I don't miss parameters buried in a list an I feel it looks a bit nicer

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
```diff
@@ -80,4 +83,44 @@ private MetricsUtils() {
...
     }
+
+    /**
+     * Convenience method to add common tags for slicing and dicing atlas metrics from Job
+     * execution environment to the map of tags passed in.
+     * Populates defaultIdValue in case ids are not present in the job execution environment
+     * E.g defaultIdValue will be added to the tags maps for the cluster_id key in case cluster_id
+     * does not exist in jobExecEnv
+     *
+     * @param tags The map of tags
```
tgianos: In intellij `cmd + option + L` auto formats the code and will space out these parameter documentation nicely. I only mention this cause I tend to do this habitually as I'm in files which leads to stupid modifications on files which really aren't code changes but formatting. Better to do it up front.

genie-core/src/test/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImplUnitTests.java
```diff
@@ -1340,4 +1354,46 @@ private JobMetadata getJobMetadata() {
...
+     * Verify the presence of failure tags.
+     * @param argumentCaptor arguments captured in the unit test
+     * @param throwable expected exception
+     */
+    private void verifyFailureTags(final ArgumentCaptor<Map<String, String>> argumentCaptor,
+                                   final Throwable throwable) {
+        final Map<String, String> failureTags = MetricsUtils.newFailureTagsMapForException(throwable);
+
+        Assert.assertTrue(argumentCaptor.getValue().get(MetricsConstants.TagKeys.STATUS).
+            equals(failureTags.get(MetricsConstants.TagKeys.STATUS)));
```
mprimi: If you use `assertEquals(x, y)` rather than `assertTrue(x.equals(y))` the error messages in case of failed assertion will actually show the two things you're comparing and provide a more helpful message.

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
mprimi: `CommonMetric` is a vague name. This helper is adding tags related to one specific thing: the job execution environment. Why not have a name that reflect this?

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
```diff
@@ -80,4 +83,44 @@ private MetricsUtils() {
...
+     * Populates defaultIdValue in case ids are not present in the objects passed in
+     * E.g defaultIdValue will be added to the tags maps for the cluster_id key in case cluster_id
+     * does not exist in Cluster object
+     *
+     * @param tags The map of tags
+     * @param cluster Cluster chosen for running the command
+     * @param command Command chosen to be executed
+     * @param defaultIdValue Default value for missing id in jobExecEnv
+     */
+    public static void addCommonMetricTags(final Cluster cluster, final Command command,
```
mprimi: Same for this, it's adding a specific set of tags, the name of this method could reflect this.

genie-core/src/main/java/com/netflix/genie/core/jobs/workflow/impl/ClusterTask.java
```diff
@@ -56,10 +55,10 @@
      * Constructor.
      *
      * @param registry The metrics registry to use
-     * @param fts File transfer service
+     * @param fts      File transfer service
      */
     public ClusterTask(@NotNull final Registry registry,
```
mprimi: This method is not following our standard indentation style.
Since you're re-indenting it would have been nice to fix. (Not worth amending this commit just for this, leave as is is ok, just keep in mind for the future).

sumitnetflix: I noticed that but this change in indentation happened by the formatter(command option L), so I left it as such

mprimi:
@sumitnetflix Ah, crap, I missed this but there's no reason for this to be in 2 commits, the second fixing something that was introduced in the first. Please squash next time, this should be one atomic change.

sumitnetflix:
@mprimi Consolidation of NO_ID_FOUND is a different change than the original change to tag metric. Had discussed with @tgianos . Hence added 2 commits.

PR #720, opened 2018-04-04
Upgrade to Flyway 4.2.0

tgianos:
Primary purpose of this PR is to upgrade Flyway to 4.2.0 from 3.x. This helps prepare for upgrading to Flyway 5.x with Genie 4 as testing revealed you can't go right from 3.x to 5.x.

See [Boot 2.0 migration guide](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#flyway)


tgianos:
@danielbwatson FYI this did turn out to be an issue

@twicksell FYI in case others face this internally

gradle.properties
```diff
@@ -25,6 +25,8 @@ spring_platform_version=Brussels-SR7
 
 ## Override Spring Platform IO Versions
 
+flyway.version=4.2.0
```
mprimi: Is this property automagically used by gradle/nebula?

tgianos: It's used by the Spring dependency management plugin to swap in the version controlled by their boms. See [here](https://docs.spring.io/spring-boot/docs/2.0.0.RELEASE/gradle-plugin/reference/html//#managing-dependencies-customizing)

PR #723, opened 2018-04-06
Add cluster and command tags on timing metrics at job completion

sumitnetflix:


genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -154,6 +156,19 @@ void handleJobCompletion(final JobFinishedEvent event) throws GenieException {
...
+            /*
+             * Get the cluster and command chosen for the job for tagging metrics.
+             * Swallowing any exception in getting those as this step is not
+             * critical and should not show that the entire job failed
+             */
+            try {
+                final Cluster cluster = jobSearchService.getJobCluster(jobId);
+                final Command command = jobSearchService.getJobCommand(jobId);
+                MetricsUtils.addCommonJobWorkflowMetricTags(cluster, command, tags);
+            } catch (final Exception e) {
```
tgianos: I think it's cleaner if you just handle the error and always put a tag. This way you can at least put a tag like "NO CLUSTER FOUND" or pass null to your utility method representing a cluster doesn't exist. This way you could so something like 
```
final Cluster cluster;
final Command command;

try {
    cluster = this.jobSearchService.getJobCluster(jobId);
} catch (final Exeption e) {
    log.warn("Unable to tag metrics with cluster due to {}, e.getMessage(), e);
    cluster = null;
}

try {
    command = this.jobSearchService.getJobCommand(jobId);
} catch (final Exeption e) {
    log.warn("Unable to tag metrics with command due to {}, e.getMessage(), e);
    command = null;
}

MetricsUtils.addCommonJobWorkflowMetricTags(cluster, command, tags);
```

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -154,6 +156,19 @@ void handleJobCompletion(final JobFinishedEvent event) throws GenieException {
...
         try {
+            /*
+             * Get the cluster and command chosen for the job for tagging metrics.
+             * Swallowing any exception in getting those as this step is not
+             * critical and should not show that the entire job failed
+             */
+            try {
+                final Cluster cluster = jobSearchService.getJobCluster(jobId);
+                final Command command = jobSearchService.getJobCommand(jobId);
+                MetricsUtils.addCommonJobWorkflowMetricTags(cluster, command, tags);
```
tgianos: Calling this method from outside the Job Workflow now doesn't make sense for this name...

sumitnetflix: In that case I will rename the method to something like addResourceTags. Does that make sense @tgianos ?

genie-core/src/main/java/com/netflix/genie/core/util/MetricsConstants.java
```diff
@@ -102,6 +102,26 @@ private TagKeys() {
...
+         * E.g this values is used for a missing cluster id when tagging
+         * metrics by cluster id
+         */
+        public static final String NO_ID_FOUND = "<no id>";
+
+        /**
+         * Constant representing no cluster found when tagging metrics
+         * with cluster tags.
+         */
+        public static final String NO_CLUSTER_FOUND = "NO_CLUSTER_FOUND";
```
mprimi: Nitpick: "No cluster found" makes it sounds like something went wrong.
In reality either we have a cluster or we don't, and in case we don't it's expected, so `"NO CLUSTER"` would be a little more readable by someone looking at metrics that is not familiar with the details. 
Also no need to underscore the literal.

genie-core/src/main/java/com/netflix/genie/core/util/MetricsConstants.java
```diff
@@ -102,6 +102,26 @@ private TagKeys() {
...
+
+        /**
+         * Constant representing no command found when tagging metrics
+         * with command tags.
+         */
+        public static final String NO_COMMAND_FOUND = "NO_COMMAND_FOUND";
```
mprimi: Same here

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
```diff
@@ -94,21 +86,25 @@ private MetricsUtils() {
...
      * does not exist in jobExecEnv
      *
      * @param tags       The map of tags
      * @param jobExecEnv Genie job execution env
      */
-    public static void addCommonJobWorkflowMetricTags(final JobExecutionEnvironment jobExecEnv,
-                                                      final Map<String, String> tags) {
-        addCommonJobWorkflowMetricTags(jobExecEnv.getCluster(), jobExecEnv.getCommand(), tags);
+    public static void addClusterAndCommandTags(final JobExecutionEnvironment jobExecEnv,
+                                                final Map<String, String> tags) {
```
mprimi: Following code convention for the rest of this codebase, the `) {` goes on a newline

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
```diff
@@ -117,16 +113,30 @@ public static void addCommonJobWorkflowMetricTags(final JobExecutionEnvironment
      * @param cluster Cluster chosen for running the command
      * @param command Command chosen to be executed
      */
-    public static void addCommonJobWorkflowMetricTags(final Cluster cluster,
-                                                      final Command command,
-                                                      final Map<String, String> tags) {
+    public static void addClusterAndCommandTags(final Cluster cluster,
+                                                final Command command,
+                                                final Map<String, String> tags) {
```
mprimi: Same here

genie-core/src/main/java/com/netflix/genie/core/util/MetricsUtils.java
```diff
@@ -117,16 +113,30 @@ public static void addCommonJobWorkflowMetricTags(final JobExecutionEnvironment
...
         /* Add cluster tags */
-        tags.put(MetricsConstants.TagKeys.CLUSTER_NAME, cluster.getName());
-        tags.put(MetricsConstants.TagKeys.CLUSTER_ID, cluster.getId().orElse(NO_ID_FOUND));
+        tags.put(MetricsConstants.TagKeys.CLUSTER_NAME,
+            cluster != null
+                ? cluster.getName()
+                : MetricsConstants.TagValues.NO_CLUSTER_FOUND);
```
mprimi: Same for `MetricsConstants.TagKeys....` here and the closing `);` here and in the rest of this block

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -154,6 +156,29 @@ void handleJobCompletion(final JobFinishedEvent event) throws GenieException {
...
+             * Swallowing any exception in getting those as this step is not
+             * critical and should not show that the entire job failed
+             */
+            Cluster cluster;
+            Command command;
+
+            try {
+                cluster = this.jobSearchService.getJobCluster(jobId);
+            } catch (final Exception e) {
+                log.warn("Unable to tag metrics with cluster due to {}", e.getMessage(), e);
```
mprimi: slf4j does some of this for you, you can simplify this to: `log.warn("Unable to tag metrics with cluster", e);`

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
```diff
@@ -154,6 +156,29 @@ void handleJobCompletion(final JobFinishedEvent event) throws GenieException {
         final Map<String, String> tags = MetricsUtils.newSuccessTagsMap();
 
         try {
+            /*
+             * Get the cluster and command chosen for the job for tagging metrics.
+             * Swallowing any exception in getting those as this step is not
+             * critical and should not show that the entire job failed
+             */
+            Cluster cluster;
```
mprimi: Strange, you don't get a warning that these can be final?

genie-web/src/test/groovy/com/netflix/genie/web/tasks/job/JobCompletionServiceSpec.groovy
```diff
@@ -121,12 +124,8 @@ class JobCompletionServiceSpec extends Specification {
                 { throw new GenieServerException("null") } >>
                 { throw new GenieServerException("null") } >>
                 { throw new GenieServerException("null") }
-        completionTimerId.withTags(MetricsUtils.newFailureTagsMapForException(new GenieServerException("null")))
```
mprimi: If you are expecting this to be called on your timer, you should prefix it with `1 *`

genie-web/src/test/groovy/com/netflix/genie/web/tasks/job/JobCompletionServiceSpec.groovy
```diff
@@ -149,10 +146,8 @@ class JobCompletionServiceSpec extends Specification {
                 .withId(jobId).withStatus(JobStatus.RUNNING).withCommandArgs(COMMAND_ARGS).build()
         1 * jobPersistenceService.updateJobStatus(jobId, _, _)
         1 * completionTimer.record(_, TimeUnit.NANOSECONDS)
-        timerTagsCapture == ImmutableMap.of(
-                MetricsConstants.TagKeys.STATUS, MetricsConstants.TagValues.SUCCESS,
-                JobCompletionService.JOB_FINAL_STATE, JobStatus.FAILED.toString()
-        )
+        verifySuccessTags(timerTagsCapture) \
+         && timerTagsCapture.get(JobCompletionService.JOB_FINAL_STATE).equals(JobStatus.FAILED.toString())
```
mprimi: No need to AND
Place one condition you expect to return `true` per line, this produces the most clear output when one of the assertions fails.

genie-web/src/test/groovy/com/netflix/genie/web/tasks/job/JobCompletionServiceSpec.groovy
```diff
@@ -185,11 +180,53 @@ class JobCompletionServiceSpec extends Specification {
...
         3 * errorCounter.increment()
+
+        when:
+        jobCompletionService.handleJobCompletion(new JobFinishedEvent(jobId, JobFinishedReason.PROCESS_COMPLETED, "null", this))
+        then:
+        1 * jobSearchService.getJob(jobId) >> new Job.Builder(NAME, USER, VERSION)
```
mprimi: You are reusing identical `Command` and `Cluster`, creating them once in the `setup:` section would make this test easier to read and maintain

genie-web/src/test/groovy/com/netflix/genie/web/tasks/job/JobCompletionServiceSpec.groovy
```diff
@@ -224,4 +261,50 @@ class JobCompletionServiceSpec extends Specification {
...
+     * @param tags tags to be verified
+     * @param throwable exception thrown on failure
+     * @return true if all failure tags exist
+     */
+    private Boolean verifyFailureTags(Map<String, String> tags,
+                                      final Throwable throwable) {
+        final Map<String, String> failureTags = MetricsUtils.newFailureTagsMapForException(throwable);
+
+        return tags.get(MetricsConstants.TagKeys.STATUS)
+                .equals(failureTags.get(MetricsConstants.TagKeys.STATUS)) \
```
mprimi: Why create a whole new map just to check the value of a key matches, as opposed to just use the constant that you expect?

genie-web/src/test/groovy/com/netflix/genie/web/tasks/job/JobCompletionServiceSpec.groovy
```diff
@@ -224,4 +261,50 @@ class JobCompletionServiceSpec extends Specification {
...
+    }
+
+    /**
+     * Verify the presence of common tags
+     *
+     * @param tags tags to be verified
+     * @return true if common tags exist
+     */
+    private Boolean verifyCommonTags(final Map<String, String> tags) {
+        return tags.containsKey(MetricsConstants.TagKeys.COMMAND_ID) \
```
mprimi: Since you are using the same values for IDs, you should check the values in this map are what you expect, not just that the keys are present.

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobCompletionService.java
sumitnetflix: Java complains if I declare as final that the variable might not be initialized

genie-web/src/test/groovy/com/netflix/genie/web/tasks/job/JobCompletionServiceSpec.groovy
sumitnetflix: This line was removed in order to fix the test

sumitnetflix:
@mprimi All your code review suggestions have been added. @tgianos fyi

PR #725, opened 2018-04-11
Profile and property cleanup

tgianos:
This PR contains various small changes (and one bigger change in the case of leveraging spring cloud aws in web server) to get rid of technical debt, better support AWS standard configuration and align with internal standards for Spring based applications.


genie-agent/src/main/resources/application.yml
```diff
@@ -5,3 +5,7 @@ cloud:
       instanceProfile: true
     stack:
       auto: false
+
+spring:
```
mprimi: This might produce the same end result in terms of stuff loaded, but is it also the same in terms of speed?
Just curious and wondering if you did this benchmark.

genie-agent/src/main/resources/application.yml
```diff
@@ -3,6 +3,9 @@ cloud:
   aws:
     credentials:
       instanceProfile: true
+    region:
```
mprimi: Same here, does this speed up, slow down, or no difference?

mprimi: Same here, does this speed up, slow down, or no difference?

genie-agent/src/main/resources/application.yml
mprimi: This might produce the same end result in terms of stuff loaded, but is it also the same in terms of speed?
Just curious and wondering if you did this benchmark.

genie-agent/src/main/resources/application.yml
mprimi: This might produce the same end result in terms of stuff loaded, but is it also the same in terms of speed?
Just curious and wondering if you did this benchmark.

genie-docs/src/docs/asciidoc/_releaseNotes.adoc
```diff
@@ -26,6 +26,32 @@ https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Gu
 NOTE: Many Spring properties were changed as part of the Boot 2.0 release. Those are not documented here. Please
 refer to their documentation.
 
+==== Added
+
+|===
+|Property |Description| Default Value
```
mprimi: Does this really belong in the release notes?

genie-docs/src/docs/asciidoc/_releaseNotes.adoc
```diff
@@ -50,6 +50,11 @@ See https://cloud.spring.io/spring-cloud-aws/[Spring Cloud AWS] for more informa
 See https://cloud.spring.io/spring-cloud-aws/[Spring Cloud AWS] for more information.
 |false
 
+|management.endpoints.web.base-path
```
mprimi: Oh, i see it's something you do abtually... (copy properties doc into release notes)

genie-web/src/main/resources/static/scripts/Main.js
```diff
@@ -32,7 +32,7 @@ export default class App extends React.Component {
   }
 
   componentDidMount() {
-    fetch("/actuator/info", null, "GET", "application/json").done(data => {
+    fetch("/admin/info", null, "GET", "application/json").done(data => {
```
mprimi: Commit message does not conform to best practices (partly because it's doing 2 things)

tgianos: How is this two things? It literally says `Update Actuator endpoints to /admin to match internal standard`?

genie-agent/src/main/resources/application.yml
tgianos: I didn't do any benchmarking. I believe Spring will still be scanning the classpath for auto configs and running an if check whether the class in question is a member of the ignore list. Now it'll still find the class but it'll check the property via the existing conditional on the auto config class. So in theory I imagine it is similar. Also this provides the benefit that someone could turn it back on if they want.

PR #729, opened 2018-04-26
Implement job cleanup

mprimi:


genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/RegexRuleSet.java
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public final class RegexRuleSet {
+    private final Response defaultResponse;
+    private final List<Rule> rules;
+
+    private RegexRuleSet(
+        final LinkedList<Rule> rules,
```
tgianos: Any reason you take in a `LinkedList` instead of just a `List` as you store it in an `ArrayList` anyway?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/RegexRuleSet.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * Utility class to match a string against an ordered set of regexes and obtain an accept/reject response.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public final class RegexRuleSet {
```
tgianos: Nice. No existing utilities in Gauva or apache commons did something like this?

mprimi: None that i could find..

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/RegexRuleSet.java
mprimi: This should be `List`, i'll change it.
Started with the idea of having API to add rules with `addFirst` and `addLast` but then switched to builder so no longer necessary for it to be a linkedlist

PR #730, opened 2018-05-01
Groundwork for persisting agent driven jobs

tgianos:
This PR includes various changes to begin supporting running and persisting jobs directly via the Genie 4 agent.

The goal of opening this PR now is to get certain things (particularly the proto file) in shared repository to prevent divergence as work continues to flesh out agent and server implementations.


genie-proto/src/main/proto/genie.proto
```diff
@@ -145,6 +154,56 @@ message JobSpecificationServiceError {
...
+        ALREADY_CLAIMED = 1;
+        NO_SUCH_JOB = 2;
+    }
+    Type type = 1;
+    string message = 2;
+}
+
+message ClaimJobResponse {
+    oneof response {
+        bool successful = 1;
```
mprimi: Not a fan of having either a boolean (which supposedly is always true, if set) or an error message.
The boolean doesn't "add" any information, you could just check if the error is set or not.
Or have the boolean outside the `oneof` and have error set when it is false.

genie-proto/src/main/proto/genie.proto
```diff
@@ -73,8 +76,29 @@ message JobMetadata {
...
+    string id = 2;
+    string name = 3;
+    string status = 4;
+    string version = 5;
+}
+
+message ExecutionResourceCriteria {
+    repeated Criterion cluster_criteria = 1;
+    Criterion command_criterion = 2;
+    repeated string application_ids = 3;
```
mprimi: Should we name this explicitly `_override` for clarity?

genie-proto/src/main/proto/genie.proto
```diff
@@ -145,6 +154,56 @@ message JobSpecificationServiceError {
...
+
+message ClaimJobResponse {
+    oneof response {
+        bool successful = 1;
+        ClaimJobError error = 2;
+    }
+}
+
+message ChangeJobStatusRequest {
+    string current_status = 1;
```
mprimi: This could be a protobuf enum clone of the JobStatus enum, ugly to duplicate, but more type safe and explicit, wdyt?

genie-proto/src/main/proto/genie.proto
```diff
@@ -52,8 +52,11 @@ message AgentRegistrationResponse {
 // Job Service
 service JobService {
     rpc reserveJobId (ReserveJobIdRequest) returns (ReserveJobIdResponse);
-    rpc resolveJobSpecification (ResolveJobSpecificationRequest) returns (JobSpecificationResponse);
-    rpc getJobSpecification (GetJobSpecificationRequest) returns (JobSpecificationResponse);
+    rpc resolveJobSpecification (JobSpecificationRequest) returns (JobSpecificationResponse);
+    rpc getJobSpecification (JobSpecificationRequest) returns (JobSpecificationResponse);
+    rpc resolveJobSpecificationDryRun (DryRunJobSpecificationRequest) returns (JobSpecificationResponse);
```
mprimi: I'm not sure why you need all 3:

```
rpc resolveJobSpecification (JobSpecificationRequest) returns (JobSpecificationResponse);
rpc getJobSpecification (JobSpecificationRequest) returns (JobSpecificationResponse);
rpc resolveJobSpecificationDryRun (DryRunJobSpecificationRequest) returns (JobSpecificationResponse);
```

The first one in particular.
In case of agent job, we send the "Job request" with the ID reservatuin, and the agent later can use `getJobSpecification`.
In case of job via API, this is not even invoked, the agent has an ID and can again just use `getJobSpecification`.

Could we get rid of `resolveJobSpecificationDryRun` and make `resolveJobSpecification` a side-effect-free API that just resolves without persisting?

genie-web/src/test/groovy/com/netflix/genie/web/controllers/DtoConvertersSpec.groovy
```diff
@@ -23,20 +23,7 @@ import com.netflix.genie.common.dto.ApplicationStatus
...
-import com.netflix.genie.common.internal.dto.v4.ClusterRequest
-import com.netflix.genie.common.internal.dto.v4.Command
-import com.netflix.genie.common.internal.dto.v4.CommandMetadata
-import com.netflix.genie.common.internal.dto.v4.CommandRequest
-import com.netflix.genie.common.internal.dto.v4.Criterion
-import com.netflix.genie.common.internal.dto.v4.ExecutionEnvironment
-import com.netflix.genie.common.internal.dto.v4.ExecutionResourceCriteria
-import com.netflix.genie.common.internal.dto.v4.JobMetadata
-import com.netflix.genie.common.internal.dto.v4.JobRequest
+import com.netflix.genie.common.internal.dto.v4.*
```
mprimi: IDEA defaults to `*` if you have more than 10.
Ok to let it do this, or shall we always disallow wildcard imports?
(we can even edit gradle configuration for idea so it generates a project with this value set to `never`)

genie-proto/src/main/proto/genie.proto
tgianos: Yes I agree. I couldn't think of a good way to do this. I had left it as `oneof` for consistency between APIs but I'm find with getting rid of it for some other mechanism. When you say "check if the error is set or not" do you mean the native gRPC error mechanism or just check if the error field is not null?

genie-proto/src/main/proto/genie.proto
tgianos: Don't really care one way or another internally for this API. It was set to this for consistency with V3. I think in the database they're called `applications_requested` cause it doesn't necessarily guarantee that they're used or even exist.

genie-proto/src/main/proto/genie.proto
tgianos: I avoided doing this for the feeling of copying and pasting over and over. We can do it for type safety but I was thinking since it's internal API we can still do something like `JobStatus.RUNNING.toString()` on one side and `JobStatus.parse(string)` on the other side to ensure the types are consistent. This is just a wire protocol.

genie-web/src/test/groovy/com/netflix/genie/web/controllers/DtoConvertersSpec.groovy
tgianos: Yeah I don't like the wildcard imports. I actually disabled it for java but I guess I forgot for Groovy. I've changed it in my IDE now. We can change it for the IDEA gradle plugin if you want I'm fine with that I just don't use it.

genie-proto/src/main/proto/genie.proto
tgianos: There are three because of the three existing cases (in API order):
1. Agent launches and reserves job id. Record exists in database but job specification hasn't been calculated yet. Agent calls this API to kick off the job specification resolution based on the values saved in the database. Specification details saved in database and returned.
2. Agent launched after API call. In this case server already did job specification resolution before agent was lost so we're just calling `get` which returns the values that were saved in the database previously and return them without resolving again
3. Dry run which requires sending a bunch of information that wouldn't be in the database cause we never reserved id or ran through the API so we have to send it all and only run the resolution side effect free

I thought we'd agreed on this but if we're still not clear we can discuss further

genie-proto/src/main/proto/genie.proto
mprimi: The second, check if error is null.
Optionally keep the boolean `success` and the contract is that if that's set to false, then error is not going to be null.

genie-proto/src/main/proto/genie.proto
mprimi: I was just suggesting in terms of readability. Someone that sees this would have to trace through a lot of code to figure out what it does, `_overrides` could provide a more intuitive hint as to its purpose.

genie-proto/src/main/proto/genie.proto
mprimi: I guess we didn't agree on a small detail: in the first case (agent claiming an ID) you were thinking the "job request" is just parked in the DB until we explicitly call resolve.
I was thinking we could resolve it then and there and avoid "burning" the ID if the request is not resolving. If it does resolve, we store request and spec, and the agent just need to do a `get` later, thus making the first of these three unnecessary.
You were instead thinking of just storing the spec and don't attempt to resolve until explicitly asked.
Either works... 
I just find easier to keep in mind that the agent always just need to do a `get`.
And `resolve` is just an API to resolve without side-effects.

genie-proto/src/main/proto/genie.proto
tgianos: I like the second option there. I'll go with that

tgianos:
@mprimi incorporated changes and added a commit at the end for some db changes

genie-web/src/main/resources/db/migration/h2/V4_0_0__Genie_4.sql
```diff
@@ -65,3 +65,36 @@ ALTER TABLE `criteria`
   ADD COLUMN `version` VARCHAR(255) DEFAULT NULL;
 ALTER TABLE `criteria`
   ADD COLUMN `status` VARCHAR(255) DEFAULT NULL;
+
+CREATE TABLE `job_requested_environment_variables` (
```
mprimi: Seems we don't preserve orders of these. I suppose that's ok as long as we make it clear that setting
```
A=foo
...
A=bar
```
does not guarantee that the value of `A` will be `bar`

tgianos: Yes. Currently in memory these are stored in a Map so no order is applied. This is (currently) meant to only be the `variables passed in by user explicitly` + `those set by server (GENIE_*)`. I currently was not planning on storing any variables set by the setup scripts at runtime. If we decide that is the case I still suspect we'll only store the final variable state (e.g. what would be printed by an `env` command) so it should be one key to one value.

PR #732, opened 2018-05-08
Persist and retrieve V4 Job Requests

tgianos:
This PR adds the capability to persist V4 job requests (only used by the Agent via CLI currently) in the database. This allows the reservation of job ids to be done for Agent CLI jobs.

Retrieval of the job request information is also implemented so that subsequent requests can leverage this information.

Subsequent to this job specification information can be saved in the database after resolution has been completed.


genie-web/src/main/resources/db/migration/mysql/V4_0_0__Genie_4.sql
```diff
@@ -137,7 +137,23 @@ CREATE TABLE `job_environment_variables` (
   ROW_FORMAT = DYNAMIC;
 
 ALTER TABLE `jobs`
-  ADD COLUMN `interactive`                      BOOLEAN       DEFAULT FALSE NOT NULL,
-  ADD COLUMN `requested_job_directory_location` VARCHAR(1024) DEFAULT NULL,
-  ADD COLUMN `requested_agent_config_ext`       TEXT          DEFAULT NULL,
-  ADD COLUMN `requested_agent_environment_ext`  TEXT          DEFAULT NULL;
+  ADD COLUMN    `interactive`                               BOOLEAN       DEFAULT FALSE NOT NULL,
```
mprimi: I'm expecting to add a bunch of these when the agent is back in usable form and people start using it.
As this is now, it's going to require a lot of changes to the DB, add a column each time, migrate, update DTOs, builders, etc.
Since these are for the moment not part of 'legacy' could we short-circuit with something easier to update?
Maybe a single column "agent settings" where for the moment we store a JSON payload

tgianos: I added multiple columns that are `ext` columns for agent config and agent environment.  Currently they're for the requests but we will need concrete `ext` columns for the final result of these things after possible server manipulation. I don't particularly love us using these columns ourselves as a blob store long term as they tightly couple us to these fields/payload structure and we can't run queries on them easily. If you're saying you want to use them short term for storage while we're still in active development they're probably a fine place for that.

mprimi: Yeah it's just that we may end up with a lot of them (for example even a simple component like artifacts cache may have half a dozen of them: location options, cleanup options, download options, etc. And same for other components.
While it's nice to capture all of this on the server, adding or changing each one of these is a lot of work right now, as it involves database migration and a whole bunch more.
I was just wondering if we can short-circuit this for now, until they're more stable.
The blob was an idea, but actually we may even want to skip the database entirely for the moment.

PR #733, opened 2018-05-09
Exponential backoff for for job completion check

mprimi:


genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitor.java
```diff
@@ -205,14 +218,22 @@ public void run() {
      */
     @Override
     public GenieTaskScheduleType getScheduleType() {
-        return GenieTaskScheduleType.FIXED_DELAY;
+        return GenieTaskScheduleType.ADAPTIVE_BACKOFF;
```
sumitnetflix: Should this be parameterized with something like fast property. In case there is an issue we can quickly fall back to FIXED_DELAY

genie-common/src/main/java/com/netflix/genie/common/util/AdaptiveBackOff.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.common.util;
```
tgianos: For Genie 3 I think this should be in `genie-core`. It's not something that would or should be exposed to through the client

genie-common/src/main/java/com/netflix/genie/common/util/AdaptiveBackOff.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @author mprimi
+ * @since 3.3.9
+ */
+public interface AdaptiveBackOff {
+
+    boolean isTimeToExecute();
```
tgianos: Javadocs needed (not sure if this part of your WIP but thought I'd mention)

genie-common/src/main/java/com/netflix/genie/common/util/AdaptiveBackOffAdaptor.java
```diff
@@ -0,0 +1,55 @@
...
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.common.util;
+
+/**
+ * Wrapper for Runnable that artificially delays the task execution.
```
tgianos: Here you call it a "wrapper" but you named the class `Adapter`. To me it seems like a wrapper as it's not really adapting anything

genie-web/src/main/java/com/netflix/genie/web/tasks/GenieTask.java
```diff
@@ -61,4 +62,20 @@ public long getFixedRate() {
     public long getFixedDelay() {
         throw new UnsupportedOperationException("This task doesn't support being scheduled at a fixed delay.");
     }
+
+    /**
+     * Get an AdaptiveBackOff which dynamically adjusts delay scheduling a tasks.
+     * @return An Adaptive Backoff object
+     */
+    public AdaptiveBackOff getGetAdaptiveBackoff() {
```
sumitnetflix: The method name looks odd - has 2 gets .

genie-web/src/main/java/com/netflix/genie/web/tasks/GenieTask.java
```diff
@@ -61,4 +62,20 @@ public long getFixedRate() {
...
+        throw new UnsupportedOperationException("This task doesn't support being scheduled with adaptive back-off.");
+    }
+
+    /**
+     * Get how often the scheduler should check for the task to be ready for execution.
+     * @return an interval in milliseconds
+     */
+    public long getGetAdaptiveBackoffCheckInterval() {
```
sumitnetflix: The method name looks odd - has 2 gets .

genie-common/src/main/java/com/netflix/genie/common/util/ExponentialBackOff.java
```diff
@@ -0,0 +1,101 @@
...
+package com.netflix.genie.common.util;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Implementation of AdaptiveBackOff that increases the delay exponentially and optionally resets in case of
+ * success and/or failure.
+ *
+ * @author mprimi
+ * @since 3.3.8
```
tgianos: 3.3.9 is the next version

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitor.java
```diff
@@ -205,14 +218,22 @@ public void run() {
...
+        return adaptiveBackOff;
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
-    public long getFixedDelay() {
-        return this.execution.getCheckDelay().orElseThrow(IllegalArgumentException::new);
+    public long getGetAdaptiveBackoffCheckInterval() {
```
sumitnetflix: The method name looks odd - has 2 gets .

genie-common/src/main/java/com/netflix/genie/common/util/ExponentialBackOff.java
```diff
@@ -0,0 +1,101 @@
...
+        this.factor = factor;
+        this.resetOnSuccess = resetOnSuccess;
+        this.resetOnFailure = resetOnFailure;
+        this.currentInterval = minInterval;
+
+        if (maxInterval < minInterval) {
+            throw new IllegalArgumentException("Maximum interval cannot be smaller than minimum");
+        }
+
+        if (factor < 1) {
```
tgianos: This case (if you want) could be covered with `@Min` `javax.validation`. However then you'd have to validate this implementation not sure if you want to do that or not just pointing it out.

genie-core/src/main/java/com/netflix/genie/core/properties/ExponentialBackOffProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ * Properties to configure an ExponentialBackOff.
+ *
+ * @author mprimi
+ * @since 3.3.8
+ */
+
+@Getter
+@Setter
+@Validated
+public class ExponentialBackOffProperties {
```
tgianos: You're planning this to be an `@ConfigurationProperties`?

genie-core/src/main/java/com/netflix/genie/core/properties/JobsProperties.java
```diff
@@ -50,4 +50,7 @@
 
     @NotNull
     private JobsUsersProperties users = new JobsUsersProperties();
+
+    @NotNull
+    private ExponentialBackOffProperties monitorExponentialBackoff = new ExponentialBackOffProperties();
```
tgianos: You're going to need to add these properties to the `properties` asciidoc

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitor.java
tgianos: Currently we don't support fast properties so it would require a reboot anyway. We could make it a property what the "default" is

genie-web/src/main/java/com/netflix/genie/web/tasks/job/JobMonitor.java
mprimi: ugh, thanks

genie-common/src/main/java/com/netflix/genie/common/util/AdaptiveBackOffAdaptor.java
mprimi: At least GoF canon, Adapter and Wrapper are the same thing https://en.wikipedia.org/wiki/Adapter_pattern

genie-core/src/main/java/com/netflix/genie/core/properties/ExponentialBackOffProperties.java
mprimi: It's not standalone, it's a sub-tree of JobsProperties, as such I didn't think that's necessary (it was not on the other properties classes this sits next to)

genie-common/src/main/java/com/netflix/genie/common/util/AdaptiveBackOffAdaptor.java
mprimi: (and it's adapting a Runnable scheduled with a random backoff to look like one that is scheduled at fixed rate)

genie-core/src/main/java/com/netflix/genie/core/util/ExponentialBackOffTrigger.java
```diff
@@ -0,0 +1,141 @@
...
+        final Long maxDelay,
+        final float factor
+    ) {
+        this.delayType = delayType;
+        this.minDelay = minDelay;
+        this.maxDelay = maxDelay;
+        this.factor = factor;
+
+        if (minDelay < 1) {
+            throw new IllegalArgumentException("Minimum delay must be a positive number");
```
tgianos: I think if you have these kind of preconditions you should capture them in the javadoc for the constructor otherwise you're just throwing unexpected exceptions.

genie-core/src/main/java/com/netflix/genie/core/util/ExponentialBackOffTrigger.java
```diff
@@ -0,0 +1,141 @@
...
+     * Constructor.
+     * @param delayType type of delay
+     * @param minDelay minimum delay in milliseconds
+     * @param maxDelay maximum delay in milliseconds
+     * @param factor multiplier factor to grow the delay
+     */
+    public ExponentialBackOffTrigger(
+        final DelayType delayType,
+        final long minDelay,
+        final Long maxDelay,
```
tgianos: Why is your maxDelay a boxed long but your minDelay is a primitive? Looks like you don't check for null anyway so may as well unbox the maxDelay no?

genie-core/src/main/java/com/netflix/genie/core/util/ExponentialBackOffTrigger.java
```diff
@@ -0,0 +1,141 @@
...
+        final float factor
+    ) {
+        this.delayType = delayType;
+        this.minDelay = minDelay;
+        this.maxDelay = maxDelay;
+        this.factor = factor;
+
+        if (minDelay < 1) {
```
tgianos: This `1` and the below `0.0f` are "magic numbers" and likely should be replaced with constants which clearly define their meaning.

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.scheduling.TriggerContext
+import spock.lang.Specification
+import spock.lang.Unroll
+
+@Category(UnitTest)
+class ExponentialBackOffTriggerSpec extends Specification {
+
+    @Unroll
+    def "NextExecutionTime with type #delayType"(
+            ExponentialBackOffTrigger.DelayType delayType,
```
tgianos: FYI you don't even have to define these input parameters with spock if you don't want you could leave the function as `()` and the variables still get assigned based on the names of the table columns in the `where` clause.

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
```diff
@@ -0,0 +1,150 @@
...
+        long min = 1L
+        long max = 16L
+        float factor = 2.0f
+        TriggerContext triggerContext = Mock(TriggerContext) {
+            lastScheduledExecutionTime() >> new Date(100)
+            lastActualExecutionTime() >> new Date(200)
+            lastCompletionTime() >> new Date(300)
+        }
+
+        setup:
```
tgianos: For my own edification, what is the difference between the above block of code which sets up some variables and this block which explicitly is labeled as setup?

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
```diff
@@ -0,0 +1,150 @@
...
+            Date nextExecutionTime = trigger.nextExecutionTime(triggerContext)
+
+            then:
+            Date expectedExecutionTime = new Date(nextExpectedExecutions[i])
+            nextExecutionTime == expectedExecutionTime
+        }
+
+        trigger.reset()
+
+        for (int i = 0; i < nextExpectedExecutions.length; i++) {
```
tgianos: Wouldn't passing the array in as a List and using a foreach loop be simpler? e.g. `for (Long time : nextExpectedExecutions) { blah blah blah }`

genie-core/src/test/java/com/netflix/genie/core/properties/ExponentialBackOffTriggerPropertiesUnitTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Unit test for ExponentialBackOffTriggerProperties.
+ *
+ * @author mprimi
+ * @since 3.3.9
+ */
+public class ExponentialBackOffTriggerPropertiesUnitTest {
```
tgianos: Any reason you did this with raw `junit` instead of `spock`?

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
mprimi: Oh good to know..

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
mprimi: Setup should include those lines..

genie-core/src/test/java/com/netflix/genie/core/properties/ExponentialBackOffTriggerPropertiesUnitTest.java
mprimi: Just cloned an existing test for a sister class, figured consistency with existing would be better than switching to spock in this case.

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
mprimi: Note: explicit args parameter give more control over type inference.
e.g. removing `long delay` and then having a table that contains `100` was passing it as `integer` and producing warnings. (easily fixed by having `100L` in the table, but earlier this conversion was done for me)

mprimi:
@tgianos addressed all your comments

genie-core/src/test/groovy/com/netflix/genie/core/util/ExponentialBackOffTriggerSpec.groovy
```diff
@@ -0,0 +1,147 @@
...
+            Date nextExecutionTime = trigger.nextExecutionTime(triggerContext)
+
+            then:
+            Date expectedExecutionTime = new Date(expectedDelay)
+            nextExecutionTime == expectedExecutionTime
+        }
+
+        where:
+        delayType                                                              | nextExpectedExecutions
+        ExponentialBackOffTrigger.DelayType.FROM_PREVIOUS_SCHEDULING           | [101L, 102, 104, 108, 116, 116, 116]
```
tgianos: Since you left the typing in did you mean to leave this first element as a long type coercion?

PR #736, opened 2018-05-18
Upgrade Spring and other libraries

mprimi:


build.gradle
```diff
@@ -98,6 +98,18 @@ configure(javaProjects) {
             dependency("org.opensaml:opensaml:2.6.4")
             dependency("org.python:jython-standalone:2.7.1")
             dependency("org.springframework.security.extensions:spring-security-saml2-core:1.0.3.RELEASE")
+            dependencySet(group: "com.fasterxml.jackson.core", version: "2.9.5") {
```
tgianos: We can discuss why you're upgrading Jackson from 2.8.x to 2.9.x in person but the easier way to do this is to set `jackson.version` in `gradle.properties` similar to how you did the `h2.version`

See the boot bom the property comes from: https://github.com/spring-projects/spring-boot/blob/v1.5.13.RELEASE/spring-boot-dependencies/pom.xml#L99

genie-common/src/test/java/com/netflix/genie/common/util/GenieDateFormatUnitTests.java
```diff
@@ -46,6 +46,6 @@ public void canFormat() {
         final StringBuffer buffer = new StringBuffer();
         final FieldPosition pos = Mockito.mock(FieldPosition.class);
 
-        Assert.assertThat(format.format(zero, buffer, pos).toString(), Matchers.is("1970-01-01T00:00:00.000Z"));
+        Assert.assertThat(format.format(zero, buffer, pos).toString(), Matchers.is("1970-01-01T00:00:00.000+0000"));
```
tgianos: I'm pretty sure this is a breaking change. Certain libraries like Kragle and I believe our UI expect the dates in the previous format with the `Z` at the end.

mprimi:
@tgianos Updated, using gradle property to pin version and reverted to previous date format.

genie-common/src/main/java/com/netflix/genie/common/util/GenieDateFormat.java
```diff
@@ -17,37 +17,52 @@
...
 
     /**
      * {@inheritDoc}
      */
     @Override
     public StringBuffer format(final Date date, final StringBuffer toAppendTo, final FieldPosition fieldPosition) {
-        final String value = ISO8601Utils.format(date, true);
-        toAppendTo.append(value);
-        return toAppendTo;
+        // StdDateFormat and ISO8601Utils do not offer a way to format consistent with the current Genie API format
```
tgianos: I think it would be much simpler to convert the `Date` to an `Instant` and then use [Instant#toString()](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#toString--)

mprimi: Instant format does not include milliseconds :-(

tgianos: Yeah I noticed that. Could still try to take advantage of [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html) if you want instead of all the Calendar stuff. Up to you.

mprimi: @tgianos tried `DateTimeFormatter` but it's not simpler than the `format` below, comes out similar to this in terms of lines of code, but it's more complex to understand, so I'd leave this as-is. 
Not great that we have to manually provide a format string, but at least it's straighforward what's going on.

PR #737, opened 2018-05-21
Add V4 Job Flag

tgianos:
This PR adds a V4 boolean to the database jobs table.

This field is intended to be temporary during V4 development cycle to allow running of Jobs via old V3 job API path (on node) and via V4 Agent (cli) and be able to filter jobs at different points in this system by this distinction.

Longer term (when V4 API is ready) this field should be dropped from the database as job running will all be done via Agent one way or another.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
@@ -121,31 +121,33 @@ public class ClusterCheckerTask extends LeadershipTask {
             .filter(host -> !this.hostName.equals(host))
             .forEach(this::validateHostAndUpdateErrorCount);
 
-        this.errorCounts.entrySet().removeIf(entry -> {
-            final String host = entry.getKey();
-            boolean result = true;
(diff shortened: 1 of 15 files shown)


genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ClusterCheckerTask.java
```diff
@@ -121,42 +121,44 @@ public void run() {
...
             }
-            return result;
-        });
+        );
         log.info("Finished checking for cluster node health.");
     }
 
     private void updateJobsToFailedOnHost(final String host) {
-        final Set<Job> jobs = jobSearchService.getAllActiveJobsOnHost(host);
+        final Set<Job> jobs = this.jobSearchService.getAllActiveJobsOnHost(host);
```
sumitnetflix: this.jobSearchService replaces jobSearchService in a lot of places. Is this needed/intentional?

tgianos: Needed? No.

Intentional? Yes. For consistency across the codebase and to make it clear what is a local variable vs. a member variable reference member variables/functions are referenced using `this.`

PR #738, opened 2018-05-22
Exponential backoff for for job completion check (#733 cherry pick)

mprimi:
This is a cherry-pick of https://github.com/Netflix/genie/pull/733

### DIFF
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -143,6 +143,18 @@ to work.
 |The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
 |100
 
+|genie.jobs.completion-check-back-off.min-interval
+|The minimum time between checks for job completion in milliseconds. Must be greater than zero.
+|100
+
+|genie.jobs.completion-check-back-off.max-interval
+|The maximum time between checks for job completion in milliseconds. This is a fallback value, the value used in most cases is specified as part of the `Command` entity for a particular job.
+|10000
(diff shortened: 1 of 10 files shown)


genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -143,6 +143,18 @@ to work.
 |The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
 |100
 
+|genie.jobs.completionCheckBackOff.minInterval
```
tgianos: I think you need to change these for Genie 4 as it builds on Boot 2 which changed how property binding works to enforce certain naming schemes. See [here](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#relaxed-binding). The gist is that non-leaf nodes of the name/path can't be camel case they need to use `-` or something.

tgianos: Their documentation is a little confusing (in so much as they say you need to use kebab case in the `prefix` of `@ConfigurationProperties` but say that camel case will work other places. I had some trouble during the Boot 2 migration with some of our existing properties and I had to switch them. Maybe they've changed something since then but better safe than sorry.

mprimi: From table 24.1 here: https://docs.spring.io/spring-boot/docs/2.0.0.RELEASE/reference/htmlsingle/#boot-features-external-config-relaxed-binding it seems like the current should still work, just not recommended.
We do have *a lot* of properties using camel.
I'll switch these to kebab.

mprimi:
@tgianos updated and rebased

PR #747, opened 2018-05-25
Fix date format returned by REST API

mprimi:


genie-common/src/main/java/com/netflix/genie/common/util/GenieDateFormat.java
```diff
@@ -34,6 +41,11 @@
  */
 public class GenieDateFormat extends StdDateFormat {
 
+    /**
+     * Pattern to validate (e.g., in tests) the expected date format this class produces.
+     */
+    @VisibleForTesting
+    public static final Pattern VALID_PATTERN = Pattern.compile("^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.\\d{3}Z$");
```
tgianos: does it need to be public or could you just make it package protected?

genie-common/src/test/java/com/netflix/genie/common/util/GenieDateFormatUnitTests.java
```diff
@@ -36,16 +42,81 @@
 @Category(UnitTest.class)
 public class GenieDateFormatUnitTests {
 
+    private static final String EXPECTED_DATE_STRING = "1970-01-01T00:00:00.000Z";
+    private static final Date DATE = new Date(0);
+    // Limit random dates used in test between 'epoch' and 'epoch + 1000 years'
+    private static final long MAX_RANDOM_DATE = TimeUnit.MILLISECONDS.convert(365 * 1000, TimeUnit.DAYS);
```
tgianos: You might want to add a supplier to `RandomSuppliers` i have for tests.

genie-web/src/test/java/com/netflix/genie/web/controllers/RestControllerIntegrationTestsBase.java
```diff
@@ -115,6 +117,21 @@
         .setDateFormat(new GenieDateFormat())
         .registerModule(new Jdk8Module());
 
+    protected final Matcher<String> dateFormatMatcher = new BaseMatcher<String>() {
```
tgianos: This part should probably be ported into `master` after this is merged

genie-common/src/main/java/com/netflix/genie/common/util/GenieDateFormat.java
```diff
@@ -65,4 +77,27 @@ public StringBuffer format(final Date date, final StringBuffer toAppendTo, final
             )
         );
     }
+
+    /**
+     * JsonSerializer to output (Optional) Date using GenieDateFormat.
+     *
+     * @since 3.3.11
+     */
+    public static class OptionalDateJsonSerializer extends JsonSerializer<Optional<Date>> {
```
tgianos: I don't see you annotating any `Optional` dates as you have the annotation on members themselves. Does jackson call the getters instead and that's why you have it like this?

genie-common/src/main/java/com/netflix/genie/common/util/GenieDateFormat.java
mprimi: It's used from the `web` module/package.

genie-common/src/main/java/com/netflix/genie/common/util/GenieDateFormat.java
mprimi: Yes

genie-web/src/test/java/com/netflix/genie/web/controllers/RestControllerIntegrationTestsBase.java
mprimi: Yes, i left the controller tests into a separate commit so we can cherry-pick

PR #750, opened 2018-06-07
Initial server side implementation of the Job File Sync Service

tgianos:
An initial implementation of the protocols and services needed to sync files from an agent job directory to a Genie server

Things are still needed for long term completeness of this feature of Genie 4:
- A way to block access to log resources until a server has caught up to expected Agent high water mark
- A way to uniquely identify a grpc connection between an agent and a server in order to de-register the appropriate observer from internal state in case the same agent reattached to the same server before a cleanup could happen
- A way to either push logs from the server to long term storage (e.g. S3) while the logs are "live" or all in one push after upload to the server is complete
- The agent side implementation of this sync protocol
- A mechanism to indicate which node a job logs are currently located on (database or other)

Currently implemented is the low level service which can write bytes to random locations in files for a given job on local disk using NIO. Also the retrieval of logs via V3 API has been modified to leverage this service in order to centralize logic for log interaction.

### DIFF
--- a/genie-proto/src/main/proto/genie.proto
+++ b/genie-proto/src/main/proto/genie.proto
@@ -116,7 +116,7 @@ message ExecutionResource {
     string id = 1;
     string setup_file = 2;
     repeated string configs = 3;
-    repeated string dependences = 4;
+    repeated string dependencies = 4;
 }
 
 message JobSpecification {
--- a/genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java
+++ b/genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java
@@ -255,7 +255,7 @@ public class JobServiceProtoConverter {
             protoResource.getId(),
             new ExecutionEnvironment(
                 ImmutableSet.copyOf(protoResource.getConfigsList()),
-                ImmutableSet.copyOf(protoResource.getDependencesList()),
+                ImmutableSet.copyOf(protoResource.getDependenciesList()),
                 protoResource.getSetupFile().isEmpty() ? null : protoResource.getSetupFile()
             )
         );
@@ -268,7 +268,7 @@ public class JobServiceProtoConverter {
             .newBuilder()
             .setId(executionResource.getId())
             .addAllConfigs(executionResource.getExecutionEnvironment().getConfigs())
-            .addAllDependences(executionResource.getExecutionEnvironment().getDependencies());
+            .addAllDependencies(executionResource.getExecutionEnvironment().getDependencies());
 
         executionResource.getExecutionEnvironment().getSetupFile().ifPresent(builder::setSetupFile);
 
--- a/genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverterSpec.groovy
+++ b/genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverterSpec.groovy
@@ -21,7 +21,6 @@ import com.google.common.collect.ImmutableMap
 import com.google.common.collect.Lists
 import com.google.common.collect.Sets
 import com.netflix.genie.common.dto.JobStatus
-import com.netflix.genie.common.exceptions.GenieNotFoundException
 import com.netflix.genie.common.internal.dto.v4.AgentClientMetadata
 import com.netflix.genie.common.internal.dto.v4.AgentConfigRequest
 import com.netflix.genie.common.internal.dto.v4.AgentJobRequest
@@ -30,24 +29,16 @@ import com.netflix.genie.common.internal.dto.v4.ExecutionEnvironment
 import com.netflix.genie.common.internal.dto.v4.ExecutionResourceCriteria
 import com.netflix.genie.common.internal.dto.v4.JobMetadata
 import com.netflix.genie.common.internal.dto.v4.JobSpecification
-import com.netflix.genie.common.internal.exceptions.unchecked.GenieApplicationNotFoundException
-import com.netflix.genie.common.internal.exceptions.unchecked.GenieClusterNotFoundException
-import com.netflix.genie.common.internal.exceptions.unchecked.GenieCommandNotFoundException
-import com.netflix.genie.common.internal.exceptions.unchecked.GenieJobNotFoundException
-import com.netflix.genie.common.internal.exceptions.unchecked.GenieJobSpecificationNotFoundException
 import com.netflix.genie.common.util.GenieObjectMapper
 import com.netflix.genie.proto.AgentConfig
 import com.netflix.genie.proto.AgentMetadata
 import com.netflix.genie.proto.DryRunJobSpecificationRequest
 import com.netflix.genie.proto.ExecutionResource
-import com.netflix.genie.proto.JobSpecificationError
 import com.netflix.genie.proto.JobSpecificationResponse
 import com.netflix.genie.proto.ReserveJobIdRequest
 import com.netflix.genie.test.categories.UnitTest
 import org.junit.experimental.categories.Category
 import spock.lang.Specification
-import spock.lang.Unroll
-
 /**
  * Specifications for the {@link JobServiceProtoConverter} utility class.
  *
@@ -213,7 +204,7 @@ class JobServiceProtoConverterSpec extends Specification {
     def agentVersion = UUID.randomUUID().toString()
     def agentPid = 21_031
 
-    def converter = new JobServiceProtoConverter();
+    def converter = new JobServiceProtoConverter()
 
     def "Can convert JobRequest to ReserveJobIdRequest and vice versa"() {
         def jobRequest = createJobRequest(id)
@@ -329,7 +320,7 @@ class JobServiceProtoConverterSpec extends Specification {
 
         when:
         def request = converter.toProtoClaimJobRequest(id, agentClientMetadata)
-        def agentClientMetadata2 = converter.toAgentClientMetadataDTO(request.getAgentMetadata());
+        def agentClientMetadata2 = converter.toAgentClientMetadataDTO(request.getAgentMetadata())
 
         then:
         request.getId() == id
@@ -493,7 +484,7 @@ class JobServiceProtoConverterSpec extends Specification {
                 .newBuilder()
                 .setId(id)
                 .addAllConfigs(configs)
-                .addAllDependences(dependencies)
+                .addAllDependencies(dependencies)
                 .setSetupFile(setupFile)
                 .build()
 
@@ -501,7 +492,7 @@ class JobServiceProtoConverterSpec extends Specification {
                 .newBuilder()
                 .setId(clusterId)
                 .addAllConfigs(clusterConfigs)
-                .addAllDependences(clusterDependencies)
+                .addAllDependencies(clusterDependencies)
                 .setSetupFile(clusterSetupFile)
                 .build()
 
@@ -509,7 +500,7 @@ class JobServiceProtoConverterSpec extends Specification {
                 .newBuilder()
                 .setId(commandId)
                 .addAllConfigs(commandConfigs)
-                .addAllDependences(commandDependencies)
+                .addAllDependencies(commandDependencies)
                 .setSetupFile(commandSetupFile)
                 .build()
 
@@ -517,7 +508,7 @@ class JobServiceProtoConverterSpec extends Specification {
                 .newBuilder()
                 .setId(application0Id)
                 .addAllConfigs(application0Configs)
-                .addAllDependences(application0Dependencies)
+                .addAllDependencies(application0Dependencies)
                 .setSetupFile(application0SetupFile)
                 .build()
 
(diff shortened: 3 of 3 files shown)


genie-web/src/main/java/com/netflix/genie/web/services/impl/DiskLogServiceImpl.java
```diff
@@ -0,0 +1,336 @@
...
+
+        if (Files.notExists(logFile)) {
+            // Make sure all the directories exist on disk
+            final Path logFileParent = logFile.getParent();
+            if (logFileParent != null) {
+                this.createOrCheckDirectory(logFileParent);
+            }
+        } else if (Files.isDirectory(logFile)) {
+            // TODO: Perhaps this should be different exception
+            throw new IllegalArgumentException(relativePath + " is a directory not a file. Unable to update");
```
sumitnetflix: Should be same exception as the one we decide to throw at line 256

mprimi: This is a user input error.
Line 256 is the unlikely event that someone is creating files inside a private folder that genie is managing.
I disagree they need to be treated the same (even if the underlying error -- file where i'm attempting to create a directory -- is the same).

mprimi: Whoops, missed placed where the other was invoked, it can be more than user error.

Still think on a very abstract level that it can be appropriate for the same underlying error/state to throw different kinds of error depending on context (and example here is user input/configuration vs. runtime unexpected state).

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/logs/JobLogFile.java
```diff
@@ -0,0 +1,65 @@
+/*
```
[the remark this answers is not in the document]
tgianos: Perhaps semantic but seems reasonable enough. If we do this the whole service should probably be named `JobFileService` instead of `LogService` for consistency I would think

genie-web/src/main/java/com/netflix/genie/web/services/impl/DiskLogServiceImpl.java
```diff
@@ -0,0 +1,336 @@
...
+
+    private final Path jobsDirRoot;
+
+    /**
+     * Constructor.
+     *
+     * @param jobsDir The job directory resource
+     * @throws IOException When the job directory can't be created or isn't a directory
+     */
+    public DiskLogServiceImpl(final Resource jobsDir) throws IOException {
```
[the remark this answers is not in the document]
tgianos: I'm not sure I understand the difference you're trying to establish between the two. Instead of `Log` think `output` or `working directory` if you'd like. In V3 these are written directly and served directly off local disk. If we put the logs that are uploaded from somewhere else in a different location than this existing job directory space it complicates the logic to return log files in response to API requests via the V3 API.

I could see an argument for a location where running job files are kept somewhere and finished jobs are placed somewhere else but i'm not sure what the value in the distinction is here.

genie-web/src/main/java/com/netflix/genie/web/services/impl/DiskLogServiceImpl.java
```diff
@@ -0,0 +1,336 @@
...
+            ? this.jobsDirRoot.resolve(jobId)
+            : this.jobsDirRoot.resolve(jobId).resolve(relativePath);
+        return new PathResource(logFile);
+    }
+
+    private void createOrCheckDirectory(final Path dir) throws IOException {
+        if (Files.notExists(dir)) {
+            Files.createDirectories(dir);
+        } else if (!Files.isDirectory(dir)) {
+            // TODO: Maybe a more specific exception here? Should be runtime maybe as there's no recovery?
```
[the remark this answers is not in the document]
tgianos: I can't see anything using this service other than other pieces of the Genie codebase. I don't expect any users to interact or upload files directly.

genie-web/src/main/java/com/netflix/genie/web/services/impl/DiskLogServiceImpl.java
tgianos: The argument could be made that since this service exist only to write files from the `Agent` and provide handles to files to the API tier on `GET` requests that the Genie application completely owns (or should own) this directory and its entire tree. As such the structure should never match these conditions and they're exceptional (e.g. someone logs into the server and creates a file on disk)

One weird edge case could be that currently this service only handles writing but if a job at runtime does file renames/deletions/etc things could get dicey. We probably need to think more about this case. Not sure if the watcher implementation on the Agent is set up to catch these events from the FS or not.

tgianos:
@mprimi @sumitnetflix I renamed a bunch of files per earlier feedback

I added an initial implementation of the grpc JobFileSyncService. It's very rough but looking for feedback so I committed it. No tests yet.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
```diff
@@ -0,0 +1,399 @@
...
+//        SimpleLoggingInterceptor.class,
+//    }
+)
+@Slf4j
+public class GRpcJobFileSyncServiceImpl extends JobFileSyncServiceGrpc.JobFileSyncServiceImplBase {
+
+    private final JobFileService jobFileService;
+    private final TaskScheduler taskScheduler;
+
+    private final Map<String, GenieJobFileSyncObserver> jobSyncRequestObservers = Maps.newConcurrentMap();
```
sumitnetflix: Instead of having  `Maps.newConcurrentMap();`, would it be better to inject this as a dependency. It might help in testing. I had a similar map of observers for `JobKillService` and I was thinking of creating a interface and injecting its Impl

tgianos: I don't agree with this. This is an internal construct for keeping state important to this particular implementation. If you provide a map externally you're effectively risking manipulation of that Map by outside sources, not to mention exposing internal implementation and breaking encapsulation. No one else should really care that you're keeping this map it's just for internal housekeeping for this class. If you really need access for it for testing you can provide a `protected` or `visablefortesting` getter to check the contents of the map.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
```diff
@@ -0,0 +1,609 @@
...
+
+                    this.successfulSyncIds.clear();
+                    this.failedSyncIds.clear();
+                }
+            }
+        }
+
+        private void handleBeginSync(final BeginSync beginSync) throws IOException {
+            // Handle the begin message provided we're still waiting for it
+            if (this.waitingForBeginMessage) {
```
mprimi: Could add a WARN if this condition is not true.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
```diff
@@ -0,0 +1,609 @@
...
+
+                this.checkIfShouldSendAck();
+            }
+        }
+
+        private void handleDeleteFile(final DeleteFile deleteFile) {
+            final String messageId = deleteFile.getId();
+            log.debug("Received file delete message {} in job {}", messageId, this.jobId);
+
+            if (this.waitingForBeginMessage) {
```
mprimi: This block appears in 3 methods, could be pulled out for less duplication and to make these methods simpler.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
```diff
@@ -0,0 +1,609 @@
...
+                }
+            }
+
+            if (sendAck) {
+                // Proactively send ACK don't wait for timed thread
+                this.sendSyncAckMessageIfNecessary();
+            }
+        }
+    }
+}
```
mprimi: This protocol is stateful on the server, if it gets any more complex we may consider using spring statemachine to encode it.

genie-web/src/main/java/com/netflix/genie/web/services/JobFileService.java
```diff
@@ -0,0 +1,106 @@
...
+     * @throws IOException On error writing the data to the given file
+     */
+    void updateFile(
+        final String jobId,
+        final String relativePath,
+        final long startByte,
+        final byte[] data
+    ) throws IOException;
+
+//    Optional<String> readLogFile(
```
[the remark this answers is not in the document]
tgianos: Yeah it can be removed. I left it in cause I spent a bunch of time on it but eventually decided to punt for now and return a resource reference and allow callers to do with it as they need. Left the code in in case I wanted to use it again. I'll save it as a gist or something somewhere for my own use if I want to revisit.

genie-proto/src/main/proto/genie.proto
```diff
@@ -213,3 +213,83 @@ message ChangeJobStatusResponse {
...
+}
+
+message BeginAcknowledgement {
+    JobDirectoryState server_directory_state = 1;
+}
+
+message ResetSync {
+}
+
+message UploadAcknowledgement {
```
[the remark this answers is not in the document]
tgianos: Yeah I like this it is cleaner than what I have. I'll make the change.

genie-web/src/main/java/com/netflix/genie/web/properties/JobFileSyncRpcProperties.java
```diff
@@ -0,0 +1,37 @@
...
+ * Properties for configuring the job file sync processes.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+@Validated
+@Getter
+@Setter
+public class JobFileSyncRpcProperties {
+    // TODO: Document properties
```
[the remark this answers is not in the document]
tgianos: Yes I should. I forgot. Will add.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
tgianos: I agree we may want to do that or perhaps should. Outside the scope of this PR but it sounds like it might be the right thing to do.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
```diff
@@ -0,0 +1,609 @@
...
+     * Activate any cleanup necessary when an agent detaches.
+     *
+     * @param jobId The id of the job the agent was responsible for
+     */
+    public void onAgentDetached(final String jobId /* TODO: This will be event type eventually likely */) {
+        // TODO: Possible race condition here if this is invoked after a new response observer has already been added
+        //       for the same job. Just putting this here as temporary logic till we figure that out
+        final JobFileSyncObserver observer = this.jobSyncRequestObservers.get(jobId);
+        if (observer != null) {
+            // Cleanup will invoke the cleanup callback and remove it from the map
```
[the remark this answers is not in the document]
tgianos: Going to hold off on modifying this until we finalize how we want to identify and remove things from map

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
tgianos: I left it in each due to wanting unique log messages that had specific details. Can change it later if we find it a burden. The logic to check whether to send the ack is already shared in one spot they all call.

tgianos:
@mprimi I've made changes based on your feedback and added one more commit at the end for a misspelling in the proto file. If you want to take one last look let me know otherwise I'll merge this tonight/tomorrow after builds

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobFileSyncServiceImpl.java
```diff
@@ -0,0 +1,612 @@
...
+     * environment. This sync complete will clean up everything locally and kick off side effect processes
+     * <p>
+     * Reconnect workflow
+     * <p>
+     * 1. Any message other than {@link BeginSync} is received by this instance first
+     * 2. A {@link ResetSync} message is sent to the client (agent) and all messages until a {@link BeginSync} is
+     * received are ignored
+     * - Only one {@link ResetSync} message is sent no matter how many other messages that aren't {@link BeginSync}
+     * are received before the client sends a {@link BeginSync}
+     * 3. Once the {@link BeginSync} message is received processing continues as per the healthy workflow
```
sumitnetflix: 👍 for the comment. Very helpful

PR #752, opened 2018-06-14
Add job kill service

sumitnetflix:
First draft of the job kill service. My intent was to

1. Have similar interfaces as defined for other services on the agent side
2. Get on the same page for the flow for a job kill. How will registration work? How will a kill from server be routed back to the agent?

No tests have been added yet. I have marked the outstanding questions in my head as TODO items. Please have a look and provide your feedback

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/KillService.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/KillService.java
@@ -1,0 +1,32 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.execution.services;
+
+/**
+ * Service responsible for killing the job.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public interface KillService {
+    /**
+     * Perform all operations associated with killing the job.
+     */
+    void kill();
+}
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/KillServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/KillServiceImpl.java
@@ -1,0 +1,44 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.execution.services.impl;
+
+import com.netflix.genie.agent.execution.services.KillService;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * Implementation of {@link KillService}.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Component
+@Lazy
+@Slf4j
+class KillServiceImpl implements KillService {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void kill() {
+        throw new UnsupportedOperationException();
+    }
+}
--- a/genie-proto/src/main/proto/genie.proto
+++ b/genie-proto/src/main/proto/genie.proto
@@ -304,3 +304,14 @@ message AgentHeartBeat {
 
 message ServerHeartBeat {
 }
+
+service JobKillService {
+    rpc registerForKillNotification (JobKillRegistrationRequest) returns (JobKillRegistrationResponse);
+}
+
+message JobKillRegistrationRequest {
+    string job_id = 1;
+}
+
+message JobKillRegistrationResponse {
+}
--- a/genie-web/src/main/java/com/netflix/genie/web/configs/GenieServicesAutoConfiguration.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/configs/GenieServicesAutoConfiguration.java
@@ -38,6 +38,7 @@ import com.netflix.genie.web.services.FileTransferFactory;
 import com.netflix.genie.web.services.JobCoordinatorService;
 import com.netflix.genie.web.services.JobFileService;
 import com.netflix.genie.web.services.JobKillService;
+import com.netflix.genie.web.services.JobKillServiceV4;
 import com.netflix.genie.web.services.JobPersistenceService;
 import com.netflix.genie.web.services.JobSearchService;
 import com.netflix.genie.web.services.JobSpecificationService;
@@ -51,9 +52,10 @@ import com.netflix.genie.web.services.impl.DiskJobFileServiceImpl;
 import com.netflix.genie.web.services.impl.FileSystemAttachmentService;
 import com.netflix.genie.web.services.impl.GenieFileTransferService;
 import com.netflix.genie.web.services.impl.JobCoordinatorServiceImpl;
+import com.netflix.genie.web.services.impl.JobKillServiceImpl;
 import com.netflix.genie.web.services.impl.JobSpecificationServiceImpl;
 import com.netflix.genie.web.services.impl.LocalFileTransferImpl;
-import com.netflix.genie.web.services.impl.LocalJobKillServiceImpl;
+import com.netflix.genie.web.services.impl.JobKillServiceV3;
 import com.netflix.genie.web.services.impl.LocalJobRunner;
 import com.netflix.genie.web.tasks.job.JobCompletionService;
 import io.micrometer.core.instrument.MeterRegistry;
@@ -101,8 +103,8 @@ public class GenieServicesAutoConfiguration {
      * @return A job kill service instance.
      */
     @Bean
-    @ConditionalOnMissingBean(JobKillService.class)
-    public JobKillService jobKillService(
+    @ConditionalOnMissingBean(JobKillServiceV3.class)
+    public JobKillServiceV3 jobKillServiceV3(
         final GenieHostInfo genieHostInfo,
         final JobSearchService jobSearchService,
         final Executor executor,
@@ -111,7 +113,7 @@ public class GenieServicesAutoConfiguration {
         @Qualifier("jobsDir") final Resource genieWorkingDir,
         final ObjectMapper objectMapper
     ) {
-        return new LocalJobKillServiceImpl(
+        return new JobKillServiceV3(
             genieHostInfo.getHostname(),
             jobSearchService,
             executor,
@@ -119,6 +121,28 @@ public class GenieServicesAutoConfiguration {
             genieEventBus,
             genieWorkingDir,
             objectMapper
+        );
+    }
+
+    /**
+     * Get an local implementation of the JobKillService.
+     *
+     * @param jobKillServiceV3   Service to kill V3 jobs.
+     * @param jobKillServiceV4      Service to kill V4 jobs.
+     * @param jobPersistenceService Job persistence service
+     * @return A job kill service instance.
+     */
+    @Bean
+    @ConditionalOnMissingBean(JobKillService.class)
+    public JobKillService jobKillService(
+        final JobKillServiceV3 jobKillServiceV3,
+        final JobKillServiceV4 jobKillServiceV4,
+        final JobPersistenceService jobPersistenceService
+    ) {
+        return new JobKillServiceImpl(
+            jobKillServiceV3,
+            jobKillServiceV4,
+            jobPersistenceService
         );
     }
 
--- a/genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/IsV4JobProjection.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/IsV4JobProjection.java
@@ -1,0 +1,34 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.web.jpa.entities.projections.v4;
+
+/**
+ * Projection for capturing if the job was launched using Genie v4.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Deprecated
+public interface IsV4JobProjection {
+    /**
+     * Job was launched using Genie v4.
+     *
+     * @return true if job was launched via Genie v4 else false
+     */
+    boolean isV4();
+}
--- a/genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
@@ -54,6 +54,7 @@ import com.netflix.genie.web.jpa.entities.FileEntity;
 import com.netflix.genie.web.jpa.entities.JobEntity;
 import com.netflix.genie.web.jpa.entities.projections.IdProjection;
 import com.netflix.genie.web.jpa.entities.projections.v4.JobSpecificationProjection;
+import com.netflix.genie.web.jpa.entities.projections.v4.IsV4JobProjection;
 import com.netflix.genie.web.jpa.entities.projections.v4.V4JobRequestProjection;
 import com.netflix.genie.web.jpa.entities.v4.EntityDtoConverters;
 import com.netflix.genie.web.jpa.repositories.JpaApplicationRepository;
@@ -461,6 +462,26 @@ public class JpaJobPersistenceServiceImpl extends JpaBaseService implements JobP
         return projection.isResolved()
             ? Optional.of(EntityDtoConverters.toJobSpecificationDto(projection))
             : Optional.empty();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Transactional(readOnly = true)
+    public boolean isV4(
+        @NotBlank(message = "Id is missing and is required") final String id
+    ) {
+        log.debug("Read v4 flag from db for job {} ", id);
+        return this.jobRepository
+            .findByUniqueId(id, IsV4JobProjection.class)
+            .orElseThrow(
+                () -> {
+                    final String errorMessage = "No job with id " + id + "exists. Unable to get v4 flag.";
+                    log.error(errorMessage);
+                    return new GenieJobNotFoundException(errorMessage);
+                }
+            ).isV4();
     }
 
     /**
--- a/genie-web/src/main/java/com/netflix/genie/web/services/JobKillServiceV4.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/services/JobKillServiceV4.java
@@ -23,21 +23,21 @@ import org.springframework.validation.annotation.Validated;
 import javax.validation.constraints.NotBlank;
 
 /**
- * Interface for services to kill jobs.
+ * Interface for services to kill V4 jobs.
  *
- * @author tgianos
- * @since 3.0.0
+ * @author standon
+ * @since 4.0.0
  */
 @Validated
-public interface JobKillService {
+public interface JobKillServiceV4 {
 
     /**
-     * Kill the job with the given id if possible. Should publish a JobFinishedEvent when done.
+     * Kill the job with the given id if possible.
      *
-     * @param id     id of job to kill
+     * @param jobId     id of job to kill
      * @param reason brief reason for requesting the job be killed
      * @throws GenieException if there is an error
      */
-    void killJob(@NotBlank(message = "No id entered. Unable to kill job.") final String id,
+    void killJob(@NotBlank(message = "No job id entered. Unable to kill job.") final String jobId,
                  @NotBlank(message = "No reason provided.") final String reason) throws GenieException;
 }
--- a/genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
+++ b/genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
@@ -1,0 +1,97 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.web.services.impl
+
+import com.netflix.genie.common.exceptions.GenieServerException
+import com.netflix.genie.common.internal.exceptions.unchecked.GenieJobNotFoundException
+import com.netflix.genie.test.categories.UnitTest
+import com.netflix.genie.web.services.JobKillServiceV4
+import com.netflix.genie.web.services.JobPersistenceService
+import org.junit.experimental.categories.Category
+import spock.lang.Specification
+
+/**
+ * Tests for JobKillServiceImpl
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Category(UnitTest.class)
+class JobKillServiceImplSpec extends Specification {
+    JobKillServiceV3 jobKillServiceV3
+    JobKillServiceImpl service
+    JobPersistenceService jobPersistenceService
+    JobKillServiceV4 jobKillServiceV4
+    String jobId
+
+    void setup() {
+        jobPersistenceService = Mock()
+        jobKillServiceV4 = Mock()
+        jobKillServiceV3 = Mock()
+        jobId = UUID.randomUUID().toString()
+        service = new JobKillServiceImpl (jobKillServiceV3, jobKillServiceV4, jobPersistenceService)
+    }
+
+    def "Invoke JobKillServiceV3 for a v3 job"() {
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> false
+        0 * jobKillServiceV4.killJob(jobId, "testing")
+        1 * jobKillServiceV3.killJob(jobId, "testing")
+    }
+
+    def "Invoke JobKillServiceV4 for a v4 job"() {
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> true
+        1 * jobKillServiceV4.killJob(jobId, "testing")
+        0 * jobKillServiceV3.killJob(jobId, "testing")
+    }
+
+    def "Percolate up the exception thrown by underlying services when killing a job"() {
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> { throw new GenieJobNotFoundException()}
+        thrown(GenieJobNotFoundException)
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> true
+        1 * jobKillServiceV4.killJob(jobId, "testing") >> { throw new GenieServerException("Error killing v4 job")}
+        thrown(GenieServerException)
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> false
+        1 * jobKillServiceV3.killJob(jobId, "testing") >> { throw new GenieServerException("Error killing v3 job")}
+        thrown(GenieServerException)
+    }
+}
--- a/genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
+++ b/genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
@@ -44,6 +44,7 @@ import com.netflix.genie.web.jpa.entities.FileEntity;
 import com.netflix.genie.web.jpa.entities.JobEntity;
 import com.netflix.genie.web.jpa.entities.TagEntity;
 import com.netflix.genie.web.jpa.entities.projections.v4.JobSpecificationProjection;
+import com.netflix.genie.web.jpa.entities.projections.v4.IsV4JobProjection;
 import com.netflix.genie.web.jpa.entities.projections.v4.V4JobRequestProjection;
 import com.netflix.genie.web.jpa.repositories.JpaApplicationRepository;
 import com.netflix.genie.web.jpa.repositories.JpaClusterRepository;
@@ -738,6 +739,62 @@ public class JpaJobPersistenceServiceImplUnitTests {
         Mockito.when(jobEntity.isResolved()).thenReturn(false);
 
         Assert.assertFalse(this.jobPersistenceService.getJobSpecification(UUID.randomUUID().toString()).isPresent());
+    }
+
+    /**
+     * If a job isn't found on querying for v4 flag.
+     */
+    @Test(expected = GenieJobNotFoundException.class)
+    public void noJobUnableToGetV4() {
+        Mockito
+            .when(this.jobRepository.findByUniqueId(Mockito.anyString(), Mockito.eq(IsV4JobProjection.class)))
+            .thenReturn(Optional.empty());
+
+        this.jobPersistenceService.isV4(UUID.randomUUID().toString());
+    }
+
+
+    /**
+     * If v4 is false in db then return false.
+     */
+    @Test
+    public void v4JobFalseReturnsFalse() {
+        final JobEntity jobEntity = Mockito.mock(JobEntity.class);
+        Mockito
+            .when(this.jobRepository.findByUniqueId(Mockito.anyString(), Mockito.eq(IsV4JobProjection.class)))
+            .thenReturn(Optional.of(jobEntity));
+
+        Mockito.when(jobEntity.isV4()).thenReturn(false);
+
+        Assert.assertFalse(this.jobPersistenceService.isV4(UUID.randomUUID().toString()));
+    }
+
+    /**
+     * If v4 is true in db then return true.
+     */
+    @Test
+    public void v4JobTrueReturnsTrue() {
+        final JobEntity jobEntity = Mockito.mock(JobEntity.class);
+        Mockito
+            .when(this.jobRepository.findByUniqueId(Mockito.anyString(), Mockito.eq(IsV4JobProjection.class)))
+            .thenReturn(Optional.of(jobEntity));
+
+        Mockito.when(jobEntity.isV4()).thenReturn(true);
+
+        Assert.assertTrue(this.jobPersistenceService.isV4(UUID.randomUUID().toString()));
+    }
+
+    /**
+     * If v4 job is not found in the db then throw a Exception.
+     */
+    @Test(expected = GenieJobNotFoundException.class)
+    public void v4JobNotFoundThrowsException() {
+        final JobEntity jobEntity = null;
+        Mockito
+            .when(this.jobRepository.findByUniqueId(Mockito.anyString(), Mockito.eq(IsV4JobProjection.class)))
+            .thenReturn(Optional.ofNullable(jobEntity));
+
+        this.jobPersistenceService.isV4(UUID.randomUUID().toString());
     }
 
     /**
--- a/genie-web/src/test/java/com/netflix/genie/web/services/impl/JobKillServiceV3UnitTests.java
+++ b/genie-web/src/test/java/com/netflix/genie/web/services/impl/JobKillServiceV3UnitTests.java
@@ -49,13 +49,13 @@ import java.util.Optional;
 import java.util.UUID;
 
 /**
- * Unit tests for the LocalJobKillServiceImpl class.
+ * Unit tests for the JobKillServiceV3 class.
  *
  * @author tgianos
  * @since 3.0.0
  */
 @Category(UnitTest.class)
-public class LocalJobKillServiceImplUnitTests {
+public class JobKillServiceV3UnitTests {
 
     private static final String ID = UUID.randomUUID().toString();
     private static final String HOSTNAME = UUID.randomUUID().toString();
@@ -64,12 +64,13 @@ public class LocalJobKillServiceImplUnitTests {
     private CommandLine killCommand;
     private JobSearchService jobSearchService;
     private Executor executor;
-    private LocalJobKillServiceImpl service;
+    private JobKillServiceV3 service;
     private GenieEventBus genieEventBus;
     private FileSystemResource genieWorkingDir;
 
     /**
      * Setup for the tests.
+     *
      *
      * @throws IOException if the job directory cannot be created
      */
@@ -82,7 +83,7 @@ public class LocalJobKillServiceImplUnitTests {
         this.jobSearchService = Mockito.mock(JobSearchService.class);
         this.executor = Mockito.mock(Executor.class);
         this.genieEventBus = Mockito.mock(GenieEventBus.class);
-        this.service = new LocalJobKillServiceImpl(
+        this.service = new JobKillServiceV3(
             HOSTNAME,
             this.jobSearchService,
             this.executor,
@@ -218,7 +219,7 @@ public class LocalJobKillServiceImplUnitTests {
      */
     @Test
     public void canKillJobRunningAsUser() throws GenieException, IOException {
-        this.service = new LocalJobKillServiceImpl(
+        this.service = new JobKillServiceV3(
             HOSTNAME,
             this.jobSearchService,
             this.executor,
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobKillService.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobKillService.java
@@ -1,0 +1,42 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.execution.services;
+
+import javax.validation.constraints.NotBlank;
+
+/**
+ * Register an agent to listen for job kill messages from the server
+ * and respond to them.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public interface AgentJobKillService {
+    /**
+     * Start listening for job termination notification.
+     * @param jobId job id
+     */
+    void start(@NotBlank(message = "Job id cannot be blank") final String jobId);
+
+    /**
+     * Stop the service.
+     *
+     */
+    void stop();
+}
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillServiceImpl.java
@@ -1,0 +1,138 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.agent.execution.services.impl.grpc;
+
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.netflix.genie.agent.execution.services.AgentJobKillService;
+import com.netflix.genie.agent.execution.services.KillService;
+import com.netflix.genie.proto.JobKillRegistrationRequest;
+import com.netflix.genie.proto.JobKillRegistrationResponse;
+import com.netflix.genie.proto.JobKillServiceGrpc;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.stereotype.Service;
+
+import javax.validation.constraints.NotBlank;
+
+/**
+ * Implementation of the {@link AgentJobKillService}.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Lazy
+@Service
+@Slf4j
+public class GRpcAgentJobKillServiceImpl implements AgentJobKillService {
+
+    private final JobKillServiceGrpc.JobKillServiceFutureStub client;
+    private final KillService killService;
+    private ListenableFuture<JobKillRegistrationResponse> jobKillFuture;
+    private boolean started;
+    private final TaskExecutor killTaskExecutor;
+
+
+    /**
+     * Constructor.
+     *
+     * @param client           The gRPC client to use to call the server
+     * @param killService      KillService for killing the agent
+     * @param killTaskExecutor A task executor to execute killing the agent
+     */
+    public GRpcAgentJobKillServiceImpl(
+        final JobKillServiceGrpc.JobKillServiceFutureStub client,
+        final KillService killService,
+        @Qualifier("sharedAgentTaskExecutor") final TaskExecutor killTaskExecutor
+    ) {
+        this.client = client;
+        this.killService = killService;
+        this.killTaskExecutor = killTaskExecutor;
+    }
+
+    @Override
+    public synchronized void start(@NotBlank(message = "Job id cannot be blank") final String jobId) {
+        //Service can be started only once
+        if (started) {
+           throw new IllegalStateException("Service can be started only once");
+        }
+
+        registerForRemoteKillNotification(jobId);
+        started = true;
+    }
+
+    @Override
+    public void stop() {
+        cancelPreviousAndUpdateJobKillFuture(null);
+    }
+
+    private void registerForRemoteKillNotification(final String jobId) {
+
+        final ListenableFuture<JobKillRegistrationResponse> future = this.client
+            .registerForKillNotification(
+                JobKillRegistrationRequest.newBuilder()
+                    .setJobId(jobId)
+                    .build()
+            );
+
+        cancelPreviousAndUpdateJobKillFuture(future);
+
+        Futures.addCallback(
+            future,
+            new JobKillFutureCallback(jobId),
+            this.killTaskExecutor
+        );
+    }
+
+    /* Cancel the current future */
+    private synchronized void cancelJobKillFuture() {
+        if (this.jobKillFuture != null) {
+            this.jobKillFuture.cancel(false);
+        }
+    }
+
+    /* Cancel current future and store the new future */
+    private synchronized void cancelPreviousAndUpdateJobKillFuture(
+        final ListenableFuture<JobKillRegistrationResponse> killFuture
+    ) {
+        cancelJobKillFuture();
+        this.jobKillFuture = killFuture;
+    }
+
+    private class JobKillFutureCallback implements FutureCallback<JobKillRegistrationResponse> {
+
+        private final String jobId;
+
+        JobKillFutureCallback(final String jobId) {
+            this.jobId = jobId;
+        }
+
+        @Override
+        public void onSuccess(final JobKillRegistrationResponse result) {
+            killService.kill();
+        }
+
+        @Override
+        public void onFailure(final Throwable t) {
+            registerForRemoteKillNotification(this.jobId);
+        }
+    }
+}
--- a/genie-agent/src/main/java/com/netflix/genie/agent/rpc/GrpcConfig.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/rpc/GrpcConfig.java
@@ -21,6 +21,7 @@ package com.netflix.genie.agent.rpc;
 import com.netflix.genie.agent.cli.ArgumentDelegates;
 import com.netflix.genie.proto.HeartBeatServiceGrpc;
 import com.netflix.genie.proto.JobServiceGrpc;
+import com.netflix.genie.proto.JobKillServiceGrpc;
 import com.netflix.genie.proto.PingServiceGrpc;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
@@ -68,5 +69,13 @@ class GrpcConfig {
     @Scope("prototype")
     HeartBeatServiceGrpc.HeartBeatServiceStub heartBeatClient(final ManagedChannel channel) {
         return HeartBeatServiceGrpc.newStub(channel);
+    }
+
+    @Bean
+    @Scope("prototype")
+    JobKillServiceGrpc.JobKillServiceFutureStub jobKillClient(
+        final ManagedChannel channel
+    ) {
+        return JobKillServiceGrpc.newFutureStub(channel);
     }
 }
--- a/genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillServiceImplSpec.groovy
+++ b/genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillServiceImplSpec.groovy
@@ -1,0 +1,170 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.execution.services.impl.grpc
+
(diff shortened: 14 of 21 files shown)


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobKillRegistrationService.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author standon
+ * @since 4.0.0
+ */
+public interface AgentJobKillRegistrationService {
+    /**
+     * Register the agent to listen for job kill messages.
+     * @param jobId id of the job agent is running
+     * @throws JobKillRegistrationException if registration fails
+     */
+    void registerAgent(@NotBlank final String jobId) throws JobKillRegistrationException;
```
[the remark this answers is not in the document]
sumitnetflix: My thought was that the service name `AgenJobKillRegistrationService` provides that context. Hence did not include that in the name. Will change to `registerForKillNotification` if that seems more clearer.

Will explore the second point you made

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillRegistrationServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    public GRpcAgentJobKillRegistrationServiceImpl(
+        final JobKillServiceGrpc.JobKillServiceFutureStub client
+    ) {
+        this.client = client;
+    }
+
+
+    @Override
+    public void registerAgent(@NotBlank final String jobId) throws JobKillRegistrationException {
+        final ListenableFuture<JobKillListenerResponse> requestFuture = this.client
```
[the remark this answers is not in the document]
sumitnetflix: sure

genie-proto/src/main/proto/genie.proto
```diff
@@ -213,3 +213,15 @@ message ChangeJobStatusResponse {
...
+service JobKillService {
+    rpc registerAgent (JobKillListenerRequest) returns (JobKillListenerResponse);
+}
+
+message JobKillListenerRequest {
+    string job_id = 1;
+}
+
+message JobKillListenerResponse {
+    string job_id = 1;
```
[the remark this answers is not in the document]
sumitnetflix: Did'nt have it initially. Thought of adding that since the agent should ideally check that the notification is indeed for itself before executing the kill. Just a safeguard mechanism.

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -206,4 +223,8 @@ private void killJobOnUnix(final int pid) throws GenieException {
             throw new GenieServerException("Unable to kill process " + pid, ioe);
         }
     }
+
+    public boolean isV4Job() {
```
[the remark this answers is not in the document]
sumitnetflix: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobKillRegistrationService.java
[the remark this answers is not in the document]
mprimi: And maybe just name it `KillService` or `JobTerminationService`, `RemoteJobTerminationService`?
The fact that there is a registration involved seems to be an implementation/internal detail.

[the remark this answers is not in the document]
mprimi: An alternative idea for this interface is to pass in an observer/listener that gets a callback if a kill signal is delivered.
As this is right now there's no way to link the signal received to the subsystems that are actually interested in receiving such signal (other than hardcoding a call to those services in your hardcoded callback).
This unnecessarily couples this service with those consumers.

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -48,6 +49,8 @@
 import java.io.IOException;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.Map;
+import java.util.function.Consumer;
 
 /**
  * Implementation of the JobKillService interface which attempts to kill jobs running on the local node.
```
mprimi: This comment is now out of date.

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -108,6 +112,12 @@ public void killJob(
         @NotBlank(message = "No id entered. Unable to kill job.") final String id,
         @NotBlank(message = "No reason provided.") final String reason
     ) throws GenieException {
+
+        if (isV4Job()) {
+            this.jobKillCallbacks.get(id).accept(id);
```
mprimi: Possible NPE here

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -206,4 +249,8 @@ private void killJobOnUnix(final int pid) throws GenieException {
             throw new GenieServerException("Unable to kill process " + pid, ioe);
         }
     }
+
+    private boolean isV4Job(final String id) {
```
tgianos: this method seems useless. Just inline the call the the persistence service.

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -108,6 +118,14 @@ public void killJob(
         @NotBlank(message = "No id entered. Unable to kill job.") final String id,
         @NotBlank(message = "No reason provided.") final String reason
     ) throws GenieException {
+
+        //TODO possible optimization. Have a single projection for getting hostname and v4 job and pass isV4.
+        //That can save one db call
+        if (isV4Job(id)) {
```
tgianos: You may want to put this after the logic that determines if a job is already finished or not below

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -176,6 +204,21 @@ public void onKillJobEvent(@NotNull final KillJobEvent event) throws GenieExcept
         this.killJob(event.getId(), event.getReason());
     }
 
+    @Override
+    public void registerCallbackForKill(
```
tgianos: it seems you're missing a deregister? Even if we don't have the outside service to shoulder tap you that it isn't connected to this node anymore I would think that after you invoke the callback on a kill signal you'd still somehow want to remove it from the map unless you're solely planning to rely on that other mechanism

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        ) {
+            this.onSuccessCallback = onSuccessCallback;
+        }
+
+        @Override
+        public void onSuccess(@Nullable final JobKillListenerResponse result) {
+            final String jobId = executionContext.getClaimedJobId();
+
+            if (result != null && result.getJobId().equals(jobId)) {
+                handleKillNotification(onSuccessCallback);
```
tgianos: I don't understand this calling in and out of the listener?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                    JobKillListenerRequest.newBuilder()
+                        .setJobId(jobId)
+                        .build()
+                ),
+            new JobKillFutureCallback(onKillCallback)
+        );
+    }
+
+    private void handleErrorListeningForKillNotification(final Runnable onSuccessCallback) {
+        registerForRemoteKillNotification(onSuccessCallback);
```
tgianos: This seems like a possible infinite loop

genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/JobV4Projection.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package com.netflix.genie.web.jpa.entities.projections.v4;
+
+/**
+ * Projection for capturing v4 field from the database.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public interface JobV4Projection {
```
tgianos: I'd make this `IsV4JobProjection`

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
```diff
@@ -465,6 +466,28 @@ public void saveJobSpecification(
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Transactional(readOnly = true)
+    public boolean isV4(
+        @NotBlank(message = "Id is missing and is required") final String id
+    ) {
+        log.debug("Requested to get v4 for job {}", id);
+        final JobV4Projection projection = this.jobRepository
```
tgianos: Unnecessary to assign this to a variable just return

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
```diff
@@ -465,6 +466,28 @@ public void saveJobSpecification(
...
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Transactional(readOnly = true)
+    public boolean isV4(
+        @NotBlank(message = "Id is missing and is required") final String id
+    ) {
+        log.debug("Requested to get v4 for job {}", id);
```
tgianos: This could be clearer, something like `Request to figure out whether job {} is a v4 job or not`

genie-web/src/main/java/com/netflix/genie/web/services/JobKillService.java
```diff
@@ -52,4 +53,15 @@ void killJob(@NotBlank(message = "No id entered. Unable to kill job.") final Str
...
     void onKillJobEvent(@NotNull final KillJobEvent event) throws GenieException;
+
+    /**
+     * Register a callback for killing the job.
+     *
+     * @param id     id of the job
+     * @param consumer consumer representing the callback to be executed
+     */
+
+    void registerCallbackForKill(
```
tgianos: I don't necessarily agree with making this a hard method of this class. This interface is to an API for simply killing jobs. Exposing a mechanism for registering callbacks exposes an internal detail. It may well be a valuable purpose but I think we could break it out into a separate interface. Then within your gRPC service you can instead inject a `List<KillCallbackListencers>` that way we can have multiple things with the ability to register. That explanation might not be clear. My brain isn't working at 100% at the moment haha.

Effectively there may be some value of providing a way for multiple services to do this or even better we limit the methods exposed based on which interface we inject into the consumers. e.g. Within your grpc service endpoint you only inject the interface that exposes this method. It shouldn't know about any other methods or that the service kills things. It only cares "give me something I can get notifications from".

Additionally we may have over complicated it if we want it totally decoupled we could use events which we already pub/sub and not reinvent this wheel. We can talk about it offline next week when @mprimi is back

This comment is a mess of thoughts so I'm going to cut it off now

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+        JobKillConsumer(final StreamObserver<JobKillListenerResponse> responseObserver) {
+            this.responseObserver = responseObserver;
+        }
+
+        @Override
+        public void accept(final String id) {
+            responseObserver.onNext(
+                JobKillListenerResponse.newBuilder().setJobId(id).build()
+            );
+            responseObserver.onCompleted();
```
tgianos: Once you've called `onCompleted` on the observer you can't use this `responseObserver` anymore. It will throw exceptions as you try to use it. So with that in mind once this i called this listener should be destroyed.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
sumitnetflix: Yup. Had discussed this with @mprimi. Instead of a retry policy we thought to have this as the initial implementation

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
sumitnetflix: Done

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
sumitnetflix: @tgianos I looked at the code and here are my thoughts. I am not clear about how the statuses are being updated today for v4 jobs by the agent. Atleast when a job is killed they are not updated. For e.g Is an exit code being populated for v4 jobs? Its being checked in this block of code 
```final JobExecution jobExecution = this.jobSearchService.getJobExecution(id);
            if (jobExecution.getExitCode().isPresent()) {
                // Job is already finished one way or another
                return;
            } ```
So its not clear to me where to fit the handling of v4 kill accept for right at the beginning. When I had an initial discussion with @mprimi we wanted the initial implementation to be a way (preferably without continuous polling) to convey to a agent that `kill yourself`. Having the handling of v4 jobs right in the beginning, as it is now, serves that purpose. So, we can have this component ready but their is no need for agent to register for kill notification until we have the entire kill functionality(status update, acknowledgement from the agent that its going to kill itself, actual kill functionality on the agent side) figured out. I have also added the code to throw an exception if a kill request comes from a v4 job and there is no registered callback. In case some one makes a kill request there will be an exception thrown clearly communicating that the job cannot be killed. 
This way we should be able to move forward while figuring out other details associated with kill.

File: genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java (hunk 9 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java]: Makes sense. Done

File: genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/JobV4Projection.java (hunk 12 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/JobV4Projection.java]: Done

File: genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java (hunk 13 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java]: Done

File: genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java (hunk 14 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java]: Done

File: genie-web/src/main/java/com/netflix/genie/web/services/JobKillService.java (hunk 15 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/services/JobKillService.java]: I have created a new interface - `JobKillCallbackService` and moved the methods to register and deregister callbacks there. `LocalJobKillServiceImpl` now implements this interface in addition `JobKillService`. I inject `JobKillCallbackService` in `GrpcJobTerminationServiceImpl`

File: genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java (hunk 16 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java]: Done

File: genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java (hunk 5 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java]: Fixed

Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java]: @tgianos I have also moved to Runnable from Consumer<String> as the callback for kill notifications as we discussed.

```@Override
    public void registerCallbackForKill(
        @NotBlank(message = "No id entered. Unable to register job for killing.") final String id,
        @NotNull final Runnable callback) {

        jobKillCallbacks.put(id, callback);

    }```

File: genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java [hunk 17]
```diff
@@ -0,0 +1,118 @@
...
+    private void handleKillNotification(final Runnable callback) {
+        callback.run();
+    }
+
+    private class JobKillFutureCallback implements FutureCallback<JobKillListenerResponse> {
+
+        private final Runnable onSuccessCallback;
+
+        JobKillFutureCallback(
+            final Runnable onSuccessCallback
```
tgianos: For these single argument methods I don't think you need to put them on separate lines. I generally wait till the method is approaching the checkstyle limit before I break the lines apart. This is 3 lines right now when it doesn't have to be.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        ) {
+            this.onSuccessCallback = onSuccessCallback;
+        }
+
+        @Override
+        public void onSuccess(@Nullable final JobKillListenerResponse result) {
+            final String jobId = executionContext.getClaimedJobId();
```
tgianos: Rather than this why do you save the entire `ExecutionContext` as a member rather than just the job id and save this call to the getter?

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImplSpec.groovy
```diff
@@ -0,0 +1,133 @@
...
+        Runnable onSuccessCallback = Mock();
+        simulationCallback = new Runnable() {
+            @Override
+            void run() {}
+        }
+
+        when:
+        service.registerForRemoteKillNotification(onSuccessCallback)
+
+        then: "onSuccess callback gets invoked"
```
tgianos: If you're going to put a label on the there you should probably put a label on the `when` as well. It should tell a story ideally.  `when X happens then Y happens`

genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/IsV4JobProjection.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Projection for capturing v4 field from the database.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public interface IsV4JobProjection {
+    /**
+     * Get the v4 flag.
```
tgianos: Nitpick but this should really be something like "Get whether this job was launched using a v4 mechanism or not" and then the return should say `true if the job is v4, false it if is v3`

We might also want to flag the whole method as `@deprecated` to make it clear we don't plan to keep it around

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    ) {
+        jobKillCallbackService.registerCallbackForKill(
+            request.getJobId(),
+            new JobKillCallback(responseObserver, request.getJobId())
+        );
+    }
+
+    //Testing utility
+    @VisibleForTesting
+    JobKillCallback getCallBack(
```
tgianos: This seems like a useless method. Why don't you just make the class itself package protected and invoke the constructor directly fo testing?

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -165,6 +187,17 @@ public void killJob(
         }
     }
 
+    private void handleV4JobKill(final String id) throws GenieException {
+        if (this.jobKillCallbacks.containsKey(id)) {
+            this.jobKillCallbacks.get(id).run();
+        } else {
+            log.error("Job not killed. No callback found for killing the job with id - {} ", id);
+            throw new GenieException(
+                HttpStatus.BAD_REQUEST.value(),
```
tgianos: 1. There's already a `GenieBadRequestException` you can use rather than hand rolling your own code here
2. I don't think this would be a `BadRequest` as that is a 4xx code which is a user error. In this case it's likely not a user error. If they request to kill a job that is already killed or can't be found we may just want to swallow it. Otherwise i would think it's some sort of `InternalServerError`

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -176,6 +209,28 @@ public void onKillJobEvent(@NotNull final KillJobEvent event) throws GenieExcept
...
+
+        jobKillCallbacks.put(id, callback);
+
+    }
+
+    @Override
+    public void deregisterCallbackForKill(
+        @NotBlank(message = "No id entered. Unable to deregister job for killing.") final String id
+    ) {
+        jobKillCallbacks.remove(id);
```
tgianos: Shouldn't you log something here?

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -176,6 +209,28 @@ public void onKillJobEvent(@NotNull final KillJobEvent event) throws GenieExcept
         this.killJob(event.getId(), event.getReason());
     }
 
+    @Override
+    public void registerCallbackForKill(
+        @NotBlank(message = "No id entered. Unable to register job for killing.") final String id,
+        @NotNull final Runnable callback) {
```
tgianos: The `) {` should be on the next line

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -176,6 +209,28 @@ public void onKillJobEvent(@NotNull final KillJobEvent event) throws GenieExcept
...
+    public void deregisterCallbackForKill(
+        @NotBlank(message = "No id entered. Unable to deregister job for killing.") final String id
+    ) {
+        jobKillCallbacks.remove(id);
+    }
+
+    @VisibleForTesting
+    //Testing utility
+    Map<String, Runnable> getJobKillCallbacks() {
+        return jobKillCallbacks;
```
tgianos: At the very least I would prefer you return some sort of Unmodifiable view out of this method using `Collections` class so that if some malicious piece of code tries to use this it can't actually modify your internal state. I don't even like the idea of exposing this at all.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
sumitnetflix: There were multiple arguments before. I removed a few but did not fix the new lines. Will fix this.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
sumitnetflix: sure

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImplSpec.groovy
sumitnetflix: 👍

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
sumitnetflix: Tried that approach earlier but for some reason was not able to do that in groovy. So, created this. Moved back to package level class and have created a junit based test. Works now.

genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/IsV4JobProjection.java
sumitnetflix: done

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -108,9 +120,21 @@ public void killJob(
...
         @NotBlank(message = "No reason provided.") final String reason
     ) throws GenieException {
+
         // Will throw exception if not found
         // TODO: Could instead check JobMonitorCoordinator eventually for in memory check
         final JobStatus jobStatus = this.jobSearchService.getJobStatus(id);
+
+        //TODO possible optimization. Have a single projection for getting hostname and isV4Job.
+        //Pass isV4Job as a method parameter. That will save a db call.
+        final boolean isV4Job = jobPersistenceService.isV4(id);
```
tgianos: Do you use this anywhere else? any reason not to just have it in the `if()` directly?

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
```diff
@@ -176,6 +222,24 @@ public void onKillJobEvent(@NotNull final KillJobEvent event) throws GenieExcept
...
+        jobKillCallbacks.put(id, callback);
+
+    }
+
+    @Override
+    public void deregisterCallbackForKill(
+        @NotBlank(message = "No id entered. Unable to deregister job for killing.") final String id
+    ) {
+        jobKillCallbacks.remove(id);
+        log.info("Deregistered kill callback for job id {}", id);
```
tgianos: I think I meant you should check the return from the `remove` API to see if it was actually removed or not

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalJobKillServiceImpl.java
sumitnetflix: had restructrued the code and this variable was left out. Fixed

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param client           The gRPC client to use to call the server.
+     * @param executionContext The execution context for the agent
+     */
+    public GRpcJobTerminationListenerServiceImpl(
+        final JobTerminationServiceGrpc.JobTerminationServiceFutureStub client,
+        final ExecutionContext executionContext
+    ) {
+        this.client = client;
+        this.jobId = executionContext.getClaimedJobId();
```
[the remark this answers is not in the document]
sumitnetflix: @tgianos I see your point. I will add a check in the constructor that if job id is null it will throw an exception. That ways we avoid there being a missing job id. We can always make sure that the agent interacts with this service at a point when we have a job id. I can make that change once we have decided what that flow is. However, since no one is registering with the service right now this should be fine. @mprimi Let me know if there are any other thoughts.

genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/IsV4JobProjection.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author standon
+ * @since 4.0.0
+ */
+public interface IsV4JobProjection {
+    /**
+     * Job was launched using Genie v4.
+     *
+     * @return true if job was launched via Genie v4 else false
+     */
+    @Deprecated
```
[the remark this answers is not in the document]
sumitnetflix: done

genie-web/src/main/java/com/netflix/genie/web/services/JobKillCallbackService.java
```diff
@@ -0,0 +1,55 @@
...
+        @NotBlank(message = "No id entered. Unable to register job for killing.") final String id,
+        @NotNull final Runnable callback
+    );
+
+    /**
+     * Deregister the callback for killing a job.
+     *
+     * @param id     id of the job
+     */
+
```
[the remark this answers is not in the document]
sumitnetflix: Fixed

genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
```diff
@@ -740,6 +741,49 @@ public void unresolvedJobReturnsEmptyJobSpecificationOptional() {
...
+        final JobEntity jobEntity = Mockito.mock(JobEntity.class);
+        Mockito
+            .when(this.jobRepository.findByUniqueId(Mockito.anyString(), Mockito.eq(IsV4JobProjection.class)))
+            .thenReturn(Optional.of(jobEntity));
+
+        Mockito.when(jobEntity.isV4()).thenReturn(true);
+
+        Assert.assertTrue(this.jobPersistenceService.isV4(UUID.randomUUID().toString()));
+    }
+
```
mprimi: Add a test for PersistenceService throwing exception?

genie-web/src/test/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImplUnitTests.java
```diff
@@ -0,0 +1,72 @@
+/*
```
mprimi: All other new tests are in Spock, any particular reason for going with JUnit?

genie-proto/src/main/proto/genie.proto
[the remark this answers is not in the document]
mprimi: `JobKillListenerRequest` -> `JobKillRequest`
`JobKillListenerResponse` -> `JobKillResponse`

genie-web/src/test/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImplUnitTests.java
mprimi: This comment is on `GRpcJobTerminationServiceImplUnitTests.java` which you added.
And I was wondering why it is JUnit and not Spock

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        SimpleLoggingInterceptor.class,
+    }
+)
+@Slf4j
+public class GRpcJobTerminationServiceImpl
+    extends JobTerminationServiceGrpc.JobTerminationServiceImplBase
+    implements JobKillServiceV4 {
+
+    private final Map<String, StreamObserver<JobKillListenerResponse>> activeStreams = Maps.newConcurrentMap();
+    private final JobSearchService jobSearchService;
```
mprimi: Why job search service if you're always looking up records by ID?
Did Tom suggest this? Seems strange to me.

genie-web/src/test/groovy/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImplSpec.groovy
```diff
@@ -0,0 +1,96 @@
...
+    def "Can register for kill notification"() {
+
+        when:
+        service.registerForKillNotification(request, responseObserver)
+
+        then:
+        service.activeStreams.get(jobId) != null
+    }
+
+    def "Can kill unfinished jobs and clean up response observer"() {
```
mprimi: I think this test covers a superset of the lines covered in the previous one (i.e. the previous does not add anything)

genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
```diff
@@ -0,0 +1,75 @@
...
+    JobKillServiceV4 jobKillServiceV4
+    String jobId
+
+    void setup() {
+        jobPersistenceService = Mock()
+        jobKillServiceV4 = Mock()
+        jobKillServiceV3 = Mock()
+        jobId = UUID.randomUUID().toString()
+        service = new JobKillServiceImpl (
+                jobKillServiceV3, \
```
mprimi: What's up with those slashes?

genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
```diff
@@ -740,6 +741,62 @@ public void unresolvedJobReturnsEmptyJobSpecificationOptional() {
...
+    /**
+     * If v4 is true in db then return true.
```
mprimi: Wrong comment

genie-web/src/test/java/com/netflix/genie/web/services/impl/JobKillServiceV3UnitTests.java
```diff
@@ -49,28 +49,29 @@
...
  * @author tgianos
  * @since 3.0.0
  */
 @Category(UnitTest.class)
-public class LocalJobKillServiceImplUnitTests {
+public class JobKillServiceV3UnitTests {
 
     private static final String ID = UUID.randomUUID().toString();
-    private static final String HOSTNAME = UUID.randomUUID().toString();
+    private static final String hostName = UUID.randomUUID().toString();
```
mprimi: This is a constant, i'm surprised checkstyle doesn't complain about it being lowercase..

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobTerminationListenerService.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.validation.constraints.NotBlank;
+
+/**
+ * Register an agent to listen for job kill messages from the server
+ * and respond to them.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public interface JobTerminationListenerService {
```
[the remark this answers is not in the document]
sumitnetflix: Hmm. What i had in mind was that on the server side I call the service as `JobTerminationService` and agent side `JobTerminationListenerService` as its listening for a termination. We can move to `AgentJobTerminationService`?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/KillServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.stereotype.Service;
+
+/**
+ * Implementation of {@link KillService}.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Service
```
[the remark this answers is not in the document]
sumitnetflix: Right. Thanks

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImpl.java
```diff
@@ -0,0 +1,146 @@
...
+    ) {
+        this.client = client;
+        this.killService = killService;
+        this.killTaskExecutor = killTaskExecutor;
+    }
+
+    @Override
+    public synchronized void start(@NotBlank(message = "Job id cannot be blank") final String jobId) {
+        //Service can be started only once
+        if (started.get()) {
```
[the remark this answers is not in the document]
sumitnetflix: Missed that. Started with AtomicBoolean before leaving for india. Will fix

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImplSpec.groovy
```diff
@@ -0,0 +1,133 @@
...
+     * Mock service for testing.
+     */
+    private class TestJobTerminationService extends JobTerminationServiceGrpc.JobTerminationServiceImplBase {
+
+        @Override
+        public void registerForKillNotification(
+                final JobKillListenerRequest request,
+                final StreamObserver<JobKillListenerResponse> responseObserver
+        ) {
+            simulationCallback.run();
```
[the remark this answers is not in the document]
sumitnetflix: I could think of another way. Not sure if it will work. Lets discuss f2f

genie-web/src/main/java/com/netflix/genie/web/configs/GenieServicesAutoConfiguration.java
```diff
@@ -122,6 +123,28 @@ public JobKillService jobKillService(
...
+     * @param jobKillServiceV3   Service to kill V3 jobs.
+     * @param jobKillServiceV4      Service to kill V4 jobs.
+     * @param jobPersistenceService Job persistence service
+     * @return A job kill service instance.
+     */
+    @Bean
+    @ConditionalOnMissingBean(JobKillService.class)
+    public JobKillService jobKillService(
+        final JobKillServiceV3 jobKillServiceV3,
+        final JobKillServiceV4 jobKillServiceV4,
```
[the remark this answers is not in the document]
sumitnetflix: Its `GRpcJobTerminationServiceImpl`

genie-web/src/main/java/com/netflix/genie/web/jpa/entities/projections/v4/IsV4JobProjection.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.netflix.genie.web.jpa.entities.projections.v4;
+
+/**
+ * Projection for capturing if the job was launched using Genie v4.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Deprecated
```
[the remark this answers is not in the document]
sumitnetflix: @tgianos said that once v4 is fully live this method will be useless. We wanted to deprecate it to remember that and not use it in more places.

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
```diff
@@ -463,6 +464,26 @@ public void saveJobSpecification(
...
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Transactional(readOnly = true)
+    public boolean isV4(
+        @NotBlank(message = "Id is missing and is required") final String id
+    ) {
+        log.debug("Requested to figure out if the job {} is a v4 job", id);
```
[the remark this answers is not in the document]
sumitnetflix: Sure. Will move to read

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+    interceptors = {
+        SimpleLoggingInterceptor.class,
+    }
+)
+@Slf4j
+public class GRpcJobTerminationServiceImpl
+    extends JobTerminationServiceGrpc.JobTerminationServiceImplBase
+    implements JobKillServiceV4 {
+
+    private final Map<String, StreamObserver<JobKillListenerResponse>> activeStreams = Maps.newConcurrentMap();
```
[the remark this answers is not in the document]
sumitnetflix: LiveStreams?

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
sumitnetflix: Its an already existing service. Only using it.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Constructor.
+     * @param jobSearchService      Job search service
+     */
+    public GRpcJobTerminationServiceImpl(final JobSearchService jobSearchService) {
+        this.jobSearchService = jobSearchService;
+    }
+
+    /**
+     * Register to get a callback when a job is killed.
```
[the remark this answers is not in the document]
sumitnetflix: Ahh.. its the old comment when we had callbacks. Thanks. Will fix

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @param request          Request to listen for job registerCallback
```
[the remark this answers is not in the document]
sumitnetflix: My bad.. Old comment

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    public void registerForKillNotification(
+        final JobKillListenerRequest request,
+        final StreamObserver<JobKillListenerResponse> responseObserver
+    ) {
+        activeStreams.put(request.getJobId(), responseObserver);
+    }
+
+    @Override
+    public void killJob(final @NotBlank(message = "No id entered. Unable to kill job.") String id,
```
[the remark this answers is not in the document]
sumitnetflix: Will fix

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        if (!jobSearchService.getJobStatus(id).isFinished()) {
+            if (responseObserver != null) {
+                responseObserver.onNext(
+                    JobKillListenerResponse.newBuilder().setJobId(id).build()
+                );
+                responseObserver.onCompleted();
+
+
+                log.info("Agent communicated for killing the job ");
+            } else { //This is bad. No responseobserver for a unfinished job. Tell the user
```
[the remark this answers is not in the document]
sumitnetflix: Comment can be improved. But it is bad in the sense that this should not happen. That tells we routed to the wrong box. This will be executed in response to the kill request from a user. Since we are throwing the exception the exception will go to the user. Hence the comment(which I will change) but the user will know

genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
sumitnetflix: Groovy multiple lines format

genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
sumitnetflix: Did that

genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
sumitnetflix: Ah Cut and Paste is the villain Thanks

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/KillService.java
```diff
@@ -0,0 +1,32 @@
...
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.execution.services;
+
+/**
+ * Service responsible for killing the agent.
```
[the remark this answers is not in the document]
tgianos: Agreed

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobTerminationListenerService.java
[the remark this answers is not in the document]
tgianos: What is the difference between a `Termination` and a `Kill`. We should be consistent or we're going to think they have a different purposes when they're part of a single purpose.

Why isn't this just something like `JobKillRpcService`?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/KillServiceImpl.java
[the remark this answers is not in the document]
tgianos: I've told @mprimi this while you were gone but we long term might want to get away from doing annotation based beans like `@Component` or `@Service` in anything that could be re-wrapped internal as it makes it hard to override them if necessary. Not critical to do right now but the auto-configurations with conditionals makes it easier to manipulate it as a library

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/KillServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+@Lazy
+@Slf4j
+@Component
+class KillServiceImpl implements KillService {
+
+    @Override
```
tgianos: Need to add `{@inheritDoc}`

genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
sumitnetflix: Groovy is different. You cannot break a sting across multiple lines without `\`. I encountered that problem once and adding `\` worked. So this was my logical extension. As explained here - https://stackoverflow.com/questions/5079797/whats-wrong-with-groovy-multi-line-string - `\` is not needed for a constructor as groovy has a context and hence can find that multiple lines belong to the constructor. Removed `\`.

mprimi: This is not a string. Groovy and Java behave the same. There's no need for those backslashes.
Just look at any Groovy file in the repository, e.g.: https://github.com/Netflix/genie/blob/master/genie-agent/src/test/groovy/com/netflix/genie/agent/execution/statemachine/StateMachineConfigSpec.groovy#L43-L53

sumitnetflix: I am not saying that this is a string. Hence removed the `\`. I was trying to tell you how I ended up adding those. Groovy and java do behave differently if you try to write a multi line string(Example -    
 String data = "a" + "b" 
    + "c"
). The link you provided is for a method call. There `\` will not be needed.

genie-proto/src/main/proto/genie.proto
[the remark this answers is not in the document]
sumitnetflix: Will do

genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
sumitnetflix: Still is valid. The test was missing

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/KillService.java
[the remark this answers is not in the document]
sumitnetflix: Fixed

genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
sumitnetflix: As I mentioned earlier, I was just trying to give the context and trying to make it clear that Java and groovy do behave differently. The slashes are already removed.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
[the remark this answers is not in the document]
sumitnetflix: Nothing is but there is a live stream/connection between the server and client. `openConnections`?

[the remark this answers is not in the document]
sumitnetflix: `killResponseObservers`?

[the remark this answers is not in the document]
mprimi: Stream and connection are not the same thing. 
In this case there is an open request, an underlying connection, but no stream.

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImplSpec.groovy
```diff
@@ -0,0 +1,192 @@
...
+        this.client = JobTerminationServiceGrpc.newFutureStub(grpcServerRule.getChannel())
+
+        when: "Starting the service"
+        service.start(null)
+
+        then: "Exception gets thrown"
+        thrown(NullPointerException)
+    }
+
+    def "Restarting service throws Exception"() {
```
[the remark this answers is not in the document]
mprimi: Double is what your test does. Call start 2 times.
"Multiple" could also be ok, if this is a generic test for multiple starts (which i think it is)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobTerminationListenerService.java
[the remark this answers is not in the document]
sumitnetflix: For agent side - Other services in the same package are named as `Agent*`. I am hinging on naming the service `AgentJobKillRpcService`. The service which will actually do the kill, we can call `JobKillService`

For the server - We can call it `JobKillService`. Since its in a different package its clear from the context that this is server side kill.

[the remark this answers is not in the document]
mprimi: Notice the convention for other services is:
`AgentFooService` (interface)
`GRpcAgentFooServiceImpl` (gRPC implementation)

genie-proto/src/main/proto/genie.proto
```diff
@@ -304,3 +304,14 @@ message AgentHeartBeat {
...
+service JobTerminationService {
+    rpc registerForKillNotification (JobKillListenerRequest) returns (JobKillListenerResponse);
+}
+
+message JobKillListenerRequest {
+    string job_id = 1;
+}
+
+message JobKillListenerResponse {
+}
```
sumitnetflix: How about `JobKillRegistrationRequest\Response` as these are request\response to register for a job kill notification

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+    ) {
+        parkedJobKillResponseObservers.put(request.getJobId(), responseObserver);
+    }
+
+    @Override
+    public void killJob(final @NotBlank(message = "No job id entered. Unable to kill job.") String jobId,
+                        final @NotBlank(message = "No reason provided.") String reason) throws GenieException {
+
+        final StreamObserver<JobKillListenerResponse> responseObserver =
+            cleanUpAndGetParkedJobKillResponseObserver(jobId);
```
sumitnetflix: The more I think about it, I think we should remove cleanup of response obsevers completely from here and have the cleanup done only when a disconnect happens. Thoughts?

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+            log.info("v4 job {} was already finished when the kill request came", jobId);
+        }
+    }
+
+    private StreamObserver<JobKillListenerResponse> cleanUpAndGetParkedJobKillResponseObserver(final String jobId) {
+
+        final StreamObserver<JobKillListenerResponse> responseObserver = parkedJobKillResponseObservers.remove(jobId);
+
+        //Note that this is possible in race conditions like job was finished so a clean up happened.
```
sumitnetflix: If we don clean up only on agent's disconnection then this will also not happen

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+        //When a disconnect happens then a clean up will happen because heart beat service will also send
+        //a notification. At this point there would be no response observer in the parkedJobKillResponseObservers map
```
sumitnetflix: If we don clean up only on agent's disconnection then this will also not happen

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+            );
+            responseObserver.onCompleted();
+
+            log.info("Agent notified for killing the job ");
+        } else if (responseObserver == null) {
+            log.error("Job not killed. No response observer found for killing the job with id - {} ", jobId);
+            throw new GenieServerException(
+                "Job not killed. No response observer found for killing the job with id - {} " + jobId);
+        } else {
+            log.info("v4 job {} was already finished when the kill request came", jobId);
```
mprimi: Wouldn't this be more readable as:
```
if observer == null
  print something
if not finished
  print something
else
  send kill
```

Personal preference, but seems more clear/elegant to me than checking the same `null` twice

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImpl.java
mprimi: I think this whole method can go, no need for this comment or these INFO messages

genie-web/src/test/groovy/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImplSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+    void setup() {
+        jobId = UUID.randomUUID().toString()
+        request = JobKillListenerRequest.newBuilder().setJobId(jobId).build()
+        response = JobKillListenerResponse.newBuilder().build()
+        service = new GRpcJobTerminationServiceImpl(jobSearchService)
+    }
+
+    def "Can kill unfinished jobs and clean up response observer"() {
+
+        service.registerForKillNotification(request, responseObserver)
```
mprimi: This is a call to the class under test. It should be in a `when` block as general rule/convention.

genie-web/src/test/groovy/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobTerminationServiceImplSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+        then:
+        1 * jobSearchService.getJobStatus(jobId) >> JobStatus.RUNNING
+        1 * responseObserver.onNext(response)
+        1 * responseObserver.onCompleted()
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        thrown(GenieServerException)
```
mprimi: As is, this will return a `500` error. Is that what we want?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobTerminationListenerService.java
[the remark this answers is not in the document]
mprimi: Why?
Both the gRPC listener and ctrl-c should use that component to shut down a launching or running job.

genie-proto/src/main/proto/genie.proto
mprimi: Which name? Listener?
Nothing else in the codebase is named Listener so don't tese messages should be named like this.
Same for Termination and Notification. They're not consistent with the rest of the naming in Java.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobTerminationListenerService.java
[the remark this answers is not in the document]
sumitnetflix: Ok will leave `KillService` around. Will later refactor code to use your component. Currently `KillServiceImpl` is not doing anything

genie-proto/src/main/proto/genie.proto
sumitnetflix: I meant sticking to the name - `JobKillRegistrationRequest\Response`

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcJobTerminationListenerServiceImplSpec.groovy
```diff
@@ -0,0 +1,192 @@
...
+        simulationCallback = new Runnable() {
+            @Override
+            void run() {}
+        }
+
+        when:
+        service.start(jobId)
+
+        then:
+        noExceptionThrown()
```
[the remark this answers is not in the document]
sumitnetflix: Tried it. Does not work with spock as client is not a Mock. Leaving as is.

genie-web/src/test/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImplUnitTests.java
mprimi: ⬆️

genie-web/src/main/java/com/netflix/genie/web/services/JobKillServiceV4.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Validated
+public interface JobKillServiceV4 {
+
+    /**
+     * Kill the job with the given id if possible.
+     *
+     * @param jobId     id of job to kill
+     * @param reason brief reason for requesting the job be killed
+     * @throws GenieException if there is an error
```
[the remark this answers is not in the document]
sumitnetflix: I think the details make more sense in the implementation `GRpcJobKillServiceImpl`. For the interface I think what we have is sufficient - Essentially that `GenieException` is thrown in case of an error. I can add more detailed description on `GRpcJobKillServiceImpl.killJob()` if that makes sense @mprimi ?

genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> true
+        1 * jobKillServiceV4.killJob(jobId, "testing")
+        0 * jobKillServiceV3.killJob(jobId, "testing")
+    }
+}
```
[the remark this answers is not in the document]
sumitnetflix: 👍

genie-web/src/main/java/com/netflix/genie/web/services/JobKillServiceV4.java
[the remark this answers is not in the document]
sumitnetflix: Added details on `GRpcJobKillServiceImpl.killJob()`

genie-web/src/test/groovy/com/netflix/genie/web/services/impl/JobKillServiceImplSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+        1 * jobKillServiceV4.killJob(jobId, "testing") >> { throw new GenieServerException("Error killing v4 job")}
+        thrown(GenieException)
+
+        when:
+        service.killJob(jobId, "testing")
+
+        then:
+        1 * jobPersistenceService.isV4(jobId) >> false
+        1 * jobKillServiceV3.killJob(jobId, "testing") >> { throw new GenieServerException("Error killing v3 job")}
+        thrown(GenieException)
```
mprimi: This could be stricter.
You could check that *exactly* the same exception, unchanged, is thrown.
Whereas you're just checking that an exception of a certain type is thrown.

sumitnetflix: Fixed

PR #753, opened 2018-06-19
Upgrade to Spring Boot 2.0.3

tgianos:
Also changes to remove conflicts with internal boot auto configurations from runtime team from some our beans and security configurations discovered during debugging session today

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/security/SecurityConfig.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/security/SecurityConfig.java
@@ -21,6 +21,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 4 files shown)


genie-web/src/main/java/com/netflix/genie/web/security/SecurityConfig.java
```diff
@@ -38,6 +39,7 @@
  */
 @Configuration
 @Slf4j
+@Order(357)
```
mprimi: Where is this magic number coming from?

tgianos: Good question. This class extends the default web security stuff and its only role is to disable all security which boot turns on by default in the situation where no other security is enabled. Before it didn't have an `@Order` on it so it was inheriting the super class one (100). If any other class in the classpath also tried to provide a default security configuration (internal security for example) it would conflict. So we just chose some random unlikely to be chosen number > 100 here to make sure it took effect after any default internal security configuration.

PR #759, opened 2018-07-03
Move `@Configuration` classes in `genie-web` module to be auto configured

tgianos:
This commit changes the `@Configuration` classes contained in the `genie-web` module to be Spring Boot Auto Configuration classes. This allows greater control for consumers of the jar to control behavior of these configurations (easier to override beans or ignore configurations completely) while simultaneously allowing us to control when the configuration is applied (e.g. after Spring Cloud AWS has configured the credentials bean). Integration with internal boot resources should be simpler.

To make this work even better we might want to explore moving away from classpath scanning for beans entirely but for now that's too big of a change and I'm committing this to unblock internal Boot 2 integration testing.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/configs/EventConfig.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/configs/EventConfig.java
@@ -19,6 +19,7 @@ package com.netflix.genie.web.configs;
 
 import com.netflix.genie.web.events.GenieEventBusImpl;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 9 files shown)


genie-web/src/main/java/com/netflix/genie/web/configs/EventConfig.java
```diff
@@ -42,6 +43,7 @@
      * @return The application event multicaster to use
      */
     @Bean
+    @ConditionalOnMissingBean(GenieEventBusImpl.class)
```
tcellucci: Not conditional on GenieEventBus, but specifically GenieEventBusImpl?

PR #762, opened 2018-07-11
Configuration flexibility and extensibility improvements

tgianos:
This PR contains various commits which attempt to improve the configuration of the Genie server side environment by leveraging Spring Boot Auto Configuration and removing static bean definitions via annotations.

This should allow easier modification of internal configuration.

All `genie` namespace properties should now have type safe property bindings as well


genie-client/src/test/resources/application-integration.yml
```diff
@@ -21,12 +21,13 @@ genie:
...
-      location: "file:///tmp/"
-    archive:
-      location: /archive/location
+    locations:
+      archives: /archive/location
+      jobs: file:///tmp/genie/jobs/
 
 spring:
+  autoconfigure:
+    exclude: org.springframework.boot.autoconfigure.data.redis.RedisAutoConfiguration, net.devh.springboot.autoconfigure.grpc.server.GrpcServerAutoConfiguration
```
mprimi: Could this be written as YAML array? Just to make it more legible..

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
```diff
@@ -40,7 +41,10 @@
 @Category(IntegrationTest.class)
 @RunWith(SpringRunner.class)
 @TestPropertySource(locations = "classpath:/PropertiesConfigIntegrationTest.properties")
-@SpringBootTest()
+@SpringBootTest(
+    classes = GenieTestApp.class,
+    properties = {"genie.grpc.server.enabled = false"}
```
mprimi: Why not put this in the YAML?

build.gradle
```diff
@@ -1,26 +1,21 @@
 buildscript {
```
mprimi: Nit: Commit title is slightly misleading, when doing a bisect lookin for a commit that broke something this could easily be overlooked

.editorconfig
```diff
@@ -10,6 +10,9 @@ indent_style = space
 [*.java]
 indent_size = 4
 
+[*.groovy]
+indent_size = 4
+
```
mprimi: We should switch to tabs... 

(JK)

genie-test-web/build.gradle
```diff
@@ -23,7 +25,9 @@ dependencies {
 }
 
 processResources {
-    filter ReplaceTokens, tokens: [
```
mprimi: This belongs in a different commit... no big deal but just in case you want to fix

genie-web/src/test/java/com/netflix/genie/web/configs/PropertyBindingIntegrationTests.java
```diff
@@ -38,14 +39,15 @@
  * @author mprimi
  * @since 3.1.0
  */
+@Ignore("This seems more like a test of Spring functionality than Genie functionality, do we need it?")
```
mprimi: To leave it disabled we might as well delete it...

genie-web/src/main/java/com/netflix/genie/web/configs/GenieApiAutoConfiguration.java
```diff
@@ -104,42 +111,37 @@ public GenieHostInfo genieHostInfo() throws UnknownHostException {
...
-    ) {
+    @Bean
+    @ConditionalOnMissingBean(name = "genieRestTemplate")
+    public RestTemplate genieRestTemplate(final HttpProperties httpProperties) {
         final HttpComponentsClientHttpRequestFactory factory = new HttpComponentsClientHttpRequestFactory();
-        factory.setConnectTimeout(httpConnectTimeout);
-        factory.setReadTimeout(httpReadTimeout);
+        factory.setConnectTimeout(httpProperties.getConnect().getTimeout());
+        factory.setReadTimeout(httpProperties.getRead().getTimeout());
         return new RestTemplate(factory);
```
twicksell: In Spring Boot you should inject RestTemplateBuilder instead of calling new RestTemplate. Otherwise you may miss out on customizations provided by other things, like metrics from Micrometer for example.

genie-web/src/main/java/com/netflix/genie/web/configs/GenieTasksAutoConfiguration.java
```diff
@@ -0,0 +1,233 @@
...
+@EnableConfigurationProperties(
+    {
+        ClusterCheckerProperties.class,
+        DatabaseCleanupProperties.class,
+        DiskCleanupProperties.class,
+        TasksExecutorPoolProperties.class,
+        TasksSchedulerPoolProperties.class
+    }
+)
+public class GenieTasksAutoConfiguration {
```
twicksell: Its often help to add a `@ConditionalOnProperty` to autoconfigs like this. Somewhere down the line you may want to disable scheduled tasks for a particular stack. You can always get around it by setting the `spring.autoconfigure.exclude ` property. Mostly a style difference, but something to consider.

genie-client/src/test/resources/application-integration.yml
tgianos: Yeah it looks like you can: https://github.com/spring-projects/spring-boot/commit/abfd139d8f24f45f8290a4a1fcdd0c649079bfe4

All the examples I'd found were like this so I copied it. I'll fix this in a subsequent commit

genie-web/src/test/java/com/netflix/genie/web/configs/PropertiesConfigIntegrationTest.java
tgianos: Cause right now the yaml defaults to having it on. I'm thinking of turning it off by default and having only the tests that need it opt in

.editorconfig
tgianos: haha

genie-web/src/test/java/com/netflix/genie/web/configs/PropertyBindingIntegrationTests.java
tgianos: Yeah I was going to delete it but I didn't want to do so without giving you a chance to comment on why it existed and if we should keep it around

genie-web/src/main/java/com/netflix/genie/web/configs/GenieApiAutoConfiguration.java
tgianos: Ok will do in subsequent commit thanks!

genie-web/src/main/java/com/netflix/genie/web/configs/GenieTasksAutoConfiguration.java
tgianos: There's still some interplay between the configs in so much as some beans in here are used in others. I'd like to detangle that web but probably not going to get to it right now

PR #766, opened 2018-07-12
Property binding improvements

tgianos:


genie-security/src/main/java/com/netflix/genie/security/oauth2/pingfederate/PingFederateSecurityConditions.java
```diff
@@ -40,11 +40,11 @@ public PingFederateRemoteEnabled() {
             super(ConfigurationPhase.PARSE_CONFIGURATION);
         }
 
-        @ConditionalOnProperty("genie.security.oauth2.enabled")
```
mprimi: The message for this commit is

genie-web/src/main/java/com/netflix/genie/web/properties/ZookeeperLeadershipProperties.java
```diff
@@ -28,11 +28,17 @@
...
 @Getter
 @Setter
 @Validated
 public class ZookeeperLeadershipProperties {
+
+    /**
+     * The property prefix for this group.
+     */
+    public static final String PROPERTY_PREFIX = "genie.zookeeper.leader";
+
```
mprimi: Moving a constant to a different place in the same file doesn't make that big of a difference.
How about moving all of these prefixes to a class?
At least by looking at that it would be easier to come up with ways to stay consistent on naming future properties.

tgianos: It makes a difference in that I can reference them in other places now as none strings, see usage of `genie.grpc.server.enabled` in like 4 classes. Keeping it in the property class allows me (at least in my mental map) to scope it. If I have to go to a constants file now I have to name them all with huge long strings and if I delete the properties file at some point remember to go delete the dangling constant reference. This seems more self contained to me.

genie-security/src/main/java/com/netflix/genie/security/oauth2/pingfederate/PingFederateSecurityConditions.java
tgianos: is what?

tgianos: oh incomplete haha

tgianos: I'll edit it

PR #770, opened 2018-07-28
Customize Agent logging configuration

mprimi:
Console: only log a very specific logger, intended for high visibility user messages.
log everything to file.


mprimi:
@tgianos @sumitnetflix this is ready for review

PR #771, opened 2018-08-07
Handle early termination in Agent

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/CleanupJobAction.java
```diff
@@ -48,6 +57,50 @@
...
+                    // No job final status (which reflect server-side status), and no sign of abort/kill.
+                    throw new IllegalStateException(
+                        "Reached cleanup state and finalJobState is null. Last job state: " + lastJobStatus
+                    );
+                }
+            } else {
+                log.debug("Job final status already updated server-side: {}", finalJobStatus);
+            }
+        } else {
+            log.debug("Job never claimed an ID, skipping server-side status update");
```
sumitnetflix: This can happen even if `lastJobStatus == null`. The log statement does not represent that

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/MonitorJobAction.java
```diff
@@ -62,10 +62,30 @@ protected Events executeStateAction(final ExecutionContext executionContext) {
...
         }
 
-        log.info("Job process completed with exit code: {}", exitCode);
+        try {
+            // Evil-but-necessary little hack.
+            // The agent and the child job process receive SIGINT at the same time.
+            // If the child terminates quickly, the code below will execute before the signal handler has a chance to
+            // set the job as killed, and the final status would be (incorrectly) reported as failed (due to non-zero
+            // exit code).
+            // So give the handler a chance to mark the context before attempting to read it.
```
sumitnetflix: Instead, why cant this code wait on a condition which will be fulfilled only when the job is marked as killed.

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
```diff
@@ -186,4 +187,33 @@ void addStateActionError(
...
     @Nullable
     String getClaimedJobId();
+
+    /**
+     * If the job was killed, return the provenance of the kill request.
+     *
+     * @return null if the job was not killed, or else the provenance of the kill signal
+     */
+    @Nullable
+    KillProvenance getJobKillProvenance();
```
tgianos: Would `KillSource` be simpler? Less to spell all over the place

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachine.java
```diff
@@ -38,4 +38,14 @@
...
      */
     States waitForStop() throws InterruptedException;
+
+    /**
+     * Force an early termination of the state machine, in response to user submitting a kill via API or ctrl-c.
+     * Notice:
+     *  - State actions are not interrupted. Only after the one currently executing is completed the stop
+     *    signal is processed (TODO: make long-running actions, such as setup, interruptible).
+     *  - Some actions are still performed before the program exits. For example publish the updated job status
+     *    server-side. Refer to the state machine configuration for details.
```
tgianos: Is there a `@see` or `@link` on this instead of just saying "refer" so we quickly know where to go?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
```diff
@@ -78,4 +78,13 @@ public States waitForStop() throws InterruptedException {
             throw e;
         }
     }
+
+    @Override
+    public void stop() {
+        log.info("Stopping state machine (in state: {})", stateMachine.getState().getId());
+        // This event is processed iff the state machine has not reached the state where the job is launched.
```
tgianos: `iff`

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
mprimi: I guess, but provenance sounds so much more pompous!

genie-agent/src/main/java/com/netflix/genie/agent/cli/ExecCommand.java
```diff
@@ -95,6 +100,16 @@ public void run() {
...
     }
 
+    @VisibleForTesting
+    void handleSigInt() {
+        UserConsole.getLogger().info(
+            "Received SIGINT, terminating job (status: {})",
+            executionContext.getCurrentJobStatus()
+        );
+        executionContext.setJobKilProvenance(ExecutionContext.KillProvenance.SYSTEM_SIGNAL);
+        stateMachine.stop();
```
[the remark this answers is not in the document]
mprimi: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachine.java
mprimi: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
mprimi: That was intentional (https://en.wikipedia.org/wiki/If_and_only_if)

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/CleanupJobAction.java
mprimi: It does.
Not having reached 'claimed' status is the important bit, it covers a range of states, including no state ever set.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/MonitorJobAction.java
mprimi: There's no guarantee that will happen, or happen in time.
So we could replace the `sleep(1000)` with `latch.wait(1000)` or similar. But in the end, the two are the same thing ("wait a little and hope the other thread does its thing").
Plus we have to share the latch between here and there.

sumitnetflix: `Thread.sleep` just seems arbitrary. With a rightly named latch and a timeout it becomes more obvious what one is doing. But you can argue that the comment you gave provides that context. I would make a call by thinking that if this is the only condition then having a sleep can be made to work. If there are more conditions in the future then having a latch would be more appropriate. For now its fine.

mprimi: Let's say we have a latch and we count it down in the signal handler.
And here we have `latch.wait(1000)`.

Case 1: job is killed: this code waits for the latch.
If it gets decremented in time, great. If not we timeout and miss the fact this was a kill.
But `sleep` provides the same guarantees.

Case 2: job is not killed.
This code waits for the latch, then times out (nobody is going to decrement it!) and moves on.
Once again, this is the same as what 'sleep' would have done.

The only benefit of a latch is waiting slightly less in case of kill (i.e. wake up and move on before timing out).

mprimi:
@sumitnetflix @tgianos any further comments?

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContextImpl.java
```diff
@@ -225,6 +227,24 @@ public void setClaimedJobId(@NotBlank final String jobId) {
         setIfNullOrTrow(jobId, claimedJobIdRef);
```
sumitnetflix: `setIfNullOrTrow` misspelled

mprimi: Ah thanks for spotting this.
I already have other commits in the pipe that modify this class, so I'll fix this as part of that, rather than in this PR.

PR #772, opened 2018-08-09
Handle v4 job kill request forwarding

sumitnetflix:
First draft for forwarding v4 job kill request to the right genie box. No tests added yet. Trying to get an early feedback on design.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/controllers/JobRestController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/controllers/JobRestController.java
@@ -27,8 +27,10 @@ import com.netflix.genie.common.dto.JobStatus;
 import com.netflix.genie.common.dto.JobStatusMessages;
 import com.netflix.genie.common.dto.search.JobSearchResult;
 import com.netflix.genie.common.exceptions.GenieException;
+import com.netflix.genie.common.exceptions.GenieNotFoundException;
 import com.netflix.genie.common.exceptions.GeniePreconditionException;
 import com.netflix.genie.common.exceptions.GenieServerException;
+import com.netflix.genie.common.internal.exceptions.unchecked.GenieJobNotFoundException;
 import com.netflix.genie.common.internal.jobs.JobConstants;
 import com.netflix.genie.common.internal.util.GenieHostInfo;
 import com.netflix.genie.web.hateoas.assemblers.ApplicationResourceAssembler;
@@ -49,8 +51,10 @@ import com.netflix.genie.web.hateoas.resources.JobResource;
 import com.netflix.genie.web.hateoas.resources.JobSearchResultResource;
 import com.netflix.genie.web.properties.JobsProperties;
 import com.netflix.genie.web.resources.handlers.GenieResourceHttpRequestHandler;
+import com.netflix.genie.web.services.AgentRoutingService;
 import com.netflix.genie.web.services.AttachmentService;
 import com.netflix.genie.web.services.JobCoordinatorService;
+import com.netflix.genie.web.services.JobPersistenceService;
 import com.netflix.genie.web.services.JobSearchService;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
@@ -96,6 +100,7 @@ import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.validation.Valid;
+import javax.validation.constraints.NotBlank;
 import java.io.IOException;
 import java.time.Instant;
 import java.util.EnumSet;
@@ -140,6 +145,8 @@ public class JobRestController {
     private final RestTemplate restTemplate;
     private final GenieResourceHttpRequestHandler resourceHttpRequestHandler;
     private final JobsProperties jobsProperties;
+    private final AgentRoutingService agentRoutingService;
+    private final JobPersistenceService jobPersistenceService;
 
     // Metrics
     private final Counter submitJobWithoutAttachmentsRate;
@@ -165,6 +172,8 @@ public class JobRestController {
      *                                         Genie File System.
      * @param jobsProperties                   All the properties associated with jobs
      * @param registry                         The metrics registry to use
+     * @param jobPersistenceService            Job persistence service
+     * @param agentRoutingService              Agent routing service
      */
     @Autowired
     @SuppressWarnings("checkstyle:parameternumber")
@@ -184,8 +193,10 @@ public class JobRestController {
         @Qualifier("genieRestTemplate") final RestTemplate restTemplate,
         final GenieResourceHttpRequestHandler resourceHttpRequestHandler,
         final JobsProperties jobsProperties,
-        final MeterRegistry registry
-    ) {
+        final MeterRegistry registry,
+        final JobPersistenceService jobPersistenceService,
+        final AgentRoutingService agentRoutingService
+        ) {
         this.jobCoordinatorService = jobCoordinatorService;
         this.jobSearchService = jobSearchService;
         this.attachmentService = attachmentService;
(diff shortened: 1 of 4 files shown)


mprimi:
Looks good so far, however it would be nice if this service can be reused to also route requests for logs. 
Does what you have in mind cover this?
If so, could you flesh it out?

sumitnetflix:
@mprimi Are you talking about a service like `JobOwnerRoutingService` with a method like `routeToOwner(jobId)` which will do the actual forwarding. This method can be used by any other service(`log`, `job kill`) to do the forwarding? Currently what I had in mind was a service which will just return the job owner hostname whether v3 or v4.

sumitnetflix:
?? Not clear to me why will it not work. The end point which serves log would use `JobOwnerService` like `JobRestController.killJob()`. Are you looking for that change?

mprimi:
Oh, you're right, it would work, assuming we keep the current plan for log syncing

genie-web/src/main/java/com/netflix/genie/web/services/JobOwnerService.java
```diff
@@ -0,0 +1,44 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: Copyright is wrong

genie-web/src/main/java/com/netflix/genie/web/services/JobOwnerService.java
```diff
@@ -0,0 +1,44 @@
...
+public interface JobOwnerService {
+
+    /**
+     * Get the hostName owning the job.
+     *
+     * @param jobId job id
+     * @return hostname of the job owner
+     * @throws GenieException if there is an error finding the host owning the job
+     */
+    String getJobOwnerHostName(
```
tgianos: `Hostname` not `HostName`

genie-web/src/main/java/com/netflix/genie/web/services/JobOwnerService.java
```diff
@@ -0,0 +1,44 @@
...
+        @NotBlank(message = "No job id entered. Unable to find the owner.") final String jobId
+    ) throws GenieException;
```
tgianos: This is that slippery slope to generic exception throwing.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobOwnerServiceImpl.java
```diff
@@ -0,0 +1,91 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: copyright. You should change your IDE template for java files

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobOwnerServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+ * Service to get information about the server owning a job.
+ * Job can be a v3 job which runs locally on a genie node. This node is the job owner.
+ * Job can be a v4 job running remotely on a agent. The genie node which has the
+ * connection to this agent is the owner.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Slf4j
+@Service
```
tgianos: This should be defined in `GenieServicesAutoConfiguration` with a `ConditionalOnMissingBean` like the rest of our services so it can be easily overridden.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobOwnerServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+@Lazy
```
tgianos: We don't have lazy beans on the server side. They're (currently) present on the agent side in an effort to speed up startup and avoid initializing anything that won't be needed. As the server side is long lived and all services will be needed at some point we can take the startup hit. This shouldn't be here.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobOwnerServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+public class JobOwnerServiceImpl implements JobOwnerService {
+
+    private final AgentRoutingService agentRoutingService;
+    private final JobSearchService jobSearchService;
+    private final JobPersistenceService jobPersistenceService;
+
+    /**
+     * Constructor.
+     *
+     * @param agentRoutingService       Service to routing information for jobs running on agent(v4 jobs)
```
tgianos: missing a space between `agent(v4`. I think you also mean `Service for routing`. Though I'm not sure the difference between `routing` and `owning`

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobOwnerServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+        final JobSearchService jobSearchService,
+        final JobPersistenceService jobPersistenceService
+
+        ) {
+        this.agentRoutingService = agentRoutingService;
+        this.jobSearchService = jobSearchService;
+        this.jobPersistenceService = jobPersistenceService;
+    }
+
+    @Override
```
tgianos: the `@Override` should be directly above the code not the comment. Also no need to duplicate the comment as it should just inherit the one from the interface using `{@inherit}` and then anything additional you can put above or below that.

sumitnetflix:
@tgianos The semantics of the `JobOwnerService` in my head are to tell which server "owns" the job. For v3 ownership means which server spawned the job. For v4, it means which server has the active connection to the agent. `AgentRoutingService` in my head is the one which is responsible for keeping track of which server owns the active connection to an agent. Its like a routing table which one can query to get information on how to "reach" an agent(via the active connection). Hence `JobOwnerService` uses it and `JobSearchService` to answer the question which server owns the job for v4 and v3 respectively. `JobKillService` then uses  `JobOwnerService` to get the owner instead of `JobSearchService` and does the request forwarding as per its need. Any other service, like the one which is trying to get the logs for a job, will use `JobOwnerService` to get the owner and do the request forwarding as per its need. Since each service might have different needs for doing forwarding(like which headers to look at) we(I and @mprimi) don't think that there is much value in having a service to do the forwarding. If the need arises later, we can refactor.

To answer your question
"If logs don't exist on any node will this service return a null/empty for which node owns a job?"
`JobOwnerService` is independent from the kind of request(`get log`, 'kill job` etc). So, it will be up to log service to decide if it needs the job owner or not. For e.g if logs are implemented such  that for v4 the logs are on the server with the active connection(owner) then log service will use `JobOwnerService` else it will not.

This PR was a quick and dirty thing I put up to get early feedback. I have left out answering few of your comments which are just a manifestation of that. 

Thanks for the comments.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobOwnerServiceImpl.java
sumitnetflix: Makes sense. Will fix

genie-web/src/main/java/com/netflix/genie/web/services/JobOwnerService.java
sumitnetflix: I would like to throw `GenieNotFoundException` here. Here is my problem. In this method I am using `jobSearchService.getJobHost`. It throws a `GenieException`. Internally its only throwing a `GenieNotFoundException`. So, either we change the `jobSearchService.getJobHost` to throw `GenieNotFoundException` or I catch `GenieException` in this method and convert it to throw `GenieNotFoundException`. Thoughts ?

sumitnetflix:
@tgianos Had a discussion with @mprimi about `JobOwnerService`. We think that this service is not doing much right now. Its just doing a switch based on a if statement to get the job owner. Its not very clear whether this will be reusable in the `getLogs` piece. So, to keep things simple I am going to add the method `getJobOwnerHostname` in `JobRestController` as a helper and get rid of the `JobOwnerService`. Later when the implementation of getting logs for v4 is more clear we can refactor.

PR #774, opened 2018-08-15
Agent routing changes listener

sumitnetflix:


genie-web/src/main/java/com/netflix/genie/web/services/AgentRoutingChangesListener.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.validation.constraints.NotBlank;
+
+/**
+ * Interface to listen for change in agent routing information.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Validated
+public interface AgentRoutingChangesListener {
```
mprimi: `AgentConnectionListener`/`AgentConnectionObserver`?

genie-web/src/main/java/com/netflix/genie/web/services/AgentRoutingService.java
```diff
@@ -63,4 +64,17 @@
      */
     void handleClientDisconnected(@NotBlank String jobId);
 
+    /**
```
mprimi: What about letting Spring do this, rather than having services explicitly register/unregister?

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobKillServiceImpl.java
```diff
@@ -58,9 +62,10 @@
     /**
      * Constructor.
      *
-     * @param jobSearchService Job search service
+     * @param jobSearchService      Job search service
      */
-    public GRpcJobKillServiceImpl(final JobSearchService jobSearchService) {
+    public GRpcJobKillServiceImpl(final JobSearchService jobSearchService
+    ) {
```
sumitnetflix: typo .. will fix `)` on the next line

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobKillServiceImpl.java
```diff
@@ -110,4 +111,20 @@ public void killJob(final @NotBlank(message = "No job id entered. Unable to kill
...
+    public void onConnected(@NotNull final String jobId) {
+        //No op
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onDisconnected(@NotNull final String jobId) {
+        parkedJobKillResponseObservers.remove(jobId);
```
mprimi: Possible race condition here.
Agent 1 disconnects, reconnects to the same server.
Then the server processes the first onDisconnect.
The second connection (still valid) will be removed from the map.

No suggestions on how to address this... I haven't looked at details and how we could make this better.
But you should probably address this scenario.

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentRoutingServiceImpl.java
```diff
@@ -87,5 +104,63 @@ public void handleClientConnected(@NotBlank final String jobId) {
     public void handleClientDisconnected(@NotBlank final String jobId) {
         log.info("Agent executing job {} disconnected", jobId);
         this.agentConnectionPersistenceService.removeAgentConnection(jobId, genieHostInfo.getHostname());
+
+        for (AgentConnectionObserver agentConnectionObserver : agentConnectionObservers) {
+            this.connectionObserversNotificationExecutor.execute(
```
mprimi: This could throw (ah, the beauty of unchecked exceptions!).
I think the exception can propagate all the way back to a gRPC thread, which might permanently disable or block the server.

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentRoutingServiceImpl.java
```diff
@@ -87,5 +104,63 @@ public void handleClientConnected(@NotBlank final String jobId) {
...
+                new AgentDisconnectedNotificationTask(agentConnectionObserver, jobId)
+            );
+        }
+    }
+
+    /**
+     * Task to deliver agent connected notification to a AgentConnectionObserver.
+     */
+    static class AgentConnectedNotificationTask implements Runnable {
```
mprimi: Use a lambda or method reference?

These 2 classes are 50 LOC (and object being created and destroyed at runtime) for two simple method calls...

genie-web/src/main/java/com/netflix/genie/web/services/AgentConnectionObserver.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.validation.constraints.NotNull;
+
+/**
+ * Interface to observe the connection status of an agent to the local genie server.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Validated
+public interface AgentConnectionObserver {
```
mprimi: Have you noticed the relationship between `GrpcHeartBeatServiceImpl` and `AgentRoutingService`?

It is essentially the same as this interface.
Maybe there is some refactoring that simplifies this whole thing.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GRpcJobKillServiceImpl.java
sumitnetflix: First thing that comes to my mind is that essentially we need some kind of ordering - a `disconnect` event should remove only the response observer which was created in the map before the `disconnect` event. So the map of response observers can also have a `creation timestamp`. That can be checked before removing. We should possibly whiteboard this.

genie-web/src/main/java/com/netflix/genie/web/configs/GenieServicesAutoConfiguration.java
```diff
@@ -408,4 +418,19 @@ public JobCompletionService jobCompletionService(
...
         );
     }
+
+    /**
+     * Get a task executor which may be shared by different components.
+     *
+     * @return A task executor
+     */
+    @Bean
+    @Lazy
```
tgianos: I'm pretty sure we already have a `TaskExecutor` in `TaskAutoConfig` or somewhere. Why do we need another one? Also it shouldn't be lazy on the server side.

Looks like this was a copy and paste from the agent.

PR #776, opened 2018-08-23
Add clean up options for the job folder

sumitnetflix:
Adding clean up options for handling the job folder on job completion. 
--no-cleanup - leaves the job folder as is
--full-cleanup - deletes the job folder
Default behavior - Delete the dependencies in the application, cluster and command folders

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
@@ -1,0 +1,52 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.cli;
+
+import com.beust.jcommander.Parameter;
+import org.springframework.stereotype.Component;
+
+/**
+ * Implementation of AgentOptions delegate.
+ */
+@Component
+public class AgentOptionsArgumentsImpl implements ArgumentDelegates.AgentOptions {
+
+    @Parameter(
+        names = {"--no-cleanup"},
+        description = "Don't remove the job folder"
+    )
+    private boolean noCleanup;
+
+    @Parameter(
+        names = {"--full-cleanup"},
+        description = "Remove the job folder"
+    )
+    private boolean fullCleanup;
+
+    @Override
+    public JobFolderCleanupOption getJobFolderCleanUpOption() {
+        if (noCleanup) {
+            return JobFolderCleanupOption.NO_CLEANUP;
+        } else if (fullCleanup) {
+            return JobFolderCleanupOption.DELETE_JOB_FOLDER;
+        } else {
+            return JobFolderCleanupOption.DELETE_DEPENDENCIES_ONLY;
+        }
+    }
+}
--- a/genie-agent/src/main/java/com/netflix/genie/agent/cli/JobFolderCleanupOption.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/cli/JobFolderCleanupOption.java
@@ -1,0 +1,40 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.agent.cli;
+
+/**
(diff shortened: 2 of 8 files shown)


genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+
+package com.netflix.genie.agent.cli;
+
+import com.beust.jcommander.Parameter;
+import org.springframework.stereotype.Component;
+
+/**
+ * Implementation of AgentOptions delegate.
```
tgianos: Would prefer a `{@link AgentOptions}` here so I can quickly click that within the IDE to go to the class. Also in the actual java docs when they're rendered it will be quicker to navigate.

Should add `author` and `since` tags as well.

genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+@Component
+public class AgentOptionsArgumentsImpl implements ArgumentDelegates.AgentOptions {
+
+    @Parameter(
+        names = {"--no-cleanup"},
+        description = "Don't remove the job folder"
```
tgianos: I would say something more like "Don't delete anything from disk in the job folder upon job completion". Cause aren't there two options that don't remove the job folder? This and the default?

genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    )
+    private boolean noCleanup;
+
+    @Parameter(
+        names = {"--full-cleanup"},
+        description = "Remove the job folder"
+    )
+    private boolean fullCleanup;
```
tgianos: Is there a way to describe the default behavior on the command line when a user types help?

genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Parameter(
+        names = {"--no-cleanup"},
+        description = "Don't remove the job folder"
+    )
+    private boolean noCleanup;
+
+    @Parameter(
+        names = {"--full-cleanup"},
+        description = "Remove the job folder"
```
tgianos: "Delete the entire job folder on upon job completion from disk"

genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    )
+    private boolean fullCleanup;
+
+    @Override
```
tgianos: Missing javadoc here. surprised compiler isn't complaining. Did we turn that off in the agent compiler arguments and forget to turn it back on? (@mprimi)

genie-agent/src/main/java/com/netflix/genie/agent/cli/AgentOptionsArgumentsImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    public JobFolderCleanupOption getJobFolderCleanUpOption() {
+        if (noCleanup) {
```
tgianos: Would prefer if you reference member variables with `this.` to make it clear where they're coming from. In this class it's pretty simple but as the classes get larger it can get more confusing.

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobFolderCleanupOption.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Clean up options for handling the job folder on job completion.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public enum JobFolderCleanupOption {
+    /**
+     * Delete the entire job folder.
```
tgianos: We probably need to make sure it's clear this means only on local disk as some contents of the job directory should have already been moved from the agent to the service for backup.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -119,40 +125,57 @@ protected Events executeStateAction(
...
-                .map(PathUtils::jobEntityDependenciesPath)
-                .map(Path::toString)
-                .map(pathString -> pathString + "/.*")
-                .map(Pattern::compile)
-                .toArray(Pattern[]::new)
-        );
+        //No clean up needed
+        if (cleanupOption == JobFolderCleanupOption.NO_CLEANUP) {
+            log.info("Job folder not deleted.");
+        } else if (cleanupOption == JobFolderCleanupOption.DELETE_JOB_FOLDER) { //Delete the job folder
```
tgianos: Comment seems pretty redundant given the names of the enumeration values

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -119,40 +125,57 @@ protected Events executeStateAction(
...
+            try {
+                log.debug("Deleting {}", jobDirectory);
+                FileSystemUtils.deleteRecursively(jobDirectoryPath);
+                log.info("Deleted job folder.");
```
tgianos: Why not capture the job directory path here too?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -119,40 +125,57 @@ protected Events executeStateAction(
...
+            } catch (final IOException e) {
+                log.warn("Failed to delete job folder: {}", jobDirectoryPath.toAbsolutePath().toString(), e);
+            }
+        } else { //Remove only the dependencies
+
+            // TODO: this is a simple and safe cleanup strategy, 3 patterns are added that match any files in the
```
tgianos: You have this as a TODO. Do you intend to change it or leave it as is. If you're leaving it then I'd remove the TODO

PR #778, opened 2018-09-07
Archive job folder from agent when job finishes

sumitnetflix:


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobFolderArchiveServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Default implementation does not archive.
+     *
+     * @param executionContext  Execution context
+     * @throws ArchiveException In case of a problem
+     */
+    @Override
+    public void archiveJobFolder(final ExecutionContext executionContext) throws ArchiveException {
+        log.info("Job folder not archived. Please provide an implementation if you wish to archive.");
+    }
```
sumitnetflix: The way I am thinking about this is that this default implementation will not do any archival. There will be a s3 based implementation provided for Netflix use case which will provide a S3 URL to the uploadService. UploadService will then compress and upload it to the S3 location

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobFolderArchiveService.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public interface JobFolderArchiveService {
+
+    /**
+     * Archive the job folder.
+     *
+     * @param executionContext Execution context
+     * @throws ArchiveException if archival fails
+     */
+    void archiveJobFolder(ExecutionContext executionContext) throws ArchiveException;
```
mprimi: Services should not be coupled with execution context. You should pass the relevant bit of the context (source folder, destination URL, etc) rather than the whole thing.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/UploadService.java
```diff
@@ -0,0 +1,42 @@
...
+public interface UploadService {
+
+    /**
+     * Compress the file and upload to a target location.
+     *
+     * @param file                 file to upload
+     * @param targetLocationURI    URI representing the target location for the file
+     * @throws UploadException     if the upload fails
+     */
+    void compressAndUpload(final File file, final URI targetLocationURI) throws UploadException;
```
mprimi: API dissonates with name. Why not make 1 service to do 1 thing (upload in this case)?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobFolderArchiveService.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+package com.netflix.genie.agent.execution.services;
+
+import com.netflix.genie.agent.execution.ExecutionContext;
+import com.netflix.genie.agent.execution.exceptions.ArchiveException;
+
+/**
+ * Service for archiving the job folder.
```
mprimi: Neither comments explain what does it mean to "archive".
Create an archive?
Upload to an archive location?
Both?

Is this specific to the job folder? If not, this service could be generic and have a generic name.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/UploadServiceImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+        zipFile(fileToZip, nameInArchive, zipOutputStream);
+
+        if (fileToZip.isDirectory()) {
+            final File[] children = fileToZip.listFiles();
+            //Making findbugs happy. No need for this ideally as fileToZip is a directory and listFiles()
+            //is guaranteed not be null
+            if (children != null) {
+                for (File childFile : children) {
+                    zipDirectory(
```
mprimi: A bit confusing that you call `zipDirectory()` for regular files (non-directories).
Could find a better name for this helper.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/UploadServiceImpl.java
```diff
@@ -0,0 +1,145 @@
...
+     * Zip a directory recursively.
+     *
+     * @param fileToZip         file/dir to zip
+     * @param nameInArchive     name for the file entry in archive
+     * @param zipOutputStream   destination output stream
+     * @throws IOException      in case of a problem
+     */
+    private void zipDirectory(
+        @NotNull final File fileToZip,
+        @NotNull final String nameInArchive,
```
mprimi: `nameInArchive` is the relative path w.r.t. the root of the job directory, no?
Why not name it as such?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobFolderArchiveService.java
sumitnetflix: Thought I had was - 
-`UploadService` is a service to upload a file/dir to a target location. This service would allow a method to compress and upload. 
-`JobFolderArchiveService` is specific for job folder archival. It will have one implementation for S3 which will know the s3 url based on the env (prod/test) where to upload the job folder. It will have further checks like if the job folder is missing(in case agent fails before a jobFolder is created), it will not use upload service to archive. That way everything related to JobFolderArchival stays in one place
- `ShutDownAction` will use this `JobFolderArchiveService` to do an upload.

We can possibly have another service `CompressService` which will compress a file/dir and return a zipped file. `UploadService` will then only upload the file provided.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/UploadServiceImpl.java
sumitnetflix: Makes sense. Will fix the name

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/UploadService.java
sumitnetflix: Can break it into `CompressService` and `UploadService` with only `upload()` method

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/UploadServiceImpl.java
sumitnetflix: Sure

sumitnetflix: On giving it a thought, I think this method signature can overall be improved. `nameInArchive/relativePathToRoot` is derivable. The method should be something like `zipRecursively(File fileToZip, File root, Outputstream os)`. I will work on making this cleaner

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentDelegates.java
```diff
@@ -65,7 +65,7 @@
 
         boolean isInteractive();
 
-        boolean isArchivalDisabled();
+        String getArchiveLocation();
```
mprimi: Separate commit?
Doesn't need to be with the S3 implementation of archival.

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestConverter.java
```diff
@@ -92,7 +92,7 @@ public AgentJobRequest agentJobRequestArgsToDTO(
             .Builder()
             .withRequestedJobDirectoryLocation(jobRequestArguments.getJobDirectoryLocation())
             .withTimeoutRequested(jobRequestArguments.getTimeout())
-            .withArchivingDisabled(jobRequestArguments.isArchivalDisabled())
+            .withArchiveLocation(jobRequestArguments.getArchiveLocation())
```
mprimi: This commit seems to be missing quite a few things.. maybe it was not meant for review yet?

sumitnetflix: The persistence layer change that you were looking for @mprimi, is already there in the file  - genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java

sumitnetflix: There is a bug which I identified.. I am currently storing only the archiveLocation provided by the user from cli which is really the s3 prefix and not the actual archiveLocation with the jobFolder name appended. I will work on fixing that. Will request a review after that.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
```diff
@@ -0,0 +1,42 @@
...
+public interface ArchivalService {
+
+    /**
+     * Archive file to a target location.
+     *
+     * @param file              file/dir to archive
+     * @param targetLocationURI target location uri for the file
+     * @throws ArchivalException if archival fails
+     */
+    void archive(final File file, final URI targetLocationURI) throws ArchivalException;
```
tgianos: As I mentioned in @mprimi review for #784 I think we should be targeting using `java.nio.*` classes going forward instead of the older `java.io.*` classes where possible.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+     * target location - s3://$bucketName/run
+     * <p>
+     * following structure is created in S3
+     * s3://$bucketName/abc -> contents of run file
+     *
+     * @param file              file/directory to archive.
+     * @param targetLocationURI target location uri for the file
+     * @throws ArchivalException if archival fails
+     */
+
```
tgianos: Extra line here

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+            log.info("Error archiving to target location: {} ", targetLocationURI);
+            throw new ArchivalException("Error archiving file: " + file.getName(), e);
+        }
+    }
+
+    /**
+     * Traverse a directory recursively and archive the files and directories
+     * underneath it. If the directory is empty, create an empty object in S3.
+     * For a regular child file create an object in S3 and copy the file contents.
+     * For a empty child directory create an empty object in S3
```
tgianos: Why bother creating an empty object in S3 for an empty directory?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    ) {
+        if (!fileToArchive.isDirectory()) {
+            uploadFile(targetLocationURI.getBucket(), targetLocationURI.getKey(), fileToArchive);
+        } else {
+            createEmptyFolder(targetLocationURI.getBucket(), targetLocationURI.getKey());
+        }
+    }
+
+    //Upload a file to the given bucket and key.
+    private void uploadFile(final String bucketName, final String key, final File file) {
```
tgianos: This seems like a useless one line method? Was there other logic you intended to do in here rather than just call the `s3` api directly

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        @NotNull final AmazonS3URI targetLocationURI
+    ) {
+        if (!fileToArchive.isDirectory()) {
+            uploadFile(targetLocationURI.getBucket(), targetLocationURI.getKey(), fileToArchive);
+        } else {
+            createEmptyFolder(targetLocationURI.getBucket(), targetLocationURI.getKey());
+        }
+    }
+
+    //Upload a file to the given bucket and key.
```
tgianos: Why is comment not in javadoc format? I know it's a private method and it doesn't really matter but for consistency it seems like the right thing to do

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    private void uploadFile(final String bucketName, final String key, final File file) {
+        amazonS3.putObject(new PutObjectRequest(bucketName, key, file));
+    }
+
+    //Create an object representing an empty folder. Key ends with File.separator.
+    //Content of the object is empty
+    private void createEmptyFolder(final String bucketName, final String key) {
```
tgianos: I'm not sure I understand the purpose of this method. S3 has no concept of `folder` so there's no real need to create one. You should just be able to keep uploading files and the prefixes will create "folders".

It seems like maybe you're trying to mimic rsync here so that the entire structure matches exactly. I wonder if that is necessary. I'm not sure if there is value in representing empty directories on a read of this archive location. Just wasted time/API calls. @mprimi thoughts?

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestConverter.java
```diff
@@ -92,7 +92,7 @@ public AgentJobRequest agentJobRequestArgsToDTO(
             .Builder()
             .withRequestedJobDirectoryLocation(jobRequestArguments.getJobDirectoryLocation())
             .withTimeoutRequested(jobRequestArguments.getTimeout())
-            .withArchivingDisabled(jobRequestArguments.isArchivalDisabled())
+            .withRequestedArchiveLocationPrefix(jobRequestArguments.getArchiveLocationPrefix())
```
tgianos: I like the name `requestedArchiveLocation` better but that's just me

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/AgentConfigRequest.java
```diff
@@ -40,6 +40,7 @@
 @SuppressWarnings("checkstyle:finalclass")
 public class AgentConfigRequest {
     private final boolean archivingDisabled;
+    private final String requestedArchiveLocationPrefix;
```
tgianos: I'm not sure that this is the right class for this data. I understand that the `archivingDisabled` flag is in here and it is distinctly possible that that is in the wrong place as well. It was dumped somewhere temporarily for v3 compatibility.

This is metadata provided by the user which impacts the job and behavior of the job within the overall ecosystem. I might feel that `JobMetadata` is a better place or even a new class entirely like `JobArchiveMetadata` or something. @mprimi thoughts?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/JobSpecification.java
```diff
@@ -90,6 +93,9 @@ public JobSpecification(
             : ImmutableMap.copyOf(environmentVariables);
         this.interactive = interactive;
         this.jobDirectoryLocation = jobDirectoryLocation;
+        this.archiveLocation = StringUtils.isBlank(archiveLocation)
```
tgianos: I'm not sure I understand the value of saving an empty string vs just saving the null? Also does a null represent that archiving is disabled?

genie-web/src/main/java/com/netflix/genie/web/jpa/entities/JobEntity.java
```diff
@@ -299,6 +300,11 @@
     @Size(max = 1024, message = "Max length in database is 1024 characters")
     private String archiveLocation;
 
+    @Basic
```
tgianos: these should be marked `optional = true` and `nullable = true`

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobSpecificationServiceImpl.java
```diff
@@ -520,4 +529,26 @@ private String tagsToString(final Set<String> tags) {
...
+     */
+    private String toArchiveLocation(
+        @NotNull final String requestedArchiveLocationPrefix,
+        @NotNull final String jobId
+    ) {
+
+        String archiveLocationPrefix = requestedArchiveLocationPrefix.trim();
+
+        //Remove the last File.seperator if the user supplied it
+        if (archiveLocationPrefix.endsWith(File.separator)) {
```
tgianos: This seems unnecessary. If the user supplied a `/` at the end why strip it off only to add it again in the `return` statement.

Just have your `if` append the `jobId` if it's there else append both.

genie-web/src/main/resources/db/migration/mysql/V4_0_0__Genie_4.sql
```diff
@@ -165,7 +165,8 @@ ALTER TABLE `jobs`
   ADD COLUMN    `agent_version`                              VARCHAR(255)  DEFAULT NULL,
   ADD COLUMN    `agent_pid`                                  INT(11)       DEFAULT NULL,
   ADD COLUMN    `claimed`                                    BOOLEAN       DEFAULT FALSE NOT NULL,
-  ADD COLUMN    `v4`                                         BOOLEAN       DEFAULT FALSE NOT NULL;
+  ADD COLUMN    `v4`                                         BOOLEAN       DEFAULT FALSE NOT NULL,
+  ADD COLUMN    `requested_archive_location_prefix`          VARCHAR(1024) DEFAULT NULL;
```
tgianos: Did you test all three of these database scripts to verify the final structure?

genie-web/src/main/resources/db/migration/postgresql/V4_0_0__Genie_4.sql
```diff
@@ -150,7 +150,9 @@ ALTER TABLE jobs
   ADD COLUMN agent_version                    VARCHAR(255)  DEFAULT NULL,
   ADD COLUMN agent_pid                        INT           DEFAULT NULL,
   ADD COLUMN claimed                          BOOLEAN       DEFAULT FALSE NOT NULL,
-  ADD COLUMN v4                               BOOLEAN       DEFAULT FALSE NOT NULL;
+  ADD COLUMN v4                               BOOLEAN       DEFAULT FALSE NOT NULL,
+  ADD COLUMN requested_archive_location_prefix VARCHAR(1024) DEFAULT NULL;
```
tgianos: Since your column here is so long (I still recommend removing `_prefix`) you should fix the formatting so it all lines up in the other rows.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java
```diff
@@ -45,7 +58,25 @@
...
     protected Events executeStateAction(final ExecutionContext executionContext) {
         log.info("Shutting down...");
-        //TODO implement this action
+
+        try {
+            log.info("Attempting to archive job folder...");
+
+            final JobSpecification jobSpecification = executionContext.getJobSpecification();
+            final String archiveLocation = jobSpecification.getArchiveLocation();
+            if (!StringUtils.isBlank(archiveLocation)) {
```
tgianos: Theres a `isNotBlank` method in `StringUtils` would help readability instead of using the `!` operator here

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java
```diff
@@ -45,7 +58,25 @@
     @Override
     protected Events executeStateAction(final ExecutionContext executionContext) {
         log.info("Shutting down...");
-        //TODO implement this action
+
+        try {
+            log.info("Attempting to archive job folder...");
```
tgianos: Would recommend logging this information after you know the location and including the location in the message something like `Attempting to backup job folder to {location}...`

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
```diff
@@ -64,10 +64,10 @@
     private boolean interactive;
 
     @Parameter(
-        names = {"--archivalDisabled"},
-        description = "Disable server-side archival of logs (does not disable logs upload and short-term retrieval)"
+        names = {"--archiveLocationPrefix"},
```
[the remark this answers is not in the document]
mprimi: Another possibility is to make this explicitly S3 specific, since that's what we support anyway right now.
i.e. `--s3ArchivalPrefix`.

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
```diff
@@ -64,10 +64,10 @@
...
+        description = "Prefix for job folder archival location(Currently only s3 URIs are supported)"
     )
-    private boolean archivalDisabled;
+    private String archiveLocationPrefix;
```
mprimi: Consider adding a validator and a converter to this option.
Validator rejects invalid values during parsing, rather than getting surprised by a bad URI at runtime.
Converter can be used to make this parameter a URI or whatever else makes sense. Rather than converting to URI every time you use it.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.netflix.genie.agent.execution.services;
+
+import com.netflix.genie.agent.execution.exceptions.ArchivalException;
+
+import java.io.File;
+import java.net.URI;
+
+/**
+ * Service to archive a local file to a target location URI.
```
mprimi: Description makes it sound like it uploads a single file.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+public interface ArchivalService {
+
+    /**
+     * Archive file to a target location.
```
mprimi: Same as above, seems like it copies a single file to the target URI

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @param file              file/dir to archive
+     * @param targetLocationURI target location uri for the file
```
mprimi: `targetURI`, `destinationURI`? i.e. `Location` just seems redundant

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+class S3ArchivalServiceImpl implements ArchivalService {
+
+    private final AmazonS3 amazonS3;
+
+    S3ArchivalServiceImpl(final AmazonS3 amazonS3) {
+        this.amazonS3 = amazonS3;
+    }
+
+    /**
+     * Store the contents of a file to S3.
```
mprimi: Same comment. Makes it sound like it archives a single file.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
mprimi: Maybe this service should explicitly require the first argument to be a folder?
I know the current implementation works for both regular and directory, but would make it simpler to say it archives a directory recursively to a remote location.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+     * <p>
+     * Example 1
+     * <p>
+     * For a dir hierarchy
+     * /tmp/${jobid}/genie/applications (empty dir)
+     * /tmp/${jobid}/genie/run (file)
+     * /tmp/${jobid}/genie/commands/presto (file)
+     * <p>
+     * dir to be archived - /tmp/${jobid}
+     * target location - s3://$bucketName/{$jobid}
```
mprimi: I think adding `${jobId} in this example is just confusing, since the service itself does nothing with it.
For file `x` uploaded to `/foo/bar`, this will create `/foo/bar/x`.
Bucket is also presented as a variable for some reason, but it's really not, as far as this service is concerned.

If you really want to keep these variables. Be consistent with syntax:
`$foo`, `{$foo}`, `${foo}`

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
mprimi: Allow to list vs. throw 404.

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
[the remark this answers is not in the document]
sumitnetflix: The problem is that there is already a field called `archiveLocation` in the db. This will store the actual archiveLocation. For e.g if the user passes in `s3://bucket/base` as the `requestedArchiveLocationPrefix`(s3 uri is just an example), then archiveLocation will store `s3://bucket/base/{jobid}`. Now, if I store `s3://bucket/base` in `requestedArchiveLocation` and `s3://bucket/base/{jobid}` in `archiveLocation` then the meaning of archiveLocation changes. One is really just the root uri and the other is the location of the job folder as it has `{jobId}` in the uri. By using `prefix` I was trying to make that clear. Hence I decided to call the field in the db `requestedArchiveLocationPrefix` and thinking backwards from there called the cli parameter with the same name. If `prefix` sounds s3 specific I can call the cli parameter `requestArchiveLocationRoot`.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+     * Traverse a directory recursively and archive the files and directories
+     * underneath it. If the directory is empty, create an empty object in S3.
+     * For a regular child file create an object in S3 and copy the file contents.
+     * For a empty child directory create an empty object in S3
+     *
+     * @param fileToArchive     file/dir to archive.
+     * @param targetLocationURI target location URI for storing the file in S3.
+     */
+    private void archiveRecursively(
+        @NotNull final File fileToArchive,
```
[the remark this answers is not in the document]
sumitnetflix: Let me check

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
sumitnetflix: Will do

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
sumitnetflix: No other logic. Will remove the method

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
sumitnetflix: For backward compatibility. Today on the v3 api when one is browsing the output folder, empty directories are shown. If we don't store empty directories in S3 they cannot be shown.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        @NotNull final AmazonS3URI targetLocationURI
+    ) {
+        final File[] children = fileToArchive.listFiles();
+
+        //children == null if fileToArchive is not a directory.
+        //children is a empty array if the directory is empty
+        if (children == null || children.length == 0) {
```
mprimi: This could be simply:
```
if (children == null) {
  uploadFile
} else {
  createEmptyFolder
  for c in children {
    archiveRecursively
  }
}
```
Combined with Tom's comment below, this should shorten this file a bit and make it easier to test.

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImplSpec.groovy
```diff
@@ -0,0 +1,205 @@
...
+    void cleanup() {
+    }
+
+    def "Archive a sample job folder. S3 objects creation calls called for files and empty directories with correct keys"() {
+
+        when:
+        s3ArchivalService.archive(jobDir, baseLocationS3URI.getURI())
+
+        then:
+        7 * amazonS3.putObject(_ as PutObjectRequest) >> {
```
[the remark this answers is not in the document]
mprimi: Spock manual has a section about order of interaction.
(and you could easily make your traversal order deterministic otherwise)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
sumitnetflix: Ok, will remove empty folder archival then.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/JobSpecification.java
sumitnetflix: Will do

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java
```diff
@@ -183,7 +183,8 @@ public JobSpecification toJobSpecificationDTO(
                 .collect(Collectors.toList()),
             protoSpec.getEnvironmentVariablesMap(),
             protoSpec.getIsInteractive(),
-            new File(protoSpec.getJobDirectoryLocation())
+            new File(protoSpec.getJobDirectoryLocation()),
+            protoSpec.getArchiveLocation()
```
mprimi: There's no null in proto fields of type string. This is not going to work.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java
```diff
@@ -249,6 +250,7 @@ public ChangeJobStatusRequest toProtoChangeJobStatusRequest(
         builder.putAllEnvironmentVariables(jobSpecification.getEnvironmentVariables());
         builder.setIsInteractive(jobSpecification.isInteractive());
         builder.setJobDirectoryLocation(jobSpecification.getJobDirectoryLocation().getAbsolutePath());
+        builder.setArchiveLocation(jobSpecification.getArchiveLocation());
```
mprimi: This will throw if the argument is null.

genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/dto/v4/JobRequestSpec.groovy
```diff
@@ -433,6 +440,7 @@ class JobRequestSpec extends Specification {
                 .build()
         def requestedAgentConfig = new AgentConfigRequest.Builder()
                 .withArchivingDisabled(archivingDisabled)
+                .withRequestedArchiveLocationPrefix(requestedArchiveLocationPrefix)
                 .withTimeoutRequested(timeout)
                 .withInteractive(interactive)
                 .withRequestedJobDirectoryLocation(jobDirectoryLocation)
```
mprimi: I suggest at least one test case where it's null (it's going to surface the bugs in conversion i pointed out above)

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobSpecificationServiceImpl.java
```diff
@@ -198,7 +201,13 @@ public JobSpecification resolveJobSpecification(final String id, @Valid final Jo
                 applicationResources,
                 this.generateEnvironmentVariables(id, jobRequest, cluster, command),
                 jobRequest.getRequestedAgentConfig().isInteractive(),
-                jobRequest.getRequestedAgentConfig().getRequestedJobDirectoryLocation().orElse(DEFAULT_JOB_DIRECTORY)
+                jobRequest.getRequestedAgentConfig().getRequestedJobDirectoryLocation().orElse(DEFAULT_JOB_DIRECTORY),
+                toArchiveLocation(
+                    jobRequest
+                        .getRequestedAgentConfig()
+                        .getRequestedArchiveLocationPrefix()
+                        .orElse(defaultArchiveLocation),
```
mprimi: Is the server always sending back an archive location if one was not specified?

genie-web/src/main/resources/db/migration/h2/V4_0_0__Genie_4.sql
```diff
@@ -146,6 +146,8 @@ ALTER TABLE `jobs`
   ADD COLUMN `claimed` BOOLEAN NOT NULL DEFAULT FALSE;
 ALTER TABLE `jobs`
   ADD COLUMN `v4` BOOLEAN NOT NULL DEFAULT FALSE;
+ALTER TABLE `jobs`
+  ADD COLUMN `requested_archive_location_prefix` VARCHAR(1024) DEFAULT NULL;
```
mprimi: @tgianos does modifying this file in place make it hard to deploy to a stack already running V4?
Or does FlyWay magic take care of it?

genie-web/src/test/java/com/netflix/genie/web/jpa/entities/JobEntityUnitTests.java
```diff
@@ -917,6 +917,20 @@ public void canSetV4() {
...
+    /**
+     * Test setter/getter.
+     */
+    @Test
+    public void canSetRequestedArchiveLocationPrefix() {
+        Assert.assertFalse(this.jobEntity.getRequestedArchiveLocationPrefix().isPresent());
+        final String requestedArchiveLocationPrefix = UUID.randomUUID().toString();
+        this.jobEntity.setRequestedArchiveLocationPrefix(requestedArchiveLocationPrefix);
+        Assert.assertThat(
+            this.jobEntity.getRequestedArchiveLocationPrefix().orElse(UUID.randomUUID().toString()),
```
mprimi: I find the `orElse` here a bit confusing... how about just `.get`?

genie-web/src/main/resources/db/migration/h2/V4_0_0__Genie_4.sql
tgianos: @mprimi Sumit asked me about this too. For now we are ok because our dev/milestone stacks internally blow away our databases on every deployment and take a clone of test and re-apply the migration script. Once we push a version of 4.0 to test/prod we will no longer be able to modify this file.

Flyway works by hashing the migration scripts (which have a convention file name definition to determine order) and using a database lookup table to see if that migration has already been applied. In this case since we blow away the databases of our development stacks it's fine but as I said once this goes to test which we clone from it will break because the hashes of this file (after modifications) would be different than the `V4_0_0__Genie_4.sql` script that was already applied to the test database acting as the source.

Once we push to test we should no longer modify this file and instead should create something like `V4_0_0_1_further_mods.sql` for subsequent modifications and up that 4th digit till official OSS 4.0 release.

It's one reason I wanted to hold off on pushing to test/prod but it shouldn't be too bad. Just more small migration script files.

Let me know if this doesn't make sense @mprimi @sumitnetflix

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
sumitnetflix: Will fix the comment

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/ArchivalService.java
sumitnetflix: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
sumitnetflix: I am a bit confused. In the previous chat we had regarding the comment you wanted me to provide an example where the example is more like how a genie job folder actually looks like. Hence {$jobId}. Lets have a quick chat and clarify @mprimi

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
sumitnetflix: This will not work for the case where the folder is not empty. It will unnecessary create an entry for an empty folder. Hence we needed the children.empty check i.e 

```if (children == null || children.length == 0) ```
Regardless, since we dont plan to save empty folders now this is redundant and will anyways be simplified in the change I will do

File: genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java (hunk 23 above)
Comment (sumitnetflix) [reply genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java]: 👍

File: genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/AgentConfigRequest.java (hunk 16 above)
Comment (sumitnetflix) [reply genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/AgentConfigRequest.java]: @mprimi Adding `JobArchiveMetaData` makes sense?

File: genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java (hunk 34 above)
Comment (sumitnetflix) [reply genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java]: I think you are reading this the other way. This one is converting from protoSpec.

File: genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java (hunk 35 above)
Comment (sumitnetflix) [reply genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java]: I was converting to an empty value in the jobSpecification constructor to handle that. 
Will be moving to a getter which returns Optional<String> for archiveLocation and will do the necessary check before populating.

File: genie-web/src/main/java/com/netflix/genie/web/services/impl/JobSpecificationServiceImpl.java (hunk 37 above)
Comment (sumitnetflix) [reply genie-web/src/main/java/com/netflix/genie/web/services/impl/JobSpecificationServiceImpl.java]: Yes, server always provides default location. This was a change after I had a discussion with Tom and had conveyed the same on #genie-developers

So there is no way to shut off archival.

File: genie-web/src/main/resources/db/migration/postgresql/V4_0_0__Genie_4.sql (hunk 21 above)
Comment (sumitnetflix) [reply genie-web/src/main/resources/db/migration/postgresql/V4_0_0__Genie_4.sql]: Shall I rename to `requested_archive_location_root` @tgianos based on the previous explanation I provided

File: genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java (hunk 24 above)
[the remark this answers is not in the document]
Comment (sumitnetflix) [reply genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java]: I can add a validator as you suggested and throw an error if its not a s3 uri. I would rather keep this parameter flexible as far as name goes so that tomorrow we can use it for other archival locations and simply relax the constraint in the validator.

File: genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java (hunk 25 above)
Comment (sumitnetflix) [reply genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java]: Sure

File: genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java (hunk 11 above)
Comment (tgianos) [reply genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java]: This is a fair point. Anecdotally it feels to me like this is useless/wasteful I can't definitively say it will be that way in every possible case. For now we could leave it in and come back later if we decide it really isn't worth it or is wasting too much time. I'm not sure I can think of a case where an empty folder is useful information for any command/process but perhaps there's something you all can think of that i'm missing. Maybe the absence of an expected file is useful information though I'd argue that the directory not being present clearly also indicates the expected file doesn't exist.

File: genie-web/src/main/resources/db/migration/postgresql/V4_0_0__Genie_4.sql (hunk 21 above)
Comment (tgianos) [reply genie-web/src/main/resources/db/migration/postgresql/V4_0_0__Genie_4.sql]: eh, I'm not going to pull my hair out either way. `requested_archive_location` vs. `archive_location` to me represents a concrete ask vs. realized value that have expected differences. That may just be my domain knowledge though. I don't think `root` or `prefix` adds any real information or value and is just more characters all over, but again it's not something worth killing ourselves over. I just go for simplicity where possible.

File: genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/AgentConfigRequest.java (hunk 16 above)
Comment (mprimi) [reply genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/AgentConfigRequest.java]: No strong preference at this point. Whatever the two of you like is ok with me.

File: genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java (hunk 35 above)
Comment (mprimi) [reply genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java]: I think the nuances of dealing with proto fields belong in the proto converter

File: genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java (hunk 34 above)
Comment (mprimi) [reply genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java]: Comment applies to conversion both ways.

File: genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java (hunk 30 above)
Comment (mprimi) [reply genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java]: Back then your service was behaving differently so old comment doesn't apply anymore.

File: genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java (hunk 32 above)
Comment (mprimi) [reply genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java]: Fine, if you want to skip empty folder creation then it is
```
if (children == null) {
  uploadFile
} else if (children.isEmpty) {
  createEmptyFolder
} else {
  for c in children {
    archiveRecursively
  }
}
```

File: genie-web/src/main/java/com/netflix/genie/web/services/impl/JobSpecificationServiceImpl.java (hunk 37 above)
Comment (mprimi) [reply genie-web/src/main/java/com/netflix/genie/web/services/impl/JobSpecificationServiceImpl.java]: I missed the part where there's no way to opt-out. I don't think it's a good idea..

File: genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/dto/v4/JobRequestSpec.groovy (hunk 36 above)
Comment (sumitnetflix) [reply genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/dto/v4/JobRequestSpec.groovy]: Sure

File: genie-web/src/test/java/com/netflix/genie/web/jpa/entities/JobEntityUnitTests.java (hunk 39 above)
Comment (sumitnetflix) [reply genie-web/src/test/java/com/netflix/genie/web/jpa/entities/JobEntityUnitTests.java]: It is confusing. Will fix

File: genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentValidators.java [hunk 40]
```diff
@@ -53,6 +54,25 @@ public void validate(final String name, final String value) throws ParameterExce
...
+    /**
+     * Validates a string parameter is a valid S3 uri.
+     */
+    public static class S3URIValidator implements IParameterValidator {
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public void validate(final String name, final String value) throws ParameterException {
```
tgianos: There is some code we already have for this in `S3FileTransferImpl` that we may want to refactor for reuse. It doesn't have to be done as part of this PR but a `tech-debt` ticket can be filed. That code is a little more robust than this validation:

```    @VisibleForTesting
    AmazonS3URI getS3Uri(final String path) throws GenieBadRequestException {
        if (!S3_PREFIX_PATTERN.matcher(path).matches()) {
            throw new GenieBadRequestException(String.format("Invalid prefix in path for s3 file %s", path));
        }
        // Delegate validation and parsing to AmazonS3URI.
        // However it cannot handle "s3n://", so strip the 'n'
        final String adjustedPath = path.replaceFirst("^s3n://", "s3://");
        final AmazonS3URI uri;
        try {
            uri = new AmazonS3URI(adjustedPath, false);
        } catch (IllegalArgumentException e) {
            throw new GenieBadRequestException(String.format("Invalid path for s3 file %s", path), e);
        }
        if (StringUtils.isBlank(uri.getBucket()) || StringUtils.isBlank(uri.getKey())) {
            throw new GenieBadRequestException(String.format("Invalid blank components in path for s3 file %s", path));
        }

        final boolean bucketPassesStrictValidation = S3_BUCKET_PATTERN.matcher(uri.getBucket()).matches();
        final boolean keyPassesStrictValidation = S3_KEY_PATTERN.matcher(uri.getKey()).matches();
        // URL fails strict validation check!
        if (!bucketPassesStrictValidation || !keyPassesStrictValidation) {
            if (this.s3FileTransferProperties.isStrictUrlCheckEnabled()) {
                throw new GenieBadRequestException(
                    String.format(
                        "Invalid bucket %s in path for s3 file %s",
                        uri.getBucket(),
                        path
                    )
                );
            } else {
                log.warn("S3 URL fails strict validation: \"{}\"", path);
                this.urlFailingStrictValidationCounter.increment();
            }
        }
        return uri;
    }```

File: genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java [hunk 41]
```diff
@@ -0,0 +1,46 @@
...
+
+import com.netflix.genie.agent.execution.exceptions.ArchivalException;
+import com.netflix.genie.agent.execution.services.ArchivalService;
+import lombok.extern.slf4j.Slf4j;
+
+import java.nio.file.Path;
+import java.net.URI;
+
+/**
+ * Implementation of ArchivalService for S3 which does no archival.
```
tgianos: This description isn't really correct. It's not an `ArchivalService for S3` it's a no op.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Slf4j
+class DummyArchivalServiceImpl implements ArchivalService {
```
tgianos: This isn't a very good name. `DefaultArchivalServiceImpl` or `NoOpArchivalServiceImpl` is much clearer. `Dummy` sounds like some sort of mock for testing.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+import java.nio.file.Path;
+import java.net.URI;
+
+/**
+ * Implementation of ArchivalService for S3 which does no archival.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Slf4j
```
tgianos: If you're not going to log anything there's no point to including a logger in the class.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+class DummyArchivalServiceImpl implements ArchivalService {
+
+    /**
+     * No archival is done.
+     * @param path      path to the file/dir to archive
+     * @param targetURI target uri for the archival location
+     * @throws ArchivalException
+     */
+    @Override
+    public void archive(final Path path, final URI targetURI) throws ArchivalException {
```
tgianos: At the least I would say this should log a message saying that it was called but no action will be taken.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
```diff
@@ -107,4 +110,27 @@ public KillService killService(final ApplicationEventPublisher applicationEventP
...
         return new LaunchJobServiceImpl();
     }
+
+    /**
+     * Provide a lazy {@link ArchivalService} bean in case no AmazonS3 client can be created.
+     * @return A {@link ArchivalService} instance
+     */
+    @Bean
+    @Lazy
+    @ConditionalOnClass(value = AmazonS3.class)
```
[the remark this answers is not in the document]
tgianos: Yeah this is definitely incorrect. it should be `@ConditionalOnMissingBean(ArchivalService.class)` and then the other one should be `@ConditionalOnBean(AmazonS3.class)`. In the future I suspect we'll want to collect all active implementations into a `List` or `Set` or best into a `FactoryBean` but the bottom line here is you only want this default version to be instantiated if no other bean of this type is available in the context.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
```diff
@@ -107,4 +110,27 @@ public KillService killService(final ApplicationEventPublisher applicationEventP
...
+    }
+
+    /**
+     * Provide a lazy {@link ArchivalService} bean if one hasn't already been defined.
+     * @param amazonS3 Amazon S3 client instance
+     * @return A {@link ArchivalService} instance
+     */
+    @Bean
+    @Lazy
+    @ConditionalOnMissingBean(ArchivalService.class)
```
tgianos: `@ConditionalOnBean(AmazonS3.class)`

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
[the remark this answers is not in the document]
sumitnetflix: Here is the issue. Lets say we name it `s3ArchivalPrefix` and kragle or some other library which have migrated to the agent is making the call to the agent passing this option `s3ArchivalPrefix`. Then we rename it. Then those libraries will have to be changed. It seems unnecessary for them to do this. Hence I would like to stick to a more generic name `archiveLocationPrefix`. Since there is a validator now, non S3 urls will anyways fail. The description also says that currently only S3 is supported. This keeps things a little more flexible.

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImplSpec.groovy
```diff
@@ -0,0 +1,206 @@
...
+class S3ArchivalServiceImplSpec extends Specification {
+    @Rule
+    TemporaryFolder temporaryFolder
+    AmazonS3 amazonS3 = Mock()
+    WritableResource resource = Mock()
+    S3ArchivalServiceImpl s3ArchivalService
+    File jobDir
+    String bucketName = "netflix-dataoven-test-users";
+    String baseLocation = "genie/main/logs";
+    File stdout
```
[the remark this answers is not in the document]
sumitnetflix: Ok. Will take that as a recommendation for future tests.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
sumitnetflix: Hmm.. I see the problem. I wrongly assumed that the beans with `@ConditionalOnMissingBean` are evaluated in the end. Will fix.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/S3ArchivalServiceImpl.java
```diff
@@ -0,0 +1,186 @@
...
+     * underneath it. If the directory is empty create an empty object in S3.
+     * For a regular child file create an object in S3 and copy the file contents.
+     * For a empty child directory create an empty object in S3
+     *
+     * @param fileToArchive           file/dir to archive.
+     * @param root                    root of the directory being archived.
+     *                                for a regular file, same as the file being archived
+     * @param baseLocationAmazonS3URI base location for storing the file in S3.
+     */
+    private void archiveRecursively(
```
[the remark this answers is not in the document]
sumitnetflix: No use. Removed.

genie-agent/src/test/groovy/com/netflix/genie/agent/cli/JobRequestArgumentsImplSpec.groovy
```diff
@@ -120,7 +123,29 @@ class JobRequestArgumentsImplSpec extends Specification {
...
+
+        when:
+        jCommander.parse("--archiveLocationPrefix", "")
+
+        then:
+        thrown(ParameterException)
+
+    }
+
+        class TestOptions {
```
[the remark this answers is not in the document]
sumitnetflix: fixed

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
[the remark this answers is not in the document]
sumitnetflix: Renamed to `NoOpArchivalServiceImpl`

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+    }
```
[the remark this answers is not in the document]
sumitnetflix: Done

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java
```diff
@@ -41,7 +54,24 @@
...
+                log.info("Attempting to archive job folder to: " + archiveLocation);
+                archivalService.archive(
+                    executionContext.getJobDirectory().toPath(),
+                    new URI(archiveLocation)
+                );
+            } else {
+                log.info("Job folder archival location is empty. Skipping archiving the job folder.");
+            }
+            log.info("Job folder archived to: " + archiveLocation);
+        } catch (ArchivalException | URISyntaxException e) {
```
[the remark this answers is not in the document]
sumitnetflix: `archivalLocation` is being read from the JobSpecificationObject which exposes it as a `Optional<String>`. Not sure what you mean by "Can type conversion be moved upstream?".

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentValidators.java
sumitnetflix: Will do

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
sumitnetflix: Fixed

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/DummyArchivalServiceImpl.java
sumitnetflix: Done

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
```diff
@@ -107,4 +110,27 @@ public KillService killService(final ApplicationEventPublisher applicationEventP
     public LaunchJobService launchJobService() {
         return new LaunchJobServiceImpl();
     }
+
+    /**
```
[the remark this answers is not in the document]
sumitnetflix: Thought about it. I think this strategy makes sense i.e if the S3 credentials are not available then we provide a min functionality which can work without the credentials. In the case of archival that translates to "No archival". For `DownloadService` it can make sense to fail because it can just not function without the credentials. Hence I dont see a discrepancy if the guiding principle is to provide min functionality without the credentails

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
sumitnetflix: I take that back. What i had was totally wrong and not just a bad assumption. Fixed it

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/JobArchivalDataRequest.java
```diff
@@ -0,0 +1,83 @@
...
+ * Job archival options for the Genie agent.
+ *
+ * @author standon
+ * @since 4.0.0
+ */
+@Getter
+@ToString(doNotUseGetters = true)
+@EqualsAndHashCode(doNotUseGetters = true)
+@SuppressWarnings("checkstyle:finalclass")
+public class JobArchivalDataRequest {
```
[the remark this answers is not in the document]
sumitnetflix: I am trying to convey with this name that this is a section of `JobRequest` which includes Job archival data.`JobArchivalRequest` seems like a request to archive the job which is not the right thing in this context

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java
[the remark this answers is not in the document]
mprimi: Why is it exposed as a string if we know it's a URL?
Upstream means converting from serialized form (a string in a proto message) to the actual type (a URI) when we go from proto to DTO

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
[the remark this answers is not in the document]
mprimi: I wouldn't call an archival service that silently *does not archive* "minimum functionality...

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/JobArchivalDataRequest.java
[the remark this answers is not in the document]
mprimi: By this logic, we should call `JobRequest` `JobRequestData`, no?
Since it's not a request to execute a job (the server executes no job) on that codepath..

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentValidators.java
sumitnetflix: Done

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
sumitnetflix: Done

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ShutdownAction.java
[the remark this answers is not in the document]
sumitnetflix: Have created a task to make `archiveLocationPrefix` a `URI`

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/ServicesAutoConfiguration.java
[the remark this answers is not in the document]
sumitnetflix: If a user does not specify `archiveLocationPrefix` and does not care about archival then failing at start up does not make sense. Having a `NoOp` archival makes sense. On the other hand, if the user specifies `archiveLocationPrefix` and wants archival but does not provide S3 credentials then one should fail instead of not archiving. I have created a task to decide how we want to handle such cases. For the time being sticking to the `NoOp` implementation approach.

PR #781, opened 2018-10-04
Upgrade to git publish plugin for documentation

tgianos:
Due to a conflict with jgit and the Nebula plugins this work has been stuck in various branches of my personal fork for a long time. The latest versions of `nebula.netflixoss` have resolved some of those `jgit` conflicts and @sghill helped me resolve some other issues to hopefully get this working.

Once this is merged it should get rid of the final gradle related warning we have lying around during the build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,8 +16,9 @@ buildscript {
 
 plugins {
     id "com.github.kt3k.coveralls" version "2.8.2"
-    id "nebula.netflixoss" version "5.1.1"
-    id "org.ajoberstar.github-pages" version "1.7.2"
+    id "nebula.netflixoss" version "6.0.3"
+    id "org.ajoberstar.grgit" version "3.0.0-beta.1"
+    id "org.ajoberstar.git-publish" version "1.0.1"
 }
 
 apply plugin: "nebula-aggregate-javadocs"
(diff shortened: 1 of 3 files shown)


genie-ui/build.gradle
```diff
@@ -1,6 +1,12 @@
-plugins {
-    id "com.moowork.node" version "1.2.0"
+buildscript {
+    repositories {
+        gradlePluginPortal()
```
mprimi: Isn't this already added globally in the main `build.gradle`?

build.gradle
```diff
@@ -409,7 +410,7 @@ task collectDocumentation(group: "documentation", description: "Copy the documen
 }
 
 gitPublish {
-
+
```
mprimi: Adding some trailing whitespace here

tgianos: it was a copy/paste/modify from the developer example. Do you just mean it's an extra new line or that there are space characters somewhere?

genie-ui/build.gradle
tgianos: @sghill is the one in the main `build.gradle` sufficient for these sub projects as well? You added it in both in your PR into my fork so I didn't touch it deferring to you as the expert but @mprimi might have a good point here

sghill: Good point @mprimi @tgianos. It should delegate to the repositories defined in the parent's build.gradle.

build.gradle
mprimi: Yes it's a line with 4 spaces

PR #784, opened 2018-10-05
Cleanup of job folder

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentDelegates.java
```diff
@@ -98,4 +99,12 @@
         boolean isJobRequestedViaAPI();
     }
 
+    /**
+     * Delegate for job folder cleanup options.
+     */
+    interface CleanupArguments {
+
+        CleanupStrategy getCleanupStrategy();
```
tgianos: Maybe because this is package protected it's not asking for javadoc here?

genie-agent/src/main/java/com/netflix/genie/agent/cli/CleanupArgumentsImpl.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    @Override
+    public CleanupStrategy getCleanupStrategy() {
+        if (skipCleanup) {
+            return CleanupStrategy.NO_CLEANUP;
+        } else if (fullCleanup) {
+            return CleanupStrategy.FULL_CLEANUP;
+        }
+
+        return CleanupStrategy.DEPENDENCIES_CLEANUP;
```
tgianos: Will the help command make it clear this is the default behavior?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -504,4 +482,49 @@ private File createJobEnvironmentFile(
         return Collections.unmodifiableMap(env);
     }
 
+    private void cleanupJobDirectory(
+        final File jobDirectory,
```
tgianos: General comment: 

Perhaps instead of using the old `java.io.File` classes we should try defaulting to using the `java.nio.Path` as the default and save us from creating paths all over?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -504,4 +482,49 @@ private File createJobEnvironmentFile(
...
+
+            case DEPENDENCIES_CLEANUP:
+                final RegexRuleSet cleanupWhitelist = RegexRuleSet.buildWhitelist(
+                    (Pattern[]) Lists.newArrayList(
+                        PathUtils.jobClusterDirectoryPath(jobDirectory, ".*"),
+                        PathUtils.jobCommandDirectoryPath(jobDirectory, ".*"),
+                        PathUtils.jobApplicationDirectoryPath(jobDirectory, ".*")
+                    )
+                        .stream()
+                        .map(PathUtils::jobEntityDependenciesPath)
```
tgianos: Would it be better to combine all these separate map statements into one map? I'm not sure if it helps performance so it might be fine to leave it this way for readability.

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentDelegates.java
mprimi: I've noticed for sure that javadoc is not required on package protected, only public.
However i thought interfaces were default public, so i would have expected an error here.

(Commenting on this interface i added is in line with the rest of the file)

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
mprimi: Migrated to `Path` since it was indeed converting to it everywhere..

mprimi: Also, didn't realize `File` is the old way and `Path` is the new one, will keep this in mind going forward.

PR #785, opened 2018-10-10
Agent restructuring

tgianos:
Restructure agent to improve ability to modify behavior internally and improve performance by tuning Spring Boot based on recommendations mostly found [here](http://presos.dsyer.com/decks/how-fast-is-spring.html) and [here](https://github.com/dsyer/spring-boot-startup-bench)


genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadataImpl.java
```diff
@@ -38,17 +36,18 @@
  * @since 4.0.0
  */
 @Slf4j
-@Component
 @Getter
-@Lazy
-class AgentMetadataImpl implements AgentMetadata {
+public class AgentMetadataImpl implements AgentMetadata {
```
mprimi: Kinda sucks that we have to make all classes public because configs are in a different package... Not much to do about it i guess. Probably not worth scattering configs all over.

genie-agent/src/main/java/com/netflix/genie/agent/cli/CliAutoConfiguration.java
```diff
@@ -0,0 +1,358 @@
...
+/**
+ * Spring auto configuration class to contain all beans involved in the CLI for the Agent.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+@Configuration
+public class CliAutoConfiguration {
+
+//    @Bean
```
mprimi: Should this be removed?

genie-agent-app/build.gradle
```diff
@@ -0,0 +1,49 @@
+import org.apache.tools.ant.filters.ReplaceTokens
```
mprimi: This file doesn't match the commit message, maybe was supposed to be in a different commit?

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadataImpl.java
tgianos: So you're ok with me making the classes public and centralizing the configuration classes? I didn't do this cause I didn't know how you'd feel about it but I wanted to. Right now the configuration classes are spread all over and they could be centralized and consolidated if we made classes public

genie-agent/src/main/java/com/netflix/genie/agent/cli/CliAutoConfiguration.java
tgianos: It's commented back in / modified in another commit later in this PR

genie-agent-app/build.gradle
tgianos: whoops yes, must've gotten added by accident it's supposed to be in the subsequent one

genie-agent/src/main/java/com/netflix/genie/agent/AgentMetadataImpl.java
mprimi: Just to make sure we're on the same page..

Option 1 - Leave implementations (i.e. most stuff that was tagged @Component earlier) package-protected, create a configuration for those beans in the same package.

Option 2 - Make the implementations public. Centralize all configurations in the `config` package

(there's more options but probably not even worth bringing up..)

I can see upside and downside to both options. 
Personally I prefer \#2, but since the convention in this repo is already \#1 i would say let's keep going with that. No strong opinion either way.

tgianos: The second option is also what I'm favoring and is how we do it in the `web` project so it would bring consistency in that respect.

tgianos:
@mprimi @sumitnetflix If you want take a final look over this. I don't think I'll merge it before I have the internal wrapper ready to go. I don't want to break testing of the agent if we don't have a replacement ready but that shouldn't take too long

tgianos:
This PR drive startup times down. Benchmarks run using https://github.com/dsyer/spring-boot-startup-bench/tree/master/launcher. Also improves the ability to modify agent code in wrappers since business logic is in thin jar with auto configs.

Before
```
# Run complete. Total time: 00:02:01

Benchmark               Mode  Cnt  Score   Error  Units
MainBenchmark.jar       avgt   10  3.559 ± 0.317   s/op
MainBenchmark.launcher  avgt   10  3.697 ± 0.171   s/op
MainBenchmark.main      avgt   10  2.523 ± 0.260   s/op
```

After:
```
# Run complete. Total time: 00:01:49

Benchmark               Mode  Cnt  Score   Error  Units
MainBenchmark.jar       avgt   10  3.288 ± 0.306   s/op
MainBenchmark.launcher  avgt   10  3.261 ± 0.354   s/op
MainBenchmark.main      avgt   10  2.281 ± 0.209   s/op
```

PR #786, opened 2018-10-11
Output redirect and job running directory

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/cli/UserConsole.java
```diff
@@ -44,6 +53,13 @@
     private static final String PID_SYSTEM_PROPERTY_NAME = "PID";
     private static final Logger LOGGER = LoggerFactory.getLogger(CONSOLE_LOGGER_NAME);
 
+    /**
+     * Because the banner is printed in the log file and not visible to the user, manually re-print it
+     * in {@code UserConsole}. Use the existing Spring configuration to control this behavior.
+     */
+    private static final String BANNER_LOCATION_SPRING_PROPERTY_KEY = "spring.banner.location";
```
tgianos: Did you find that spring didn't already have constants for these values? There might also be a `@ConfigurationProperties` you can inject for `spring.banner` that would alleviate the need for you to do this manually

Looks like `SpringApplication` has some constants:
```	/**
	 * Default banner location.
	 */
	public static final String BANNER_LOCATION_PROPERTY_VALUE = SpringApplicationBannerPrinter.DEFAULT_BANNER_LOCATION;

	/**
	 * Banner location property key.
	 */
	public static final String BANNER_LOCATION_PROPERTY = SpringApplicationBannerPrinter.BANNER_LOCATION_PROPERTY;```

mprimi: Ah! i found the constants in `SpringApplicationBannerPrinter`, but they're package private.
Didn't notice `SpringApplication` kindly making them public :-)

mprimi:
@tgianos re: job URL
One of the commit prints the job ID (Dan asked for "id or URL").
We could construct and print a URL, but there'd be no guarantee it's the correct one. The only thing the agent knows about the server is hostname and port of the gRPC endpoint, but there's no guarantee that can be used to forge a valid http link.

PR #789, opened 2018-10-23
Arg parsing with separator and create job directory

mprimi:


genie-agent-app/src/main/java/com/netflix/genie/GenieAgentApplication.java
```diff
@@ -60,6 +61,12 @@ private int run(final String[] args) {
         final SpringApplication app = new SpringApplication(GenieAgentApplication.class);
         // Disable parsing of command-line arguments into properties.
         app.setAddCommandLineProperties(false);
-        return SpringApplication.exit(app.run(args));
+
+        //TODO: workaround for https://jira.spring.io/browse/SPR-17416
+        // Spring chokes on argument '--' (a.k.a. bare double dash) conventionally used to separate options from
+        // operands. Perform a token replacement to avoid triggering an error in Spring argument parsing.
+        // Later the original token is restored before the actual Genie argument parsing.
+        final String[] editedArgs = Util.mangleBareDoubleDash(args);
```
tgianos: Instead of mangling and un-mangling did you consider just saving the original in some shared state location and later using it after Spring parsing? It's a minute optimization so not sure if it's worth it or not but would save you from doing all the replace logic twice

genie-agent/src/main/java/com/netflix/genie/agent/cli/Util.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Hidden constructor.
+     */
+    private Util() {
+    }
+
+    /**
+     * Replace all "bare double dash" arguments in the input array.
+     * @param args the arguments array
```
tgianos: Does your auto config within intellij not put a `\n` between the function description and the first `@param` tag? I'm just curious cause I've seen some of @sumitnetflix code format like this too and I don't want to get those weird just formatting commits so we should settle on one or the other

genie-agent/src/main/java/com/netflix/genie/agent/cli/Util.java
```diff
@@ -63,4 +64,46 @@ private Util() {
...
+
+    /**
+     * Get a subset of arguments before the double dash (a.k.a. options).
+     *
+     * @param args the raw array of arguments
+     * @return an array of arguments. Possibly empty, possibly the same as the input array.
+     */
+    public static String[] getOptionArguments(final String[] args) {
+        final List<String> temporary = Arrays.asList(args);
+        final int index = temporary.indexOf(BARE_DOUBLE_DASH);
```
tgianos: It's late and I'm tired so I could be totally off on this but I feel like this could be done in a single pass and save all the conversions and checks. Maybe just iterate through the array and write each item into a resultant array until you encounter the double dash and then return the result array?

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+
+/**
+ * Container for operand arguments (arguments meant for the underlying command, that the agent should not try to
+ * interpret.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+class MainCommandArguments {
```
tgianos: So this is effectively `commandArgs` in a job request right?

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final List<String> arguments = Lists.newArrayList();
+    private final List<String> argumentsSeadonlyView = Collections.unmodifiableList(arguments);
+
+    List<String> get() {
+        return argumentsSeadonlyView;
+    }
+
+    void set(final String[] operandArguments) {
+        arguments.addAll(Arrays.asList(operandArguments));
```
tgianos: Wonder if it's faster to do a `for` with an `add` for each item instead of creating an ArrayList object just to iterate through it and have the underlying code add it. Idle curiosity not sure if you know or not

genie-agent-app/src/main/java/com/netflix/genie/GenieAgentApplication.java
mprimi: Yeah, essentially using a global variable to store the args, then pass empty args to boot and then retrieve the original args from the global variable afterwards, from inside the app.
It just felt heavy-handed. `--` is really the only thing that Spring doesn't like, we may have reasons to let it process the rest.

genie-agent/src/main/java/com/netflix/genie/agent/cli/Util.java
mprimi: It doesn't in the template, but it does when i do a full file re-format of the file.
I'll fix this, and I think it does look better and we should stick with it where possible.

genie-agent/src/main/java/com/netflix/genie/agent/cli/Util.java
mprimi: This could indeed be a lot simpler code at the cost of allocating a couple more objects.
Instinctively went for the more memory efficient, but probably not worth in retrospect, since it's only ever invoked once.

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
mprimi: Yes, JobRequestArgument (interface) doesn't change, and neither does anything downstream of it.
This is just a (package private) trick to manually bind these arguments, rather than letting jCommander do it.

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
mprimi: Probably.. this just went for the one-liner..

mprimi:
Updated with minor changes.

genie-agent/src/main/java/com/netflix/genie/agent/cli/Util.java
mprimi: Fixed

genie-agent/src/main/java/com/netflix/genie/agent/cli/Util.java
mprimi: Simplified

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
```diff
@@ -0,0 +1,46 @@
...
+ * Container for operand arguments (arguments meant for the underlying command, that the agent should not try to
+ * interpret.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+class MainCommandArguments {
+
+    private final List<String> arguments = Lists.newArrayList();
+    private final List<String> argumentsSeadonlyView = Collections.unmodifiableList(arguments);
```
sumitnetflix: typo `argumentSeadonlyView`. Looks like you needed `argumentReadonlyView`

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
```diff
@@ -0,0 +1,46 @@
...
+
+    List<String> get() {
+        return argumentsSeadonlyView;
```
sumitnetflix: same typo

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
```diff
@@ -0,0 +1,46 @@
...
+package com.netflix.genie.agent.cli;
+
+import com.google.common.collect.Lists;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Container for operand arguments (arguments meant for the underlying command, that the agent should not try to
```
sumitnetflix: An example of the kind of arguments that can go here will be helpful

genie-agent/src/main/java/com/netflix/genie/agent/cli/MainCommandArguments.java
mprimi: Fixed, thanks.

PR #790, opened 2018-10-26
Add metric for number of agents connected

mprimi:


mprimi:
@tgianos @sumitnetflix this is very simple change but expecting a bit of back and forth on the details.
Metric name, description, and how we track this can be done in a number of slightly different ways and we just need to pick one that works for everybody.

For example, we could separately count and publish the number of connections associated to an agent that *claimed* a job ID.
Or we could conflate this with jobs running and name it `genie.agent.jobs.running.gauge` (similar to the existing `genie.jobs.running.gauge`).

Let me know what you think.

sumitnetflix:
This looks good to me as a starting point.

genie-web/src/main/java/com/netflix/genie/web/rpc/grpc/services/impl/v4/GrpcHeartBeatServiceImpl.java
```diff
@@ -55,20 +57,25 @@
...
+        final MeterRegistry registry
     ) {
         this.agentRoutingService = agentRoutingService;
         this.taskScheduler = taskScheduler;
+        this.registry = registry;
         this.sendHeartbeatsFuture = this.taskScheduler.scheduleWithFixedDelay(
             this::sendHeartbeats,
             HEART_BEAT_PERIOD_MILLIS
         );
+        this.registry.gauge("genie.grpc.connections.gauge", this.connectionsCounter);
```
tgianos: Instead of all the setting below you should just be able to provide a function to the meter registry with the gauge name and it will sample it for you. In particular it can track collection sizes which is what you're doing here if I'm not mistaken.
[Micrometer Gauge Docs](https://micrometer.io/docs/concepts#_gauges)

tgianos: Fair enough. It's just boilerplate so no big deal. You could use a synchronized map or since it's just a metric and we're not doing something like a financial transaction leave it as a regular map and just take the value as is. If you're planning on using this for OOS or something we'd likely use a rolling count anyway. Either way so long as we're sure we catch all the increment/decrement places this is fine, just thought it was simpler to let the library track the size for us.

mprimi: I'm doing some operations on that map that would not be thread safe with a synchronized map (i.e. lock, inspect, replace, unlock).

PR #796, opened 2018-11-13
Provide default job folder for V4 jobs

tgianos:
This commit adds a default dummy V4 job folder to the system at startup. Since V4 logging isn't fully fleshed out/solved yet current requests through V3 UI for a V4 job output would cause errors. This update creates a dummy folder that enables smoother behavior for users. They'll never see real logs but at least it won't throw a lot of errors or try to forward around the system. Job forwarding look up is short circuited if the job is flagged as V4 to save extra hops.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/resources/handlers/GenieResourceHttpRequestHandler.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/resources/handlers/GenieResourceHttpRequestHandler.java
@@ -20,6 +20,7 @@ package com.netflix.genie.web.resources.handlers;
 import com.netflix.genie.common.internal.jobs.JobConstants;
 import com.netflix.genie.web.resources.writers.DirectoryWriter;
 import com.netflix.genie.web.services.JobFileService;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.core.io.Resource;
 import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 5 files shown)


genie-web/src/main/java/com/netflix/genie/web/resources/handlers/GenieResourceHttpRequestHandler.java
```diff
@@ -57,6 +60,12 @@
      */
     public static final String GENIE_JOB_ID_ATTRIBUTE = GenieResourceHttpRequestHandler.class.getName() + ".jobId";
 
+    /**
+     * The id of a job to use as a placeholder for all requests for V4 job output temporarily while we develop V4 job
+     * support.
+     */
+    public static final String V4_MOCK_JOB_ID = UUID.randomUUID().toString();
```
mprimi: This is a static constant that chances for every instance or every time the application starts.
Would it be less confusing to use an actual string, `mock-agent-job-id-logs-placeholder` or something?

genie-web/src/main/java/com/netflix/genie/web/resources/handlers/GenieResourceHttpRequestHandler.java
```diff
@@ -72,6 +81,15 @@ public GenieResourceHttpRequestHandler(final DirectoryWriter directoryWriter, fi
         super();
         this.directoryWriter = directoryWriter;
         this.jobFileService = jobFileService;
+
+        try {
+            this.jobFileService.createJobDirectory(V4_MOCK_JOB_ID);
```
mprimi: Could also create empty `stdout` and `stderr` since it's kinda part of the contract that those files exist.
Not saying it's our API, but downstream crawlers may *assume* they exist, and they wouldn't be wrong normally.

tgianos:
@mprimi I've addressed your comments and re-pushed

PR #799, opened 2018-11-20
Fix environment parse error surfaced by Travis build

mprimi:


genie-agent/src/test/groovy/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobActionSpec.groovy
```diff
@@ -354,6 +355,7 @@ class SetUpJobActionSpec extends Specification {
         envMap.get(JobConstants.GENIE_CLUSTER_DIR_ENV_VAR) == jobDir.toString() + "/" + JobConstants.GENIE_PATH_VAR + "/" + JobConstants.CLUSTER_PATH_VAR + "/" + cluster.getId()
         envMap.get("SERVER_VARIABLE_KEY") == "SERVER VARIABLE VALUE"
         envMap.get("SETUP_VARIABLE_KEY") == "SETUP VARIABLE VALUE"
+        envMap.get("TRICKY_VARIABLE") == "'foo\nbar\ny'all'"
```
tgianos: Is this all the weird cases you can think of right now or just the one tripping up on travis currently?

mprimi: I added this to make the test fail and therefore validate the patch works.
Multi-line value containing a single quote is what was failing in Travis.

PR #801, opened 2018-11-30
Better status cleanup and more

mprimi:


genie-agent/src/test/groovy/com/netflix/genie/agent/execution/exceptions/AgentExecutionExceptionsSpec.groovy
```diff
@@ -27,31 +27,32 @@ class AgentExecutionExceptionsSpec extends Specification {
     final Throwable CAUSE = new IOException()
 
     @Unroll
-    def "Constructors for exception class #aClass"(Class<? extends Throwable> aClass) {
+    def "Constructors for exception class #aClass"(Class<? extends Throwable> aClass, boolean checked) {
```
tgianos: Just an FYI that you don't need to pass the arguments in the `()` if you don't want to

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/ResolveJobSpecificationAction.java
```diff
@@ -126,6 +126,15 @@ protected Events executeStateAction(final ExecutionContext executionContext) {
             executionContext.setCurrentJobStatus(JobStatus.RESOLVED);
         }
 
+        UserConsole.getLogger().info(
+            "Job request criteria resolved to command '{}' on cluster '{}'",
+            jobSpecification.getCommand().getId(),
```
tgianos: We may want to think of sending back the name with the job specification if we intend to do this. It's distinctly possible that these ID's are pretty non-sensical for end users and would be better to have `on command {name} (id: {id})` as the log message or something

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/HandleErrorAction.java
```diff
@@ -58,14 +58,14 @@ protected void executePreActionValidation() {
...
+        final Optional<JobStatus> finalJobStatus = executionContext.getFinalJobStatus();
         final Optional<String> claimedJobId = executionContext.getClaimedJobId();
 
-        if (currentJobStatus.isPresent() && claimedJobId.isPresent() && currentJobStatus.get().isActive()) {
+        if (claimedJobId.isPresent() && !finalJobStatus.isPresent()) {
             try {
                 agentJobService.changeJobStatus(
                     claimedJobId.get(),
-                    currentJobStatus.get(),
+                    executionContext.getCurrentJobStatus().get(),
```
tgianos: Might be better here instead of just `get()` to explicitly use `orElse(null)` so it's really clear otherwise I think according to the docs you're going to throw a `NoSuchElementException` [Optional#get](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#get--)

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -102,7 +102,10 @@ protected void executePreActionValidation() {
     protected Events executeStateAction(
         final ExecutionContext executionContext
     ) {
+        log.info("Setting up job...");
+
         final String claimedJobId = executionContext.getClaimedJobId().get();
+        final JobSpecification jobSpecification = executionContext.getJobSpecification().get();
```
tgianos: Possible exception here if there is no `jobSpecification`. Perhaps you're making the assumption it's already there but you might want to use `orElseThrow()` and control what exception and message are thrown

genie-agent/src/main/java/com/netflix/genie/agent/cli/ExecCommand.java
```diff
@@ -103,7 +103,7 @@ public ExitCode run() {
             throw new RuntimeException("Job execution error", firstActionErrorException);
         }
 
-        final JobStatus finalJobStatus = executionContext.getFinalJobStatus();
+        final JobStatus finalJobStatus = executionContext.getFinalJobStatus().get();
```
tgianos: `orElseThrow` here as well?

PR #805, opened 2018-12-05
Remove EC2 local hostname as option for hostname resolution

tgianos:
In discussions with people it is preferred to just use the internal EC2 IP address to avoid
other potential issues. Now we'll just use the IPV4 address and fallback to whatever can be
found from InetAddress.getHostname otherwise. This will behave similar to Genie 3.x now.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/configs/aws/GenieAwsApiAutoConfiguration.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/configs/aws/GenieAwsApiAutoConfiguration.java
@@ -30,9 +30,6 @@ import org.springframework.cloud.aws.context.annotation.ConditionalOnAwsCloudEnv
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
-import java.net.InetAddress;
-import java.net.UnknownHostException;
-
 /**
  * Beans and configuration specifically for MVC on AWS.
  *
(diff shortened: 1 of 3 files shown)


genie-web/src/main/java/com/netflix/genie/web/configs/aws/GenieAwsApiAutoConfiguration.java
```diff
@@ -58,11 +58,6 @@
     @Bean
     @ConditionalOnMissingBean(GenieHostInfo.class)
     public GenieHostInfo genieHostInfo() throws UnknownHostException {
-        final String ec2LocalHostName = EC2MetadataUtils.getLocalHostName();
-        if (StringUtils.isNotBlank(ec2LocalHostName)) {
-            return new GenieHostInfo(ec2LocalHostName);
-        }
-
         final String ec2Ipv4Address = EC2MetadataUtils.getPrivateIpAddress();
```
mprimi: Does this work on non-EC2 hosts? i.e. jenkins?
Could we add a unit-test that checks for this string to be an IPv4 string?

tgianos: This auto config is protected by `@ConditionalOnAwsCloudEnvironment` so it won't even be activated if not on an EC2 node there is a different `GenieHostInfo` bean created by default in `GenieApiAutoConfiguration` if this AWS version wasn't used. The fallback case should really never be hit but is there for completeness.

 I could add said unit test but aren't I then really checking the functionality of `EC2MetadataUtils`?

tgianos:
@mprimi @sumitnetflix I made a couple of changes. Take a look please.

PR #806, opened 2018-12-05
Finer grained configuration of active jobs limit

mprimi:


mprimi:
PR against the wrong branch 😅

PR #807, opened 2018-12-05
Make active jobs user limit value customizable on a per-user basis

mprimi:
The existing mechanism to limit the maximum number of jobs applies equally to all users.
This can result in having to configure its value to a 'known accepted worst case'. And that limit is then applied to everyone else.

The new mechanism allows to set a default value that applies to most users, plus a set of custom limits for individual users.
These limits can be higher or lower than the default.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
@@ -17,11 +17,13 @@
  */
 package com.netflix.genie.core.properties;
 
+import com.google.common.collect.Maps;
 import lombok.Getter;
 import lombok.Setter;
 import org.springframework.validation.annotation.Validated;
 
 import javax.validation.constraints.Min;
+import java.util.Map;
 
 /**
  * Properties related to user limits in number of active jobs.
@@ -46,4 +48,16 @@ public class JobsUsersActiveLimitProperties {
     private boolean enabled = DEFAULT_ENABLED;
     @Min(value = 1)
     private int count = DEFAULT_COUNT;
+    private Map<String, Integer> overrides = Maps.newHashMap();
+
+    /**
+     * Get the maximum number of jobs a user is allowed to run concurrently.
+     * Checks whether the user has a special limit associated, and if not it returns the global default.
+     *
+     * @param user the user name
+     * @return the maximum number of jobs
+     */
+    public int getUserLimit(final String user) {
+        return overrides.getOrDefault(user, count);
+    }
 }
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -135,8 +135,12 @@ to work.
 |false
 
 |genie.jobs.users.activeLimit.count
-|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
+|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true. This limit applies to users that don't have an override set via `genie.jobs.users.activeLimit.overrides.<user-name>`.
 |100
+
+|genie.jobs.users.activeLimit.overrides.<user-name>
+|The maximum number of active jobs that user 'user-name' is allowed to have. This property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
+|-
 
 |genie.jobs.completionCheckBackOff.minInterval
 |The minimum time between checks for job completion in milliseconds. Must be greater than zero.
--- a/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
@@ -245,7 +245,7 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
             log.info("Checking if can run job {} from user {}", jobRequest.getId(), jobRequest.getUser());
             final JobsUsersActiveLimitProperties activeLimit = this.jobsProperties.getUsers().getActiveLimit();
             if (activeLimit.isEnabled()) {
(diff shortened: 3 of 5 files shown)


genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -135,9 +135,13 @@ to work.
 |false
 
 |genie.jobs.users.activeLimit.count
-|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
+|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true. This limit applies to users that don't have a more specific value set, see below.
```
tgianos: Rather than `see below` maybe a direct reference to `see genie.jobs.users.activeLimit.specialUser.user-name` cause below could change at any time if someone copies this without knowing?

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -135,9 +135,13 @@ to work.
...
 |100
 
+|genie.jobs.users.activeLimit.specialUser.<user-name>
```
tgianos: Just a note: This is fine in Boot 1.5 but in boot 2.+ this will break their new [relaxed bindings](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0) rules. I think in the `master` branch I changed `activeLimit` to `active-limit` already but I don't exactly remember.

Also instead of `specialUser` maybe just `whitelist`?

genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
```diff
@@ -46,4 +48,9 @@
     private boolean enabled = DEFAULT_ENABLED;
     @Min(value = 1)
     private int count = DEFAULT_COUNT;
+    private Map<String, Integer> specialUser = Maps.newHashMap();
+
+    public int getUserLimit(final String user) {
+        return specialUser.getOrDefault(user, count);
```
tgianos: I see in this diff that `count = DEFAULT_COUNT`. I assume somewhere else in here it is binding the value provided in the properties file that could overwrite that?

genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
```diff
@@ -46,4 +48,9 @@
     private boolean enabled = DEFAULT_ENABLED;
     @Min(value = 1)
     private int count = DEFAULT_COUNT;
+    private Map<String, Integer> specialUser = Maps.newHashMap();
+
+    public int getUserLimit(final String user) {
```
tgianos: Missing javadoc to explain the fallback process?

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Indeed, i know porting this to `master` will be slightly different. But trying to get the simplest possible change into `3.x.x`.

Whitelist usually denotes "you're on the list, go ahead" -type of behavior.
This is not. We can use this to set lower-than-default or higher-than-default.

genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
mprimi: It is initialized to default, but when spring properties evaluate and bind values it can set it to something else (via generated setter).

genie-docs/src/docs/asciidoc/_properties.adoc
tgianos: makes sense. `userOverride` then?

mprimi: (on the other side, I'm not a big fan of `specialUser` either. Any other idea?

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -135,9 +135,13 @@ to work.
 |false
 
 |genie.jobs.users.activeLimit.count
-|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
+|The maximum number of active jobs a user is allowed to have. Once a user hits this limit, jobs submitted are rejected. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true. This limit applies to users that don't have an override set via `genie.jobs.users.activeLimit.overrides.<user-name>`.
```
sumitnetflix: There is a typo before your change - `This is property is`. It will be good to fix

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -135,9 +135,13 @@ to work.
...
 |100
 
+|genie.jobs.users.activeLimit.overrides.<user-name>
+|The maximum number of active jobs that user 'user-name' is allowed to have. This is property is ignored unless `genie.jobs.users.activeLimit.enabled` is set to true.
```
sumitnetflix: Typo `This is property is`

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Fixed, thanks

mprimi:
@tgianos can you take another look?

genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
```diff
@@ -46,4 +48,15 @@
     private boolean enabled = DEFAULT_ENABLED;
     @Min(value = 1)
     private int count = DEFAULT_COUNT;
+    private Map<String, Integer> overrides = Maps.newHashMap();
+
+    /**
+     * Get the maximum number of jobs a user is allowed to run concurrently.
+     * Checks whether the user has a special limit associated, and if not it returns the global default.
+     * @param user the user name
```
tgianos: nitpick here: when I run the auto formatting in Intellij it will add a new line between the description and this first `@param`. To avoid these formatting subsequent commits can you fix it now? You might have done it outside of intellij?

PR #809, opened 2018-12-13
Make user job limit properties dynamic/fast

mprimi:
In order to be able to tweak concurrent job limit without release and deployment, make this group of property dynamic.
This is implemented by declaring this group as a standalone bean (as opposed to the natural nesting) and accessing it through a provider.

The combination of `@Scope("prototype")` and `@ConfigurationProperties` ensures the properties reflect changes to the environment, while at the same time avoid re-binding unnecessarily.

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
@@ -55,6 +55,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.hibernate.validator.constraints.NotBlank;
 import org.hibernate.validator.constraints.NotEmpty;
 import org.springframework.aop.TargetClassAware;
+import org.springframework.beans.factory.ObjectFactory;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
@@ -92,6 +93,7 @@ public class JobCoordinatorServiceImpl implements JobCoordinatorService {
     private final CommandService commandService;
     private final List<ClusterLoadBalancer> clusterLoadBalancers;
     private final JobsProperties jobsProperties;
+    private final ObjectFactory<JobsUsersActiveLimitProperties> jobsUsersActiveLimitPropertiesProvider;
     private final String hostName;
 
     // Metrics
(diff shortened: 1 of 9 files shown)


genie-web/src/main/java/com/netflix/genie/web/configs/PropertiesConfig.java
```diff
@@ -77,4 +79,18 @@ public HealthProperties healthProperties() {
     public S3FileTransferProperties s3FileTransferProperties() {
         return new S3FileTransferProperties();
     }
+
+    /**
+     * Subset of properties concerned with per-user number of active jobs.
+     * Notice the 'prototype' scope. If this bean is injected via {@code javax.inject.Provider} or Spring's
```
tgianos: Better to use `@link` here instead of `@code` so the javadocs are linked? not sure.

genie-web/src/main/java/com/netflix/genie/web/configs/PropertiesConfig.java
```diff
@@ -77,4 +79,18 @@ public HealthProperties healthProperties() {
...
+     * {@code ObjectFactory}, then values are bound every time, i.e., fast/dynamic properties behavior.
+     *
+     * @return The active job user limit property structure
+     */
+    @Bean
+    @Scope("prototype")
+    @ConfigurationProperties("genie.jobs.users.activeLimit")
```
tgianos: To be clear here you made it possible to completely turn off the feature dynamically instead of just making the `override` map dynamic? I ask cause above it means you have to call `get()` every time just to check enabled. Probably not a huge deal just wanted to be clear

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
```diff
@@ -92,6 +93,7 @@
     private final CommandService commandService;
     private final List<ClusterLoadBalancer> clusterLoadBalancers;
     private final JobsProperties jobsProperties;
+    private final ObjectFactory<JobsUsersActiveLimitProperties> jobsUsersActiveLimitPropertiesProvider;
```
tcellucci: I would have preferred `javax.inject.Provider` here to avoid vendor-specific class, but ObjectFactory is perfectly serviceable

genie-web/src/main/java/com/netflix/genie/web/configs/PropertiesConfig.java
mprimi: Correct.

genie-core/src/main/java/com/netflix/genie/core/services/impl/JobCoordinatorServiceImpl.java
mprimi: I used `ObjectFactory` because this is a patch for our stable for `3.3.x` branche and i choose to avoid introducing a new dependency.
When i port this change to `master` I will use `Provider`.

PR #810, opened 2018-12-13
Upgrade to Gradle 5

tgianos:
We may want to squash merge this one but I've left them as separate commits for now to make it clear what was done in different pieces and why.


genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -513,7 +514,7 @@ private File createJobEnvironmentFile(
         // Variables in environment file are base64 encoded to avoid escaping, quoting.
         // Decode all values.
         env.keySet().forEach(key ->
-            env.compute(key, (k, v) -> new String(Base64.decodeBase64(v)))
+            env.compute(key, (k, v) -> new String(Base64.decodeBase64(v), StandardCharsets.UTF_8))
```
tgianos: @mprimi `Spotbugs` was complaining about this not being portable due to internationalization rules since we didn't provide a character encoding. I added `UTF-8` but wanted to check with you if you think this is ok. I looked at the bash script where you're encoding the variables and didn't see anything specific to encoding.

genie-test/src/main/java/com/netflix/genie/test/suppliers/RandomSuppliers.java
```diff
@@ -33,16 +34,15 @@
      * Get a random String.
      */
     public static final Supplier<String> STRING = UUID.randomUUID()::toString;
+    private static final Random RANDOM = new Random();
```
tgianos: Spotbugs complained about inefficiencies of the other implementation so switched to this.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
tgianos: Refreshing this now that the commit exists again. @mprimi take a look

mprimi: Encoding would transform everything into plain old ASCII, so parsing as UTF-8 should be fine.

PR #813, opened 2019-01-07
Make per-user limit on number of jobs dynamic (master)

mprimi:
Attempt to fetch the per-value limit from the environment, and fall back to the current or default if an override doesn't exist.


genie-web/src/main/java/com/netflix/genie/web/properties/JobsUsersActiveLimitProperties.java
```diff
@@ -72,6 +78,19 @@
...
+            return env.getProperty(
+                USER_LIMIT_OVERRIDE_PROPERTY_PREFIX + user,
+                Integer.class,
+                this.count
+            );
+        }
+        return this.count;
+    }
+
+    @Override
```
tgianos: missing `@InheritDoc`?

PR #814, opened 2019-01-08
Make per-user limit on number of jobs dynamic (3.3.x)

mprimi:


genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
```diff
@@ -34,7 +35,23 @@
 @Getter
 @Setter
 @Validated
-public class JobsUsersActiveLimitProperties {
+public class JobsUsersActiveLimitProperties implements EnvironmentAware {
+
+    /**
+     * The property prefix for job user limiting.
+     */
+    public static final String PROPERTY_PREFIX = "genie.jobs.users.active-limit";
```
tgianos: In 3.3.x wasn't this `activeLimit`? Will this necessitate changes for everyone? Not sure if spring Boot 1.5 will treat it the same as 2.0 does.

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -131,11 +131,11 @@ to work.
 |false
 
 |genie.jobs.users.activeLimit.enabled
```
tgianos: Note this doesn't match the prefix you have in the code

genie-core/src/main/java/com/netflix/genie/core/properties/JobsUsersActiveLimitProperties.java
mprimi: This is wrong indeed. I cherry-picked and missed the fact this does something different..

PR #816, opened 2019-01-15
Agent/Server handshake

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobService.java
```diff
@@ -39,6 +40,16 @@
...
 public interface AgentJobService {
 
+    /**
+     * Perform server handshake. Before going any further, ensure the server is reachable and that this agent is
+     * compatible with it.
+     *
+     * @param agentClientMetadata metadata about the client making this request
+     * @throws HandshakeException if the server rejects this client
+     */
+    void handshake(
```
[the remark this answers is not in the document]
mprimi: I agree in principle, but if you look at the rest of this interface, it's the pattern we've been using.
For example, failure to reserve a job, resolve a specification, etc. could all return some sort of "outcome object" with an details.
What we're doing instead is either returning the result of successful operation or an exception with details.

We could deviate from this for this specific `handshake` API, I thought about it.
I just saw more value in sticking to the existing convention, since the alternative (outcome return value) doesn't add a lot of value.

genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/HandshakeException.java
```diff
@@ -0,0 +1,47 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
```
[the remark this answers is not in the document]
mprimi: Done!

genie-web/src/main/java/com/netflix/genie/web/services/AgentFilterService.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Service to identify agent/clients that the server wants to refuse service to.
+ * To be used, for example, to blacklist clients running a given version that is known to be deprecated or otherwise
+ * incompatible.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public interface AgentFilterService {
```
tgianos: Would `AgentVerificationService` be a better name? Just want to make clear what it is. I guess HTTP Filters can do rejections too so I could see why this would work as well just throwing it out there

genie-web/src/main/java/com/netflix/genie/web/services/AgentFilterService.java
```diff
@@ -0,0 +1,87 @@
...
+    @AllArgsConstructor
+    @Getter
+    class InspectionReport {
+        private final InspectionDecision decision;
+        private final String message;
+
+        /**
+         * The possible decisions of an {@link AgentMetadataInspector}.
+         */
+        public enum InspectionDecision {
```
tgianos: Is there value in this beyond pass fail? your `Inconclusive` passes but really adds nothing

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void acceptOrThrow(@Valid final AgentClientMetadata agentClientMetadata) throws GenieAgentRejectedException {
+        for (final AgentMetadataInspector agentMetadataInspector : agentMetadataInspectorList) {
+            final InspectionReport outcome = agentMetadataInspector.inspect(agentClientMetadata);
+            final InspectionReport.InspectionDecision decision = outcome.getDecision();
+            final String message = outcome.getMessage();
+
+            log.info(
```
tgianos: Should probably be debug?

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+                agentClientMetadata,
+                decision.name(),
+                message
+            );
+
+            if (decision == AgentFilterService.InspectionReport.InspectionDecision.REJECT) {
+                throw new GenieAgentRejectedException("Agent rejected, " + message);
+            } else if (decision == AgentFilterService.InspectionReport.InspectionDecision.ACCEPT) {
+                // Bypass the rest of the inspectors.
+                return;
```
tgianos: I don't understand why we would bypass the rest of them? Wouldn't we want to execute all the available inspectors and only fully Accept when all say Accept?

genie-web/build.gradle
```diff
@@ -55,6 +55,9 @@ dependencies {
     compile("org.springframework.cloud:spring-cloud-starter-zookeeper")
     compile("org.springframework.integration:spring-integration-zookeeper")
     compile("org.springframework.retry:spring-retry")
+    compile("org.apache.maven:maven-artifact:3.5.4") {
```
tgianos: Generally we don't include versions in the subproject `build.gradle` and instead define them in the dependency management section of the parent build file so that we can ensure the same version is used across the project if it needs to be

genie-web/src/main/java/com/netflix/genie/web/util/BaseRegexAgentMetadataInspector.java
```diff
@@ -0,0 +1,98 @@
...
+ * Abstract base class for {@link AgentFilterService.AgentMetadataInspector} based on regex matching.
+ * Can be configure to accept or reject in case of match. It is a pass-through otherwise.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+abstract class BaseRegexAgentMetadataInspector implements AgentFilterService.AgentMetadataInspector {
+
+    private final AtomicReference<Pattern> patterncReference = new AtomicReference<>();
```
tgianos: looks like an extra `c` snuck in here

genie-web/src/main/java/com/netflix/genie/web/util/BaseRegexAgentMetadataInspector.java
```diff
@@ -0,0 +1,98 @@
...
+    private final AgentFilterService.InspectionReport.InspectionDecision decisionIfMatch;
+    private String lastCompiledPattern;
+
+    BaseRegexAgentMetadataInspector(
+        final AgentFilterService.InspectionReport.InspectionDecision decisionIfMatch
```
tgianos: Blargh. Maybe based on this you should break `InspectionReport` into its own class file? Pretty long coordinate

genie-web/src/main/java/com/netflix/genie/web/util/MinimumVersionAgentMetadataInspector.java
```diff
@@ -0,0 +1,94 @@
...
+
+        final String minimumVersionString = this.agentFilterProperties.getMinimumVersion();
+        if (StringUtils.isNotBlank(minimumVersionString)) {
+            // Minimum version is set, check it
+            final String agentVersionString = agentClientMetadata.getVersion().orElse(null);
+            if (StringUtils.isNotBlank(agentVersionString)) {
+                final ArtifactVersion minimumVersion = new DefaultArtifactVersion(minimumVersionString);
+                final ArtifactVersion agentVersion = new DefaultArtifactVersion(agentVersionString);
+
+                System.out.println(
```
tgianos: Looks like this was left in from debugging/development

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
```diff
@@ -0,0 +1,135 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: Wrong year

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
```diff
@@ -0,0 +1,135 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
+
+import java.util.List;
+
+/**
+ * Auto-configuration for the {@link AgentFilterService} default implementation.
```
tgianos: If this is an auto configuration you need to add it `src/main/resources/META-INF/spring.factories`

genie-web/src/main/java/com/netflix/genie/web/services/AgentFilterService.java
mprimi: Acccept/Reject stop the processing short. Whatever filter comes next won't be checked.

`INCONCLUSIVE` means, you pass this filter, you may move on to the next one.
Would `CONTINUE` be more clear?

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
mprimi: 👍

genie-web/src/main/java/com/netflix/genie/web/services/AgentFilterService.java
tgianos: I think going commit by commit hurts a little bit to get the full picture of how you were going to use this. I think `CONTINUE` would be more clear. It seems you're trying to go with a HTTP filter like mechanism. In which case "Accept" really is "continue" I think? I don't know that there needs to be a third option as we should evaluate all filters that are available and no short circuit unless one fails

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
mprimi: This way of accept/reject/continue is slightly more expressive than the 'all filters must pass' model.

For example: suppose i have a minimum version of `1.2.6` but would also like to allow `1.2.2`.
I can add `1.2.2` to the whitelist filter and it'll be let trough despite the minimum version.

I don't think in practice we'll need this kind of stuff. But when implementing filtering i just went instinctively for the pattern that I know to be more flexible.

mprimi:
Re-implementing with simplified ACCEPT/REJECT logic.

PR #817, opened 2019-01-18
Agent version filter

mprimi:
Introduce `handshake` gRPC protocol between Agent and server.

Provides an opportunity to the server to turn away clients running a version that is known to be incompatible or deprecated.
The filtering is extensible via Spring injection and the built-in metadata inspectors are responsive to dynamic changes of property values, allowing changing of filters without application re-deployment.

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
@@ -19,6 +19,7 @@ package com.netflix.genie.agent.execution.services.impl.grpc;
 
 import com.google.common.util.concurrent.ListenableFuture;
 import com.netflix.genie.agent.execution.exceptions.ChangeJobStatusException;
+import com.netflix.genie.agent.execution.exceptions.HandshakeException;
 import com.netflix.genie.agent.execution.exceptions.JobIdUnavailableException;
 import com.netflix.genie.agent.execution.exceptions.JobReservationException;
 import com.netflix.genie.agent.execution.exceptions.JobSpecificationResolutionException;
@@ -37,6 +38,8 @@ import com.netflix.genie.proto.ClaimJobError;
 import com.netflix.genie.proto.ClaimJobRequest;
 import com.netflix.genie.proto.ClaimJobResponse;
 import com.netflix.genie.proto.DryRunJobSpecificationRequest;
+import com.netflix.genie.proto.HandshakeRequest;
+import com.netflix.genie.proto.HandshakeResponse;
 import com.netflix.genie.proto.JobServiceGrpc;
 import com.netflix.genie.proto.JobSpecificationError;
 import com.netflix.genie.proto.JobSpecificationRequest;
@@ -79,6 +82,44 @@ class GRpcAgentJobServiceImpl implements AgentJobService {
         this.jobServiceProtoConverter = jobServiceProtoConverter;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void handshake(
+        final AgentClientMetadata agentClientMetadata
+    ) throws HandshakeException {
+        final HandshakeRequest request;
+        try {
+            request = jobServiceProtoConverter.toHandshakeRequest(agentClientMetadata);
+        } catch (final GenieConversionException e) {
+            throw new HandshakeException("Failed to construct request from parameters", e);
+        }
+
+        final HandshakeResponse response = handleResponseFuture(this.client.handshake(request));
+
+        switch (response.getType()) {
+            case ALLOWED:
+                log.info("Successfully shook hand with server");
+                break;
+
+            case REJECTED:
+                log.warn("Server rejected handshake with this agent");
+                throw new HandshakeException("Server rejected client: " + response.getMessage());
+
+            default:
+                throw new GenieRuntimeException(
+                    "Error during handshake: "
+                        + response.getType().name()
+                        + " - "
+                        + response.getMessage()
+                );
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public String reserveJobId(
         final AgentJobRequest agentJobRequest,
--- a/genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/exceptions/unchecked/GenieAgentRejectedExceptionSpec.groovy
+++ b/genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/exceptions/unchecked/GenieAgentRejectedExceptionSpec.groovy
@@ -1,0 +1,63 @@
+/*
+ *
+ *  Copyright 2019 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.common.internal.exceptions.unchecked
+
+import spock.lang.Specification
+
+/**
+ * Specifications for the {@link GenieAgentRejectedException} class.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+class GenieAgentRejectedExceptionSpec extends Specification {
+
+    def "Can construct"() {
+        String message = UUID.randomUUID().toString()
+        Throwable cause = new Exception()
+        GenieAgentRejectedException exception
+
+        when:
+        exception = new GenieAgentRejectedException()
+
+        then:
+        exception.getMessage() == null
+        exception.getCause() == null
+
+        when:
+        exception = new GenieAgentRejectedException(message)
+
+        then:
+        exception.getMessage() == message
+        exception.getCause() == null
+
+        when:
+        exception = new GenieAgentRejectedException(message, cause)
+
+        then:
+        exception.getMessage() == message
+        exception.getCause() == cause
+
+        when:
+        exception = new GenieAgentRejectedException(cause)
+
+        then:
+        exception.getMessage() == cause.toString()
+        exception.getCause() == cause
+    }
+}
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -23,6 +23,22 @@ that isn't listed in this map the default credentials configured will be used
 no value is specified the bucket is assumed to be in the same region as the Genie process.
 |
 
+|genie.agent.filter.enabled
+|If set to `true`, enables the built-in agent filter service. The filter behavior is controlled by other active `genie.agent.filter.*` properties.
+|
+
+|genie.agent.filter.version.minimum
+|(Dynamic) The minimum version an agent needs to be (e.g., `1.2.3`) in order to communicate with this server. The filter needs to be enabled for this to take effect.
+|
+
+|genie.agent.filter.version.blacklist
+|(Dynamic) A regex matched against agent version (e.g., `1\.2\..*`), matching agents are rejected. The filter needs to be enabled for this to take effect.
+|
+
+|genie.agent.filter.version.whitelist
+|(Dynamic) A regex matched against agent version (e.g., `1\.2\..*`), agents not matching are rejected. The filter needs to be enabled for this to take effect.
+|
+
 |genie.aws.credentials.role
 |The AWS role ARN to assume when connecting to S3. If this is set Genie will create a credentials provider that will
 attempt to assume this role on the host Genie is running on
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -10,336 +10,416 @@ The Spring properties described here are ones that we have overridden from Sprin
 
 ==== Genie Properties
 
+Properties marked 'dynamic' reflect change of property value in the environment happening at runtime.
+Whereas static properties values are bound during application startup and do not change after the application is up and running.
+
 |===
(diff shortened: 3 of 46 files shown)


genie-web/src/main/java/com/netflix/genie/web/util/AgentMetadataInspector.java
```diff
@@ -0,0 +1,36 @@
...
+ * Component that inspects an Agent client metadata and makes decision on whether it is allowed to proceed.
+ */
+public interface AgentMetadataInspector {
+    /**
+     * Perform inspection of an Agent client metadata.
+     *
+     * @param agentClientMetadata the agent client metadata
+     * @return the inspection outcome
+     */
+    InspectionReport inspect(@Valid AgentClientMetadata agentClientMetadata);
```
tgianos: FYI I think for Spring to activate the `@Valid` you need to put `@Validated` on the interface here.

genie-web/src/main/java/com/netflix/genie/web/util/InspectionReport.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public enum Decision {
+        /**
+         * Subject passed the inspection and is allowed to proceed.
+         */
+        ACCEPT,
+        /**
+         * Subject failed the inspection and is not allowed to proceed.
+         */
+        REJECT,;
```
tgianos: unnecessary trailing comma here

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -13,6 +13,23 @@ The Spring properties described here are ones that we have overridden from Sprin
 |===
 |Property |Description| Default Value
 
+
+|genie.agent.filter.enabled
+|If set to `true`, enables the built-in agent filter service. The filter behavior is controlled by other active `genie.agent.filter.*` properties.
+|
+
+|genie.agent.filter.version.minimum
+|(Dynamic) The minimum version an agent needs to be (e.g., `1.2.3`) in order to communicate with this server. The filter needs to be enabled for this to take effect.
```
tgianos: Rather than `(Dynamic)` why don't we add another column to the table that just says whether it's true or false?

genie-web/build.gradle
```diff
@@ -55,6 +55,9 @@ dependencies {
     compile("org.springframework.cloud:spring-cloud-starter-zookeeper")
     compile("org.springframework.integration:spring-integration-zookeeper")
     compile("org.springframework.retry:spring-retry")
+    compile("org.apache.maven:maven-artifact:3.5.4") {
```
tgianos: Move version to the parent `DependencyManagement` section

genie-web/src/main/java/com/netflix/genie/web/properties/AgentFilterProperties.java
```diff
@@ -0,0 +1,98 @@
...
+import javax.annotation.Nullable;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Properties for the {@link com.netflix.genie.web.services.AgentFilterService}.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@ConfigurationProperties
```
tgianos: I'd recommend putting the `PROPERTY_PREFIX` in here in the `prefix` field

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @param agentMetadataInspectorList the list of inspectors to consult
+     */
+    public AgentFilterServiceImpl(
+        final List<AgentMetadataInspector> agentMetadataInspectorList
+    ) {
+        this.agentMetadataInspectorList = agentMetadataInspectorList;
+    }
+
+    @Override
```
tgianos: Missing javadoc... not sure why checkstyle didn't fail

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+            log.debug(
+                "Inspector: {} inspected: {}, decision: {} ({})",
+                agentMetadataInspector,
+                agentClientMetadata,
+                decision.name(),
+                message
+            );
+
+            if (decision == InspectionReport.Decision.REJECT) {
+                return InspectionReport.newRejection("Rejected by: " + agentMetadataInspector + ": " + message);
```
tgianos: You're making an assumption as to the format of the `toString` for `agentMetadataInspector` here. Would it be better to use the class name? The debug statement as well. Either way you should probably assign the value you want to use once instead of potentially implicitly calling `toString` twice

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Fallback NOOP {@link AgentMetadataInspector} created only if no other
+     * bean of this type is present, to satisfy the dependency expressed by {@link AgentFilterService}.
+     *
+     * @return an noop inspector that always returns CONTINUE
+     */
+    @Bean
+    @ConditionalOnMissingBean(AgentMetadataInspector.class)
```
tgianos: Looking at this configuration class it seems impossible for this to even happen. It's not like your other beans have conditionals on them.

genie-web/src/test/groovy/com/netflix/genie/web/configs/GenieAgentFilterAutoConfigurationSpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+ *     limitations under the License.
+ *
+ */
+
+package com.netflix.genie.web.configs
+
+import com.netflix.genie.web.properties.AgentFilterProperties
+import spock.lang.Specification
+
+class GenieAgentFilterAutoConfigurationSpec extends Specification {
```
tgianos: Look into Spring's `ApplicationContextRunner`. I've used it in a few places to test these auto configuration classes. it lets you verify your conditionals work as expected etc. you can find some examples in the code base or I can show you.

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+ * service loads and uses:
+ * - Version whitelist: accept agent IFF version matches a given pattern
+ * - Version blacklist: reject agent whose version matches a given pattern
+ * - Minimum version: reject agent whose version is lower than a given minimum
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Configuration
+@ConditionalOnProperty(value = AgentFilterProperties.VERSION_FILTER_ENABLED_PROPERTY, havingValue = "true")
```
tgianos: I'm not sure but verify the default setting for `match if missing` is what you want or explicitly set it

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
+
+import java.util.List;
+
+/**
+ * Auto-configuration for the {@link AgentFilterService} default implementation.
```
tgianos: If you expect this to be an `AutoConfiguration` you need to add it to `src/main/resources/META-INF/spring.factories` so that Boot knows to pick it up.

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
```diff
@@ -0,0 +1,131 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos: Wrong year.

genie-common-internal/src/test/groovy/com/netflix/genie/common/internal/exceptions/unchecked/GenieAgentRejectedExceptionSpec.groovy
```diff
@@ -0,0 +1,63 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
```
tgianos: Year :-)

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentJobServiceImpl.java
```diff
@@ -48,25 +51,42 @@
...
         this.meterRegistry = meterRegistry;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void handshake(
+        @Valid final AgentClientMetadata agentClientMetadata
+    ) throws GenieAgentRejectedException {
```
tgianos: I think it would be good in here to add some metrics for success/failure/reason

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Added a commit that moves this to a separate column

genie-web/src/main/java/com/netflix/genie/web/properties/AgentFilterProperties.java
mprimi: I had to remove it, it would fail to bind.
That's because this class does not actually bind any property during startup.
The ENABLE property key is defined here, but used to enable the configuration.
And the rest of these properties are dynamic.

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentFilterServiceImpl.java
mprimi: Strange indeed, public method on a public class...

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
mprimi: Ah, indeed. Forgot to take this out.

genie-web/src/main/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfiguration.java
mprimi: `matchIfMissing` is default `false` which is what I want.

genie-web/src/main/java/com/netflix/genie/web/services/impl/AgentJobServiceImpl.java
mprimi: Done

mprimi:
@tgianos addressed all comments and ready for another pass.

genie-web/src/main/java/com/netflix/genie/web/services/AgentFilterService.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.validation.Valid;
+
+/**
+ * Service to block agent/clients that the server wants to refuse service to.
+ * For example, blacklist clients running a deprecated or incompatible versions.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public interface AgentFilterService {
```
tgianos: I think you're still missing `@Validated` here

genie-web/src/main/java/com/netflix/genie/web/util/InspectionReport.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ */
+
+package com.netflix.genie.web.util;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Representation of the outcome of an inspection performed by an {@link AgentMetadataInspector}.
```
tgianos: Missing author info if you care

genie-web/src/test/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+
+/**
+ * Tests for behavior of {@link GenieAgentFilterAutoConfiguration}.
+ *
+ * @author tgianos
```
tgianos: Wrong author

genie-web/src/test/java/com/netflix/genie/web/configs/GenieAgentFilterAutoConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+@Category(IntegrationTest.class)
+public class GenieAgentFilterAutoConfigurationIntegrationTest {
+
+    /**
+     * Test expected context when configuration is enabled.
+     */
+    @Test
+    public void testContextIfEnabled() {
+        new ApplicationContextRunner()
```
tgianos: Just an FYI you can make the part with the `AutoConfigurations.of` common member of the class to reuse. It all acts as a builder so anything you add/modify within each test method will build on the common. It looks like this part is common to all the tests where you set only this auto configuration

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
```diff
@@ -79,6 +82,43 @@
...
+        }
+
+        final HandshakeResponse response = handleResponseFuture(this.client.handshake(request));
+
+        switch (response.getType()) {
+            case ALLOWED:
+                log.info("Successfully shook hand with server");
+                break;
+
+            case REJECTED:
```
tgianos: Did you want to log something here or just relying on something upstream logging it?

PR #821, opened 2019-01-24
fix improve panel

Arhell:


tgianos:
Thanks @Arhell . Fixes #822

PR #825, opened 2019-01-29
Upgrade to Spring Boot 2.1

tgianos:
Updates to Spring Framework 5.1 transitively and Spring Cloud Greenwich.RELEASE which is the cloud release which supports Boot 2.1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -144,7 +144,7 @@ configure(javaProjects) {
             //       For now staying on 1.0.3 as not really thrilled with adding yet another maven repo
             dependency("org.springframework.security.extensions:spring-security-saml2-core:1.0.3.RELEASE")
             dependency("org.springframework.security.oauth.boot:spring-security-oauth2-autoconfigure:2.0.1.RELEASE")
-            dependencySet(group: "org.springframework.statemachine", version: "2.0.2.RELEASE") {
+            dependencySet(group: "org.springframework.statemachine", version: "2.0.3.RELEASE") {
                 entry "spring-statemachine-core"
                 entry "spring-statemachine-test"
             }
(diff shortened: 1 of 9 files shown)


genie-web/src/test/groovy/com/netflix/genie/web/services/impl/DiskJobFileServiceImplSpec.groovy
```diff
@@ -24,17 +24,17 @@ import org.apache.commons.codec.digest.DigestUtils
...
 import java.nio.charset.Charset
 import java.nio.file.Files
 import java.nio.file.Paths
+
 /**
  * Specifications for the {@link DiskJobFileServiceImpl} class.
  *
- * @author tgianos
- * @since 4.0.0
+ * @author tgianos* @since 4.0.0
```
mprimi: Missing line

build.gradle
```diff
@@ -91,7 +91,7 @@ configure(javaProjects) {
             dependency("commons-httpclient:commons-httpclient:3.1")
             dependency("commons-io:commons-io:2.6")
             dependency("commons-validator:commons-validator:1.6")
-            dependencySet(group: "io.grpc", version: "1.12.0") {
```
mprimi: Is this really necessary? Do we run into some kind of trouble by leaving it higher than recommended?

If i remember correctly, we were getting some very weird concurrency bug that would corrupt the network buffers, it was a known issue, and upgrading to 1.12 fixed it.

tgianos: I don't know that it is. You didn't mention that issue yesterday when I asked. It was easier to back down than battle dependency hell internally with the versions but I can remove it and deal with that if there is some functional issue that 1.12 fixed. Since the build passed I didn't think it was an issue. I'll remove this and see if I can resolve the internal versioning.

tgianos:
@chali Backing Gradle back down here until [Gradle Git Publish Issue](https://github.com/ajoberstar/gradle-git-publish/issues/65) is addressed

tgianos:
@twicksell FYI in case you have any issues here

twicksell:
@tgianos lgtm 👍

PR #826, opened 2019-01-29
Reusable job archive functionality

tgianos:
Primarily this PR contains commits towards serving job directory files out of the archive location after a job is complete. A lot of refactoring is done to make code that was exclusive to the `agent` re-usable in the `agent` and `web` code bases.

Functionality to create a job directory manifest in the archive location for quick reference is also included.

Downgrading gRPC dependency from 1.12.0 to 1.10.1 is done to maintain compatibility with internal dependencies and avoid dependency conflicts with transitive clients


genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/JobDirectoryManifest.java
```diff
@@ -0,0 +1,407 @@
...
+            String mimeType = null;
+            if (!directory) {
+                try (InputStream data = Files.newInputStream(entry, StandardOpenOption.READ)) {
+                    md5 = DigestUtils.md5Hex(data);
+                } catch (final IOException ioe) {
+                    // For now MD5 isn't critical or required so we'll swallow errors here
+                    log.error("Unable to create MD5 for {} due to error", entry, ioe);
+                }
+
+                try (TikaInputStream inputStream = TikaInputStream.get(entry)) {
```
mprimi: Spring is currently doing file content type detection, is there no way to reuse their code, rather than this new dependency?
Just curious, if this works I don't care much what we end up using.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/services/impl/JobArchiveServiceImpl.java
```diff
@@ -52,8 +56,22 @@ public JobArchiveServiceImpl(final List<JobArchiver> jobArchivers) {
      */
     @Override
     public void archiveDirectory(final Path directory, final URI target) throws JobArchiveException {
+        // Create a manifest and write it to the root of the job directory
```
mprimi: I would not put it in root, since it's an "internal" file, i'd put it in `genie/`.

tgianos: A fair point. I thought about this too. I figured since it would only live in the archive location and never be seen by the user it didn't really matter and I wasn't sure we were keeping the `genie` directory around in 4. I can put it there for now. We can move it later if need be.

tgianos: I've put this in `genie/`. I don't really love the implementation but I want to get this merged. Will likely want to revisit at some point in the near future.

PR #827, opened 2019-02-04
Feature to serve job directory files from archive location if available

tgianos:
The primary feature in this PR is to make it so if a job is complete (whether it was run via V3 API or V4 agent) and the job directory was archived the system can serve the directory resources from that archive location. This will enable us to continue serving job resources long after the logs have been deleted from disks or nodes have been replaced.

Other commits exist to support this change.

Jobs launched via V3 API call that are still in an active state will still have files transparently served off of disk on the node executing the job.

V4 agent jobs which are still active will currently return a 404 to all requests but eventually this will be supported by reverse proxy over the gRPC connection.

There are still several rough edges in the implementation in terms of handling errors and returning correct HTTP response codes however want to get this committed so it can be tested/iterated on.


genie-web/src/main/java/com/netflix/genie/web/controllers/ControllerUtils.java
```diff
@@ -18,6 +18,7 @@
 package com.netflix.genie.web.controllers;
```
mprimi: Extra 'to' in the commit message?

genie-web/src/test/java/com/netflix/genie/web/controllers/ControllerUtilsUnitTests.java
```diff
@@ -77,4 +77,25 @@ public void canGetRemainingPath() {
             .thenReturn("/api/v3/jobs/{id}/output/**");
         Assert.assertThat(ControllerUtils.getRemainingPath(request), Matchers.is("stdout"));
     }
+
```
mprimi: Maybe test corner cases just to freeze the behavior?
i.e. empty 'path', request not containing path, ...

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/JobDirectoryManifest.java
```diff
@@ -262,6 +263,10 @@ private ManifestEntry buildEntry(
             final boolean directory
         ) throws IOException {
             final String path = this.root.relativize(entry).toString();
+            final Path fileName = entry.getFileName();
+            final String name = fileName == null
```
mprimi: Should this error out instead of silently adding an entry with an empty name?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/JobDirectoryManifest.java
```diff
@@ -319,6 +315,24 @@ private ManifestEntry buildEntry(
                 children
             );
         }
+
+        private String getMimeType(final String name, final Path path) {
+            // TODO: Loose conventions in here mostly for backwards compatibility but if there's empty
```
mprimi: We could/should move this `filename`->`mime type` map into configuration..

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
```diff
@@ -602,10 +583,48 @@ public void updateJobStatus(
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Transactional(readOnly = true)
     public Optional<JobStatus> getJobStatus(@NotBlank(message = "Job id is missing and is required") final String id) {
+        // TODO: Need to decide on consistency of API. Do we throw exception? Return Optional? Runtime or Checked?
+        //       getJobArchiveLocation throws exception cause the underlying return type is already optional. So
+        //       do we return Optional<Optional<String>> that's really weird
         return this.jobRepository.findByUniqueId(id, JobStatusProjection.class).map(JobStatusProjection::getStatus);
```
mprimi: The other two here are throwing `GenieNotFoundException`. What's the reason for not doing so for this one?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/JobDirectoryManifest.java
tgianos: It was really put in to avoid the extreme edge case of the `Path` returning `null` and then calling `toString()` on it. The reason I went with empty string is cause that matches the behavior of the `path` attribute above where if you relativize the root to itself it returns empty string. The file name was really only used in this case for display in the UI so it's not critical.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/JobDirectoryManifest.java
tgianos: That's a good idea! I'll add a TODO so can remember to come back to it

genie-web/src/main/java/com/netflix/genie/web/jpa/services/JpaJobPersistenceServiceImpl.java
mprimi: At least in this file, everything else is throwing `GenieJobNotFoundException`.
I know we need some consistency and a strategy at project level, but everything in this file is behaving the same,  I think having a single method behave differently for no apparent reason is more confusing, no?

genie-web/src/test/java/com/netflix/genie/web/controllers/ControllerUtilsUnitTests.java
tgianos: Added another commit with some improved functionality

PR #832, opened 2019-02-27
Add spring-boot-autoconfigure-processor

tgianos:
Per the [docs](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-developing-auto-configuration.html#boot-features-custom-starter-module-autoconfigure) this annotation processor speeds up startup time by creating a metadata file of the conditions on auto configuration classes at startup time. Might help agent startup time so no harm in adding it.

### DIFF
--- a/genie-client/build.gradle
+++ b/genie-client/build.gradle
@@ -11,13 +11,13 @@ dependencies {
     api("com.github.fge:json-patch")
     api("com.squareup.okhttp3:okhttp")
     api("com.squareup.retrofit2:retrofit")
+    api("com.squareup.retrofit2:converter-jackson")
 
     /*******************************
      * Implementation Dependencies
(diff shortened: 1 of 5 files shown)


genie-client/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
     api("com.github.fge:json-patch")
     api("com.squareup.okhttp3:okhttp")
     api("com.squareup.retrofit2:retrofit")
+    api("com.squareup.retrofit2:converter-jackson")
```
mprimi: The message says 'moved', but it seems this was added

tgianos: Bah you're right I must have copied instead of cut it out cause it's still in the implementation section. Let me change that.

PR #837, opened 2019-03-13
Move default property handling to EnvironmentPostProcessor's

tgianos:
The `@PropertySource` "fix" in PR #834 didn't do what it was intended to do due to how Spring would create an environment before those property sources were evaluated. So they weren't accounted for during application startup. This commit effectively reverts that change and moves it to EnvironmentPostProcessor implementation which loads the properties before auto configs are evaluated. Additionally this had the benefit of letting us switch back to using YAML files for the properties which (IMO) is much more readable and consistent with our other files.


genie-web/src/main/java/com/netflix/genie/web/configs/processors/DefaultPropertiesPostProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+ * @author tgianos
+ * @since 4.0.0
+ */
+public class DefaultPropertiesPostProcessor implements EnvironmentPostProcessor {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void postProcessEnvironment(final ConfigurableEnvironment environment, final SpringApplication application) {
```
mprimi: 5 string constants in this method could be pulled out

tgianos:
@mprimi I've addressed your comments from the review and added tests for `agent` and `web` properties. Have a look.

PR #838, opened 2019-03-13
Misc improvements

mprimi:


mprimi:
@tgianos breaking my giant PR in 2. 
These are general improvements, the rest is log streaming.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/JobDirectoryManifest.java
```diff
@@ -223,11 +222,13 @@ public long getTotalSizeOfFiles() {
 
         private final Path root;
         private final ImmutableMap.Builder<String, ManifestEntry> builder;
+        private final Metadata metadata;
         private final TikaConfig tikaConfig;
 
         ManifestVisitor(final Path root, final ImmutableMap.Builder<String, ManifestEntry> builder) throws IOException {
             this.root = root;
             this.builder = builder;
+            this.metadata = new Metadata();
```
tgianos: My only drawback with this is now we're creating one of these every time we create a visitor rather than just a shared one for the entire app. So it's kind of a runtime performance downgrade vs. testing convenience. Won't break the world obviously but we do invoke this visitor relatively often on the Agent to push manifests right?

mprimi: That is indeed the tradeoff. Tho if you look at `Metadata` and it's constructor, it is pretty cheap. only creates an empty map member.

PR #839, opened 2019-03-14
Segment tests into facets

tgianos:
Opening PR for feedback


tgianos:
@mprimi I redid the commits to try to break out the reformatting from the facet restructuring. Take a look if you get a chance

mprimi:
@tgianos There's stil a lot in that third commit that is unrelated to what the commit message says.. But ultimately i agree this is is not worth obsessing over. Just hard to review what actually changed as result of faceting because it's drowned in hundreds of other changes.

PR #840, opened 2019-03-19
UI enhancements

cabhishek:


genie-web/src/main/resources/static/scripts/Job.js
```diff
@@ -77,19 +77,19 @@ export default class Job extends Page {
...
-    return [
+      },
       { label: "Tags", name: "tag", value: "", type: "input" },
       { label: "Cluster Name", name: "clusterName", value: "", type: "input" },
       { label: "Cluster ID", name: "clusterId", value: "", type: "input" },
       { label: "Command Name", name: "commandName", value: "", type: "input" },
       { label: "Command ID", name: "commandId", value: "", type: "input" },
       { label: "Grouping", name: "grouping", value: "", type: "input" },
-      { label: "Grouping Instance", name: "groupingInstance", value: "", type: "input" },
+      {
```
mprimi: Any reason for changing this?

cabhishek: Its the code formatter. We are using [Prettier](https://prettier.io/) and its the default formatting for a complex object.

cabhishek:
> Change LGTM, only comment is about the commit message.
> We try to comply to [the 7 rules](https://chris.beams.io/posts/git-commit/#seven-rules) in this repo.
> 
> A good commit message could be:
> 
> ```
> UI enhancements
> 
>  * Change default from 25 results to 100
>  * Add 'command' column in job view
>  * Default show all search fields in job view
> ```
> Or, even better, create 3 commits, each one doing just 1 thing (and the subject for each can be the bullet points above, no need for a body).

I updated the commit message. I wasn't sure if you'd like one commit/PR with bundled small features or individual commit/PR. I'll follow the later going forward and open a PR for the single ticket requesting the UI feature.

genie-web/src/main/resources/static/scripts/Job.js
mprimi: It decided to format that one line while leaving all the other lines above untouched? :-/

mprimi:
@cabhishek 

One PR for everything is great

One commit for everything is ok. One commit per change is better (easier to understand/review/cherry-pick)

genie-web/src/main/resources/static/scripts/Job.js
cabhishek: That's cause Prettier defaults to line width of 80 and the last object if written on one line has a length of 87. And All other lines are under 79.

genie-web/src/main/resources/static/scripts/Page.js
```diff
@@ -53,7 +53,7 @@ export default class Page extends React.Component {
 
   loadPageData(query) {
     if ($.isEmptyObject(query)) {
-      query = { size: 25 };
+      query = { size: 100 };
```
mprimi: @cabhishek I'm assuming this will apply to all pages, right? (i.e. not just jobs).

@tgianos given the lazy loading of tags we saw yesterday, maybe we should hold off on this, WDYT?

genie-web/src/main/resources/static/scripts/components/JobDetails.js
```diff
@@ -46,9 +51,19 @@ export default class JobDetails extends React.Component {
...
-        .done(command => {
-          this.setState({ command });
+
+      $.when(
+        fetch(job._links.cluster.href),
+        fetch(job._links.command.href),
+        fetch(job._links.applications.href)
+      )
+        .done((clusters, commands, applications) => {
+          console.log(applications);
```
mprimi: Forgot to remove console log?

genie-web/src/main/resources/static/scripts/components/JobDetails.js
```diff
@@ -46,9 +51,19 @@ export default class JobDetails extends React.Component {
...
+          this.setState({
+            cluster: clusters[0],
+            command: commands[0],
+            applications: applications[0]
```
mprimi: cluster and command are a single entity, but application is an array.
Seems strange they get all treated the same here.

genie-web/src/main/resources/static/scripts/components/JobDetails.js
cabhishek: @mprimi The console log was removed in the next commit. Maybe when we squash all the commit this won't be an issue?

genie-web/src/main/resources/static/scripts/Page.js
tgianos: I think the original request from @rdblue, amongst others, mostly pertained to jobs which aren't impacted by this since tags aren't included in the top level page.

However, @cabhishek is the preference saved if a user selects `100` as the load size? If so these users could just set it themselves. If not I don't necessarily see much harm leaving this at 100 for the jobs page since the search there already returns the proper projection.

genie-web/src/main/resources/static/scripts/components/JobDetails.js
cabhishek: @mprimi This is specific to `jquery`'s `when` interface. The `applications[0]` returns an array hence its assgined to the key `applications`. Where as `commands[0]` returns an object and hence the key is named `command`.

genie-web/src/main/resources/static/scripts/Page.js
cabhishek: @tgianos Preference is not saved in a cookie. But the `size` param is present in the URL and persisted across different server requested e.g during pagination or during link share.

cabhishek: @mprimi yes, all pages now have size of `100` by default.

PR #844, opened 2019-04-03
Fix H2 migration script

tgianos:
Also fix bug with default properties files still existing and messing up `genie-app` execution.


genie-web/src/main/java/com/netflix/genie/web/jpa/utils/package-info.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ */
+
+/**
+ * Utilities for working with JPA or the Genie databases in general.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+package com.netflix.genie.web.jpa.utils;
```
mprimi: How about the usual annotations we have in other modules?

tgianos: Forgot it. Added. Thanks.

PR #845, opened 2019-04-04
Fix bug where cluster and command criterion weren't backported to V3

tgianos:
Job Requests representations properly

Id and Name parameters were being lost. This seems to have been a miss in conversion from the database record to the V3 representation.

Added a test to the JobSearchService integration tests which should cover the case. It was broken when I started this work and now it passes.

### DIFF
--- a/genie-web/src/integTest/java/com/netflix/genie/web/jpa/services/JpaJobSearchServiceImplIntegrationTest.java
+++ b/genie-web/src/integTest/java/com/netflix/genie/web/jpa/services/JpaJobSearchServiceImplIntegrationTest.java
@@ -23,6 +23,7 @@ import com.google.common.collect.Sets;
 import com.netflix.genie.common.dto.Application;
 import com.netflix.genie.common.dto.Job;
 import com.netflix.genie.common.dto.JobMetadata;
+import com.netflix.genie.common.dto.JobRequest;
 import com.netflix.genie.common.dto.JobStatus;
 import com.netflix.genie.common.dto.search.JobSearchResult;
 import com.netflix.genie.common.exceptions.GenieException;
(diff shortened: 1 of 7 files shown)


genie-web/src/integTest/java/com/netflix/genie/web/jpa/services/JpaJobSearchServiceImplIntegrationTest.java
```diff
@@ -324,10 +325,24 @@ public void canGetJobStatus() throws GenieException {
      */
     @Test
     public void canGetJobRequest() throws GenieException {
+        final JobRequest job1Request = this.service.getJobRequest(JOB_1_ID);
         Assert.assertThat(
-            this.service.getJobRequest(JOB_1_ID).getCommandArgs().orElseThrow(IllegalArgumentException::new),
+            job1Request.getCommandArgs().orElseThrow(IllegalArgumentException::new),
             Matchers.is("-f query.q")
         );
+        Assert.assertThat(
```
mprimi: Why not check all of them for equality?

(currently a mix of size checking and equality for just one particular field)

tgianos: Because I was modifying to test for the bug not completely re-write the tests for everything

mprimi: Right but you check one of the criteria with:
`Matchers.is(Sets.newHashSet("genie.id:cluster1", "genie.name:h2query", "sla", "yarn"))`
but not the rest of the bunch... why not check all of them?

tgianos: eh fair. Probably laziness and the knowledge this `JobSearchService` will be deleted but I'll add the rest of them the criteria to be comprehensive.

tgianos: and split your precious commits :-)

PR #848, opened 2019-04-05
Active job aggregate metrics

mprimi:


genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaJobRepository.java
```diff
@@ -84,4 +87,26 @@
...
     // TODO: Explore deleteFirst{N}ByCreatedBefore
     Slice<IdProjection> findByCreatedBefore(@NotNull final Date date, @NotNull Pageable pageable);
+
+    /**
+     * Produce a count of jobs in the given states for each user.
+     *
+     * @param statuses the job statuses to filter by
+     * @return a set of {@link UserJobAmountAggregateProjection}
+     */
+    @Query("SELECT j.user, COUNT(j) FROM JobEntity j WHERE j.status IN (:statuses) GROUP BY j.user")
```
tgianos: Any particular reason to break this into two different queries and not just make one query that fetches both pieces of information count/memory?

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -364,6 +364,10 @@ number of CPU cores x 2 + 1
 to the number of CPU cores x 2 + 1
 |1
 
+|genie.tasks.userMetrics.enabled
```
tgianos: might want to make this key `user-metrics` now to save yourself problems in master with the changes to property bindings spring did

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/UserMetricsTask.java
```diff
@@ -0,0 +1,148 @@
...
+    public GenieTaskScheduleType getScheduleType() {
+        return GenieTaskScheduleType.FIXED_RATE;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public long getFixedRate() {
+        return 5 * 1000; // 5 seconds
```
tgianos: Should this be a property?

genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaJobRepository.java
mprimi: Didn't even think of this, would make this and downstream code simpler... How about going in with this for `3.3.x`, since it's close to EOL anyway, and do it as single query in `master`?

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Just kept going with convention used by other similar switches..

genie-core/src/main/java/com/netflix/genie/core/jpa/repositories/JpaJobRepository.java
tgianos: Seems fine to me.

genie-docs/src/docs/asciidoc/_properties.adoc
tgianos: Those all had to be flipped if they were camel case in 4.x. If you're ok with making sure you do that then sure leave it like this. Just pointing it out that you could just do it once for a cleaner cherry-pick if you're going for that.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/UserMetricsTask.java
mprimi: Done

mprimi:
@tgianos  Added a properties class (and test) for this task. Ready for a quick look and merge once checks pass.

PR #849, opened 2019-04-08
UI Improvements

tgianos:
Cherry picks all commits from #840 into master.

### DIFF
--- a/genie-ui/src/main/web/scripts/utils.js
+++ b/genie-ui/src/main/web/scripts/utils.js
@@ -1,6 +1,6 @@
 import $ from "jquery";
-import moment from "moment";
-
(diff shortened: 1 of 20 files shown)


genie-ui/src/main/web/scripts/utils.js
```diff
@@ -1,6 +1,6 @@
 import $ from "jquery";
-import moment from "moment";
-
+import moment from "moment-timezone";
+window.moment = moment;
```
cabhishek: @tgianos ugh, this is not needed. The assignment was for testing.

cabhishek: Fixed - https://github.com/Netflix/genie/pull/850

PR #852, opened 2019-04-19
Add Agent connection metrics and health indicator

tgianos:


genie-web/src/main/java/com/netflix/genie/web/services/JobMetricsService.java
```diff
@@ -37,4 +37,13 @@
...
      */
     int getUsedMemory();
+
+    /**
+     * Get the number of agents currently connected to this server.
+     *
+     * @return The number of agents connected to the server
+     */
+    default long getNumActiveAgents() {
+        return 0L;
```
mprimi: Why the default value for this one?

tgianos: Just to make it easier if something didn't want to implement all of them. It was a remnant of when I had it in the `JobMetricsService` and didn't want to implement it in 2 or 3 places. Doesn't really harm anything.

PR #853, opened 2019-04-22
Switch from 'ps' to 'kill -0' to check on children processes

mprimi:
Using 'ps' to check for process termination can sometimes produce false-negatives.
I.e. returning with non-zero exit code when the process is actually still running.
This situation can lead Genie to mark a job as failed when it was actually still running.

Switch to using 'kill -0' as per POSIX recommendation.

I.e. returning with non-zero exit code when the process is actually still running.
This situation can lead Genie to mark a job as failed when it was actually still running.

Switch to using  as per POSIX recommendation.
(this can still produce false negatives, such as when permissions are lacking, or when kill fails to launch due to resource exhaustion)

### DIFF
--- a/genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
+++ b/genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
@@ -45,22 +45,34 @@ public class UnixProcessChecker implements ProcessChecker {
     /**
      * Constructor.
      *
-     * @param pid      The process id to check.
-     * @param executor The executor to use for generating system commands.
-     * @param timeout  The time which after this job should be killed due to timeout
+     * @param pid           The process id to check.
+     * @param executor      The executor to use for generating system commands.
+     * @param timeout       The time which after this job should be killed due to timeout
+     * @param checkWithSudo Whether the checker requires sudo
      */
-    public UnixProcessChecker(@Min(1) final int pid, @NotNull final Executor executor, @NotNull final Date timeout) {
+    UnixProcessChecker(
+        @Min(1) final int pid,
+        @NotNull final Executor executor,
+        @NotNull final Date timeout,
+        final boolean checkWithSudo
+    ) {
         if (!SystemUtils.IS_OS_UNIX) {
             throw new IllegalArgumentException("Not running on a Unix system.");
         }
 
         this.executor = executor;
 
-        // Using PS for now but could instead check for existence of done file if this proves to have bad performance
-        // send output to /dev/null so it doesn't print to the logs
-        this.commandLine = new CommandLine("ps");
-        this.commandLine.addArgument("-p");
-        this.commandLine.addArgument(Integer.toString(pid));
+        // Use POSIX compliant 'kill -0 <PID>' to check if process is still running.
+        if (checkWithSudo) {
+            this.commandLine = new CommandLine("sudo");
+            this.commandLine.addArgument("kill");
+            this.commandLine.addArgument("-0");
+            this.commandLine.addArgument(Integer.toString(pid));
+        } else {
+            this.commandLine = new CommandLine("kill");
+            this.commandLine.addArgument("-0");
+            this.commandLine.addArgument(Integer.toString(pid));
+        }
 
         this.timeout = new Date(timeout.getTime());
         this.dateFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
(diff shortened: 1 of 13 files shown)


mprimi:
@tgianos re-worked this PR to do a few things better:
 - use sudo if we are using sudo to launch
 - Abstract `UnixProcessChecker` from classes using `ProcessChecker`
Can you take another look?

genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
```diff
@@ -80,4 +92,32 @@ public void checkProcess() throws GenieTimeoutException, ExecuteException, IOExc
             );
         }
     }
+
+    /**
+     * Factory for {@link ProcessChecker} for UNIX systems.
```
tgianos: since it's a factory why would it be specific to UNIX systems? I'm probably missing some of your logic but I would think this would be some standalone factory class outside the scope of just the UnixProcesChecker that would create process checkers on a per OS basis. Why would we have a factory per implementation?

genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
```diff
@@ -80,4 +92,32 @@ public void checkProcess() throws GenieTimeoutException, ExecuteException, IOExc
...
+            this.executor = executor;
+            this.checkWithSudo = checkWithSudo;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public ProcessChecker get(final int pid, final Date timeout) {
+            return new UnixProcessChecker(pid, this.executor, timeout, this.checkWithSudo);
```
tgianos: Shouldn't your factory have a check here if the OS is Unix as this process checker is dependent on that? Throw exception if on some other OS?

genie-web/src/main/java/com/netflix/genie/web/configs/JobConfig.java
```diff
@@ -192,4 +197,28 @@ public WorkflowTask jobKickoffTask(
...
+     * @return a {@link ProcessChecker.Factory}
+     */
+    @Bean
+    @Autowired
+    @ConditionalOnMissingBean(ProcessChecker.Factory.class)
+    public ProcessChecker.Factory processCheckerFactory(
+        final Executor executor,
+        final JobsProperties jobsProperties
+    ) {
+        if (SystemUtils.IS_OS_UNIX) {
```
tgianos: Per my earlier comments now that I'm down here I kind of see what you're trying to do which is to fail the app coming up if it isn't supported. I still think this could be done cleaner in a centralized factory just in the constructor of said factory if the OS isn't supported.

genie-web/src/main/java/com/netflix/genie/web/configs/JobConfig.java
```diff
@@ -192,4 +197,28 @@ public WorkflowTask jobKickoffTask(
...
+
+    /**
+     * Create a {@link ProcessChecker.Factory} suitable for UNIX systems.
+     *
+     * @param executor       The executor where checks are executed
+     * @param jobsProperties The jobs properties
+     * @return a {@link ProcessChecker.Factory}
+     */
+    @Bean
+    @Autowired
```
tgianos: You don't need `@Autowired` on `@Bean` functions

genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
mprimi: Isn't that already the case?
All users are just depending on and using `ProcessChecker.Factory`.

This implementation is only able to produce UNIX process checkers, but nobody needs to know.

tgianos: Then there's no value in a factory right?

genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
mprimi: When this factory is asked to produce a process checker, it does.
Checking wether this is the right environment to produce such an instance seems like overreach to me.

The checks do exist in other places:
 - In configuration, we only create this factory if the OS is unix
 - The process checker itself refuses to run unless the os is unix

genie-web/src/main/java/com/netflix/genie/web/configs/JobConfig.java
mprimi: My thought is that it is better to fail early (during app startup) than fail in weird ways later (app is up and accepted jobs, but now we find out we don't know how to check on them)

genie-web/src/main/java/com/netflix/genie/web/configs/JobConfig.java
mprimi: I just kept consistent with the "style" of the rest of this file...
I'll remove all unnecessary `@Autowired`

genie-core/src/main/java/com/netflix/genie/core/util/UnixProcessChecker.java
mprimi: Before we were calling `new UnixProcessChecker()`. in different parts of the code.
Now we call `ProcessChecker.Factory.get()` in those places.

We inject the factory, we configure it externally, etc. Why is this no value?

tgianos: In my opinion in this design your "factory"should be more of a factory method than a factory class cause your factory only produces one thing and is tied to the parent class it's coming from. I'd break it out into a totally separate thing but that's just me. Like i said merge it if you're comfortable with it.

mprimi: Factory is just meant to decouple and encapsulate object creation. 
Nothing says it needs to create different types of objects.

The factory interface i added is generic.
The implementation is specific but it can be swapped out, changed later, etc.

PR #856, opened 2019-04-29
Agent job reaping

mprimi:


mprimi:
@tgianos only the last 2 commits in this PR are relevant, the first 3 will disappear from here once the other PR goes in.

genie-web/src/main/java/com/netflix/genie/web/properties/AgentCleanupProperties.java
```diff
@@ -0,0 +1,52 @@
...
+    public static final String PROPERTY_PREFIX = "genie.tasks.agent-cleanup";
+
+    /**
+     * The property that determines if the {@link com.netflix.genie.web.tasks.leader.AgentJobCleanupTask} is enabled.
+     */
+    public static final String ENABLED_PROPERTY = PROPERTY_PREFIX + ".enabled";
+
+    private boolean enabled = true;
+
+    private long refreshInterval = 10_000;
```
tgianos: if they're longs I'd suggest putting the `L` after them but not a huge deal obviously. The compiler probably does it anyway.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/AgentJobCleanupTask.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Set;
+
+/**
+ * Leader task that cleans up jobs whose agent crashed or disconnected.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public class AgentJobCleanupTask extends LeadershipTask {
```
tgianos: It would be nice if you published some metrics on this. It doesn't have to be part of this PR but similar to how we track "lost jobs" this would be good information to have.

mprimi: Yes i was gonna, but wanted to make sure this goes in today so left it in a (WIP) commmit.
I'm adding 
 - Gauge with currently disconnected jobs (size of the map)
 - Counter of jobs marked killed
Anything else you can think of?

tgianos: Counters may be kind of screwy as the interval is so long apart we'll just see random bursts. I think I used a gauge for that with lost jobs then each value reported stayed until the next iteration.

mprimi: They're not random bursts.. we increment the count when a job is killed.
So yes, it is bursty but it's what's actually happening.

tgianos: Meaning there is no information we can really glean from a counter. It'll just be spikes at some rate. Counters are only valuable when there is changes in rates over time. This will go from 0 to some number and we'll just end up normalizing it to the raw number anyway by dividing by the duration to get a value. So to me the count serves no real purpose but we can get the same info from a counter as a gauge we'll just have to find the spikes.

mprimi: I don't get it. We're just counting the number of events (jobs marked failed in this case).

What's wrong with this approach?

PR #857, opened 2019-04-29
Tweak log levels

tgianos:
In order to still get relevant entry point information but not switch the overall logs to debug this PR primarily changes all the controller top level logs to be level `info`


genie-web/src/main/java/com/netflix/genie/web/controllers/ApplicationRestController.java
```diff
@@ -135,7 +135,7 @@ public ApplicationRestController(
     @DeleteMapping
     @ResponseStatus(HttpStatus.NO_CONTENT)
     public void deleteAllApplications() throws GenieException {
-        log.debug("Delete all Applications");
+        log.warn("Called to delete all Applications");
```
mprimi: Not worth changing, but I think this should be INFO. It's nothing out of the extraordinary, it's just an entry point.
Same for the ones below.

PR #860, opened 2019-04-30
Logging and metrics

mprimi:


genie-web/src/main/java/com/netflix/genie/web/tasks/leader/AgentJobCleanupTask.java
```diff
@@ -39,26 +42,40 @@
 @Slf4j
 public class AgentJobCleanupTask extends LeadershipTask {
     private static final String STATUS_MESSAGE = "Agent AWOL for too long";
-    private final Map<String, Instant> awolJobDeadlines = Maps.newHashMap();
+    private static final String DISCONNECTED_AGENT_JOBS_TIMEOUT_METRIC_NAME = "";
```
tgianos: I assume this isn't supposed to be blank... same as the next one

mprimi:
Codecov is complaining about reduced coverage on 3 files that i didn't touch...

PR #865, opened 2019-05-07
Change command args string argument to not split on whitespace

tgianos:
The command args field was modified to be a List<String> in the data model and in the database in order to better support escaping and other issues with bash we've had in the past. However to enable this we were splitting naively on whitespace which could have unintended consequences. This commit reverts that behavior to treat anyone calling the deprecated "string" builder API for the JobRequest to just treat that as a single argument blob. The List<String> builder method still will treat it correctly and join it with a space on a call to get. This should behave as it has in the past with 3 until we have a V4 API. The agent will continue to behave the right way as it only supports List<String> command arguments.


mprimi:
`git am <`[0001-Example-StringTokenizer.patch.txt](https://github.com/Netflix/genie/files/3158838/0001-Example-StringTokenizer.patch.txt)

PR #869, opened 2019-05-11
Add ClusterChecker metrics for remote healthchecks

mprimi:


genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -406,7 +406,7 @@ determined by Zookeeper or other mechanism via Spring
 |genie.tasks.cluster-checker.healthIndicatorsToIgnore
 |The health indicator groups from the actuator /health endpoint to ignore when determining if a node is lost or not as
 a comma separated list
-|memory,genieMemory,discoveryComposite
+|genieMemory,mail,genieAgent
```
tgianos: Just an FYI that this list may be different internally or set to a different property at least. Probably will want to double check before pushing

PR #871, opened 2019-05-14
Reorganize package com.netflix.genie.web.jpa

mprimi:
Rename to com.netflix.genie.web.data.
Lower `jpa` in the hierarchy as it contains implementations.


genie-web/src/test/java/com/netflix/genie/web/data/services/jpa/package-info.java
```diff
@@ -17,8 +17,8 @@
  */
 
 /**
- * Classes to test the jpa implementations of the Genie services.
+ * Classes to test the data implementations of the Genie services.
```
tgianos: this is were it gets a bit weird (not that it's any less clear than before). Perhaps we should move some service interfaces that are only for data into the data package and out of services.

mprimi: I was going to suggest the opposite.. move all services in the services package, wether they access data or not.

PR #876, opened 2019-05-24
Fix user resource usage metrics

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/listeners/LoggingListener.java
```diff
@@ -93,7 +92,7 @@ public void eventNotAccepted(final Message<Events> event) {
      */
     @Override
     public void transition(final Transition<States, Events> transition) {
-        UserConsole.getLogger().info(
+        log.info(
```
ajoymajumdar: Is it intentional to not use UserConsole.getLogger()?

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/UserMetricsTask.java
```diff
@@ -88,37 +92,77 @@ public long getFixedRate() {
...
-            if (!summaries.containsKey(user)) {
-                this.publishUserMetrics(user, 0, 0);
-            }
-        }
+        // Update number of active users
+        log.info("Number of users with active jobs: {}", summaries.size());
+        this.activeUsersCount.set(summaries.size());
+
+        // Track users who previously had jobs but no longer do
+        final Set<String> usersToReset = Sets.newHashSet(this.userResourcesRecordMap.keySet());
```
ajoymajumdar: Can use Sets.difference(this.userResourcesRecordMap.keySet(),summaries.keySet())

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/listeners/LoggingListener.java
mprimi: Yes, this was printing the agent state machine transition changes in the user terminal.
That's a bit too low level, so I'm just sending it to the log file instead.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/UserMetricsTask.java
mprimi: There are a couple of ways, such as `Sets.difference` that do look better in terms of readability. However they produce views of the sets and then throw `ConcurrentModificationException` when I'm iterating this difference and removing from the map.

PR #877, opened 2019-05-28
Re-implement ClusterCheckerTask

mprimi:
- Inspect response body in case of success and failure
 - Treat empty response as a failuregit
 - Change metrics
  * Publish counter for each indicator, unless UP or ignored, tagged with target node name
  * Publish counter for each failure to talk to a node, tagged with target node name


genie-docs/src/docs/asciidoc/_metrics.adoc
```diff
@@ -343,29 +343,29 @@ published within the local JVM and available on the Actuator `/metrics` endpoint
 |JobMonitoringCoordinator
 |-
 
-|genie.tasks.clusterChecker.errorCounts.gauge
-|Number of Genie nodes that the current leader failed is presently failing to contact
+|genie.tasks.clusterChecker.unhealtyHosts.gauge
```
tgianos: we haven't been very good about it but there is a release notes type section of the ascidocs where you might want to capture that you're changing these metrics names

mprimi: I thought about it, but we've missed quite a few so i don't know that the document is very valuable as it is incomplete.
Beside, if one wants to see what changed, this would kinda work: `git diff v3.3.x v.4.0.x -- genie-docs/src/docs/asciidoc/_metrics.adoc`

PR #878, opened 2019-05-30
Add options for job dependencies, configurations to Agent

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentConverters.java
```diff
@@ -175,4 +176,54 @@ public JsonNode convert(final String value) {
             }
         }
     }
+
+    static final class UriOrLocalPathConverter implements IStringConverter<String> {
+        static final String ATTACHMENT_HELP_MESSAGE = "ATTACHMENTS:\n"
+            + "Different kinds of job-specific files can be attached to a job.\n"
```
tgianos: This is likely a minor thing but I worry about causing confusion with users by overloading the term "attachment" in different contexts. Historically, and I'm not saying we can't improve or correct this if we think it's wrong, we've used `attachment` to mean a file not available anywhere else that is being sent along with the job request to Genie. 

The wording in this help message makes it seem like all different types of files are these kinds of historical `attachments`. We've used `dependency` to mean something available somewhere else. 

I'm not sure this needs to be solved right now I just want to avoid users getting confused when we're debugging something and they say it's an `attachment`. If that makes sense.

This all may be part of a bigger discussion about how we evolve file handling for jobs but I wanted to make sure I pointed it out so at least we think through it. This is likely ok for now as most end users won't see it since internally we wrap in that Python lib.

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentConverters.java
```diff
@@ -175,4 +176,54 @@ public JsonNode convert(final String value) {
...
+            + "execution (conditional on cleanup and archival options).\n"
+            + "\n"
+            + "Attachments can either be valid URIs or paths to local files, example:\n"
+            + "  s3://configurations/hadoop/spark/1.6.1/hive-site.xml\n"
+            + "  http://some-domain.org/some-project/my-config.properties\n"
+            + "  file:///tmp/myscript.presto\n"
+            + "  file:/tmp/query.sql\n"
+            + "  ./myscript.sql (shortcut for file:/${PWD}/myscript.sql)\n";
+
+        @Override
```
tgianos: `@inheritdoc`?

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
```diff
@@ -162,6 +162,29 @@
     )
     private boolean jobRequestedViaAPI;
 
+    @Parameter(
+        names = {"--jobConfiguration"},
+        description = "URI or path of a job-level configuration file to attach, can be repeated",
```
tgianos: all of these say URI or path. If it's a path how are we handling it? We're just going to copy it into the working directory locally or are we going to convert it and push it somewhere in S3 first?

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentConverters.java
mprimi: I agree and this is actually my attempt at setting things straight by moving our lexicon in a specific direction.
In V3, we have these 3 types of files, plus we have a fourth: "anonymous" attachments (anonymous because they are not recorded in the request or anywhere else). 

I think we should move away from anonymous attachments and instead force the users to think in terms of configs, and dependencies and setup. 
And consider the word "attachment" for a config or a dependency or a setup file that is uploaded along with the job request.

genie-agent/src/main/java/com/netflix/genie/agent/cli/ArgumentConverters.java
mprimi: 👍

genie-agent/src/main/java/com/netflix/genie/agent/cli/JobRequestArgumentsImpl.java
mprimi: You can see how it's handled in `ArgumentConverters.UriOrLocalPathConverter.class`, in short it's either a URI already, or i assume it's a local path and try to convert it to a URI (`file:///Users/foo/.../myfile.sql`).

So in the spec back from server, everything is a URI, and the resource loaded loads all of them regardless of where they come from (local disk, s3, ...)

I don't think we should be in the business of uploading to S3 to make a local file available remotely.
First of all, it's unnecessary.
Second, we are "sharing" a user file without the user realizing.

PR #879, opened 2019-06-03
Minor operational improvements

mprimi:


genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/DirectoryManifest.java
```diff
@@ -379,10 +379,10 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
         @Override
         public FileVisitResult visitFileFailed(final Path file, final IOException ioe) {
             if (ioe instanceof FileSystemLoopException) {
-                log.warn("Detected file system cycle visiting while visiting {}. Skipping.", file, ioe);
+                log.warn("Detected file system cycle visiting while visiting {}. Skipping.", file );
```
tgianos: You might want to add a note in the code why the exception isn't included in the logs since it is inconsistent with our other logging for exceptions.

mprimi: If you do a quick search for `log.error` or `log.warn` there's actually quite a few other places where we don't log the exception for the same reason.

PR #880, opened 2019-06-05
Metrics improvements and bug fixes

mprimi:


genie-web/src/main/java/com/netflix/genie/web/tasks/leader/UserMetricsTask.java
```diff
@@ -68,7 +67,9 @@ public UserMetricsTask(
         this.registry = registry;
         this.jobSearchService = jobSearchService;
         this.userMetricsProperties = userMetricsProperties;
-        this.registry.gauge(USER_ACTIVE_USERS_METRIC_NAME, this.activeUsersCount);
+        this.activeUsersCount = new AtomicLong(0);
+        this.activeUsersCountGauge = Gauge.builder(USER_ACTIVE_USERS_METRIC_NAME, activeUsersCount::get)
```
tgianos: Pretty sure you can do this all in one line:
```this.activeUsersCount = this.registry.gauge(USER_ACTIVE_USERS_METRIC_NAME, new AtomicLong(0));```
per [micrometer docs](https://micrometer.io/docs/concepts#_gauges)

tgianos: Nevermind I see you're trying to keep a hard reference so maybe the builder is the only way unless you pull it back out from the registry after this builder.

mprimi: Yes switching to hard reference is the main change in this diff.

PR #884, opened 2019-07-10
Refactor gRPC server configuration

tgianos:
Previously we've relied on a 3rd party gRPC Spring Boot starter (https://github.com/yidongnan/grpc-spring-boot-starter) to bootstrap our server side gRPC configuration and server. This enforced certain conventions that made it hard to make the system pluggable. For example it relied on annotation driven bean discovery which makes it really hard to override beans if we want.

Additionally it didn't line up with a way to apply internal gRPC conventions if desired.

This commit removes that dependency and replaces it with a more "native" level gRPC configuration that should alleviate the afformentioned concerns.

There are some other bleed through changes that may have been picked up while scrolling through files but they were minor (spelling fixes, code formatting, class misnaming)


tgianos:
@twicksell if you get a chance can you take a glance over this, particularly the auto configurations, to verify that we could plugin internal gRPC mechanisms if desired (`ManagedServer`, `MetatronChannel`) etc.

tgianos:
@elandau if you have a chance to take a look as well that'd be helpful

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServerAutoConfiguration.java
```diff
@@ -0,0 +1,115 @@
...
+    @Bean
+    @ConditionalOnMissingBean(Server.class)
+    public Server gRpcServer(
+        @Value("${grpc.server.port:0}") final int port,
+        final Set<BindableService> services,
+        final List<ServerInterceptor> serverInterceptors
+    ) {
+        final NettyServerBuilder builder = NettyServerBuilder.forPort(
+            port == RANDOM_PORT
+                ? SocketUtils.findAvailableTcpPort()
```
elandau: I'd prefer to set the port to 0 and let the socket bind to an ephemeral port.  There's no chance of collision that way.

tgianos: Good point I'll fix that

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServerAutoConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param port               The port this server should listen on
+     * @param services           The gRPC services this server should serve
+     * @param serverInterceptors The {@link ServerInterceptor} implementations that should be applied to all services
+     * @return A Netty server instance based on the provided information
+     */
+    @Bean
+    @ConditionalOnMissingBean(Server.class)
+    public Server gRpcServer(
+        @Value("${grpc.server.port:0}") final int port,
```
mprimi: We don't use `@Value` in the rest of the codebase (that i'm aware of). Any particular reason for straying from convention here?

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServicesAutoConfiguration.java
```diff
@@ -0,0 +1,183 @@
...
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+/**
+ * Configures various gRPC services and related beans if gRPC functionality is enabled.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+@Configuration
+@ConditionalOnProperty(value = GRpcServerProperties.ENABLED_PROPERTY, havingValue = "true")
```
mprimi: This tangles these services creation to wether or not the server is enabled.
While it does make sense, it seems unnecessary, they don't actually depend on it.

genie-web/build.gradle
```diff
@@ -116,10 +116,10 @@ def genieVersion = project.version.toString()
 asciidoctor {
     dependsOn integrationTest
     inputs.dir snippetsDir
-    attributes       \
-                  "snippets": snippetsDir,
-            "revnumber": genieVersion,
-            "docinfo1": ""
+    attributes        \
+                   "snippets" : snippetsDir ,
```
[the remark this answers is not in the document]
tgianos: Yeah this whole thing has driven me nuts for a while. IDE constantly complains it is wrong but it has worked and I was afraid to touch it. Let me see if I can fix it in a subsequent commit.

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServerAutoConfiguration.java
tgianos: yeah. We've tended to avoid it. I used it here (for now) because:
- We already have a `GRpcServerProperties` class but its prefix is `genie.grpc.server` and I couldn't decide if I wanted to move `port` under there or not and since I couldn't decide I didn't want to either create a new class or rename the existing one
- This is the property set from the grpc library we were using so theoretically I wouldn't have to change internal configuration to plot this in
- It might also be the property of the internal gRPC starters. I have to double check that

Effectively I'm not sure what we'll want to do here and since we weren't using the gRPC port anywhere else (yet) I was comfortable putting it in temporarily while fleshing out the rest of this

I'll add a `TODO` so it's clear I don't intend it to stick around

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServicesAutoConfiguration.java
tgianos: That is true and I'd like to remove it however it kind of messes up other things since we have certain entanglements that we should clean up. There are at least two service interfaces that are directly implemented by these service beans:

- `JobKillServiceV4` in `GRpcJobKillServiceImpl`
- `AgentFileStreamService` in `GRpcAgentFileStreamServiceImpl`

Implementations for these interfaces are required whether the gRPC server is started or not. So elsewhere in `GenieServicesAutoConfiguration` we have fallback anonymous implementations if gRPC is off.

This is kind of why I was asking if we can just leave gRPC on all the time as right now it's critical. If we disable this conditional then these beans would always be created and when those services are invoked and gRPC wasn't enabled it would behave really weird.

To me this smells like some tangling we should to fix to make the delineation between modules cleaner.  Let me know if that makes sense.

PR #886, opened 2019-07-15
Enable daemon, enable info logging, add more memory to gradle

chali:


chali:
Merged the changes from a different branch in the and see #887

PR #888, opened 2019-07-16
Changes to set groundwork for local agent launch

tgianos:
A few commits which set some of the groundwork to enable launching API jobs using the Genie agent instead of the 3.x job execution code

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServicesAutoConfiguration.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServicesAutoConfiguration.java
@@ -25,7 +25,6 @@ import com.netflix.genie.proto.HeartBeatServiceGrpc;
 import com.netflix.genie.proto.JobKillServiceGrpc;
 import com.netflix.genie.proto.JobServiceGrpc;
 import com.netflix.genie.proto.PingServiceGrpc;
-import com.netflix.genie.web.properties.GRpcServerProperties;
 import com.netflix.genie.web.rpc.grpc.services.impl.v4.GRpcAgentFileStreamServiceImpl;
 import com.netflix.genie.web.rpc.grpc.services.impl.v4.GRpcHeartBeatServiceImpl;
 import com.netflix.genie.web.rpc.grpc.services.impl.v4.GRpcJobKillServiceImpl;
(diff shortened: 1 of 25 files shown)


genie-web/src/main/java/com/netflix/genie/web/exceptions/checked/AgentLaunchException.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package com.netflix.genie.web.exceptions.checked;
+
+/**
+ * An exception for when the server can't launch an agent for whatever reason.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+public class AgentLaunchException extends Exception {
```
mprimi: Adding a mandatory constructor argument with the array of executable and arguments could be nice to debug.

genie-web/src/test/groovy/com/netflix/genie/web/exceptions/checked/AgentLaunchExceptionSpec.groovy
```diff
@@ -0,0 +1,62 @@
+/*
```
mprimi: At some point we may want to collapse all these boilerplate tests into one like I'm doing for `AgentExecutionExceptionsSpec`

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServicesAutoConfiguration.java
```diff
@@ -51,7 +49,6 @@
  * @since 4.0.0
  */
 @Configuration
-@ConditionalOnProperty(value = GRpcServerProperties.ENABLED_PROPERTY, havingValue = "true")
```
mprimi: Have you noticed any different in runtime for tests?
Any integration test that does not exclude this configuration now will bring up a gRPC server.
And i wonder if that makes any noticeable difference.

genie-web/src/main/java/com/netflix/genie/web/exceptions/checked/AgentLaunchException.java
tgianos: Seems useful. I'll add it in a subsequent commit in a different PR

genie-web/src/main/java/com/netflix/genie/web/configs/grpc/GenieGRpcServicesAutoConfiguration.java
tgianos: Haven't noticed any noticeable difference but disabling it is still possible by providing a `noop` bean of type `GRpcServerManager` in the test context. Then it just won't start the server. But really it's a netty server listening on a port. Seems almost instant at least compared to tomcat.

PR #891, opened 2019-07-22
URL-encoding in forwarding and ScriptLoadBalancer metrics

mprimi:
- Avoid URL-encoding when forwarding from one node to another.
 - Fix incorrect ScriptLoadBalancer metrics

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
@@ -177,6 +177,8 @@ public class ScriptLoadBalancer implements ClusterLoadBalancer {
             for (final Cluster cluster : clusters) {
                 if (clusterId.equals(cluster.getId())) {
                     tags.add(Tag.of(MetricsConstants.TagKeys.STATUS, STATUS_TAG_FOUND));
+                    tags.add(Tag.of(MetricsConstants.TagKeys.CLUSTER_NAME, cluster.getMetadata().getName()));
+                    tags.add(Tag.of(MetricsConstants.TagKeys.CLUSTER_ID, cluster.getId()));
                     return cluster;
                 }
             }
--- a/genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
@@ -17,6 +17,7 @@
  */
 package com.netflix.genie.web.services.loadbalancers.script;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Sets;
 import com.netflix.genie.common.dto.JobRequest;
@@ -37,6 +38,7 @@ import org.springframework.core.task.AsyncTaskExecutor;
 import org.springframework.scheduling.TaskScheduler;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.script.Bindings;
 import javax.script.Compilable;
 import javax.script.CompiledScript;
@@ -55,9 +57,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


mprimi:
@tgianos one more commit to refactor and test `ScriptLoadBalancer`

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -155,12 +157,72 @@ public Cluster selectCluster(
...
+
+            if (clusterId == null) {
+                tags.add(Tag.of(MetricsConstants.TagKeys.STATUS, STATUS_TAG_NO_PREFERENCE));
+                log.debug("Script returned null, no preference");
+                return null;
+            }
+
+            for (final Cluster cluster : clusters) {
+                if (clusterId.equals(cluster.getId())) {
+                    tags.add(Tag.of(MetricsConstants.TagKeys.STATUS, STATUS_TAG_FOUND));
```
tgianos: Maybe we should tag with the cluster id selected as well?

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -155,12 +157,72 @@ public Cluster selectCluster(
...
+    static class Evaluator {
+
+        private final ObjectMapper mapper;
+        private final AsyncTaskExecutor asyncTaskExecutor;
+        private final long timeoutLength;
+
+        Evaluator(
+            final ObjectMapper mapper,
+            final AsyncTaskExecutor asyncTaskExecutor,
+            final long timeoutLength) {
```
tgianos: usually this `) {` is on the next line

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -155,12 +157,72 @@ public Cluster selectCluster(
...
+            return null; // Should throw
+
+        } finally {
+            this.registry
+                .timer(SELECT_TIMER_NAME, tags)
+                .record(System.nanoTime() - selectStart, TimeUnit.NANOSECONDS);
+        }
+    }
+
+    static class Evaluator {
```
tgianos: This makes it package private which would force anyone trying to extend it to put that code in this package. Might want to make it explicitly `protected`?

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -174,139 +236,139 @@ public Cluster selectCluster(
...
-                )
+        private final Environment environment;
+        private final MeterRegistry registry;
+        private final GenieFileTransferService fileTransferService;
+
+
+        Loader(
+            final TaskScheduler taskScheduler,
+            final Environment environment,
+            final GenieFileTransferService fileTransferService,
```
tgianos: Not to increase the scope creep here but we'll probably want to get rid of this service soon so if you were so inclined to flip this to use the `ResourceLoader` I wouldn't object

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
```diff
@@ -174,139 +236,139 @@ public Cluster selectCluster(
...
-            );
-            this.isConfigured.set(false);
-        } finally {
-            this.isUpdating.set(false);
-            this.registry
-                .timer(UPDATE_TIMER_NAME, tags)
-                .record(System.nanoTime() - updateStart, TimeUnit.NANOSECONDS);
-            log.debug("Refresh completed");
+        @Nullable
+        CompiledScript get() {
```
tgianos: Is this one of those `@VisibleForTesting` things? Seems unnecessary

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
mprimi: Would be nice to have.
Tho then we should also do the same in the class that uses this component (`JobSpecService`), since more than one `LoadBalancer` may be used.

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
mprimi: It's part of the script `Loader` interface so it can be mocked easily.
But yeah, package private is probably not the best choice, i'll fix it.

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
mprimi: Added a commit that adds metrics tags

genie-web/src/main/java/com/netflix/genie/web/services/loadbalancers/script/ScriptLoadBalancer.java
mprimi: Fixed, thank you.

PR #893, opened 2019-08-01
Bug - File streaming failing for job whose ID contains certain characters

mprimi:


genie-web/src/main/java/com/netflix/genie/web/agent/resources/AgentFileProtocolResolver.java
```diff
@@ -55,34 +55,57 @@ public AgentFileProtocolResolver(
...
      */
     @Override
     public Resource resolve(final String location, final ResourceLoader resourceLoader) {
         log.debug("Attempting to resolve if {} is an Agent file resource or not", location);
         final URI uri;
+        final String jobId;
+        final Path relativePath;
+
         try {
             uri = URI.create(location);
```
tgianos: Why not use `new URI` here so you get the checked exception and catch that instead of `IllegalArgumentException` that `create` wraps the syntax exception in

genie-web/src/test/groovy/com/netflix/genie/web/agent/resources/AgentFileProtocolResolverSpec.groovy
```diff
@@ -46,7 +47,7 @@ class AgentFileProtocolResolverSpec extends Specification {
         setup:
         String jobId = UUID.randomUUID().toString()
         Path relativePath = Paths.get("foo/bar.txt")
-        URI uri = new URI(AgentFileProtocolResolver.URI_SCHEME, jobId, "/" + relativePath, null)
+        URI uri = AgentFileProtocolResolver.createUri(jobId, "/" + relativePath)
```
tgianos: Are you always expecting the `path` to have the `/` as a start? I didn't see in the code if you're checking for it. What happens if it's not there?

genie-web/src/main/java/com/netflix/genie/web/agent/resources/AgentFileProtocolResolver.java
```diff
@@ -64,14 +66,19 @@ public AgentFileProtocolResolver(
      * @throws URISyntaxException if constructing the URI fails
      */
     public static URI createUri(final String jobId, final String path) throws URISyntaxException {
-        return new URI(AgentFileProtocolResolver.URI_SCHEME, jobId, path, null);
+        final String encodedJobId = Base64.encodeBase64URLSafeString(jobId.getBytes());
```
tgianos: Now that you're encoding the ID are you missing decoding it somewhere?

I don't see it anywhere

genie-web/src/test/groovy/com/netflix/genie/web/agent/resources/AgentFileProtocolResolverSpec.groovy
mprimi: No, works in both cases. I'll add a test to document this behavior.

genie-web/src/main/java/com/netflix/genie/web/agent/resources/AgentFileProtocolResolver.java
mprimi: We're implementing an interface here.
So throwing a checked exception would require catching and rewrapping and throwing. i.e. what `create` is already doing.

Beside, this is out of the scope of this PR and i'd rather not touch it.

genie-web/src/test/groovy/com/netflix/genie/web/agent/resources/AgentFileProtocolResolverSpec.groovy
mprimi: Added

PR #894, opened 2019-08-02
Completed Job summary DTO and Service

mprimi:


mprimi:
@tgianos since the rest of this PR is "trivial" but a lot of busy work, do you mind taking a look at this approach and letting me know if there's any red flags?

PR #895, opened 2019-08-06
More work towards local job runs via agent

tgianos:
Much of this PR is work towards being able to launch a job locally on Genie hardware using the agent instead of the old V3 logic. It's still groundwork and isn't hooked up to the actual API yet.

The primary change here, and why this is a PR now, is changing how arguments are parsed on a V3 job submission. The agent changed, for the better, how we handled job arguments. This change didn't work well with how V3 was treating the command args as one unified string. Effectively without escaping single quotes properly after submission the agent would treat the string as one big argument instead of N arguments as was the intent by the user. This PR contains a change, as the final commit, which splits the V3 command arg string on whitespace and `'` and from there on treats all the arguments as independent. In serializing to the V3 run script, and as responses to `GET` requests, it will re-wrap all arguments in `'`. This means that the command arg string won't look exactly like what the user originally submitted however it will behave more correctly across V3 and V4 going forward.

The hope here is to merge this, test it with a `RC` to make sure it doesn't really break anything and then move forward with the continued port of job launch to agent on top of this change.

### DIFF
--- a/genie-common-internal/src/main/java/com/netflix/genie/common/internal/services/impl/FileSystemJobArchiverImpl.java
+++ b/genie-common-internal/src/main/java/com/netflix/genie/common/internal/services/impl/FileSystemJobArchiverImpl.java
@@ -83,7 +83,20 @@ public class FileSystemJobArchiverImpl implements JobArchiver {
         private final Path source;
         private final Path target;
 
-        JobDirectoryCopier(final Path source, final Path target) {
+        JobDirectoryCopier(final Path source, final Path target) throws IOException {
+            // If the source doesn't exist or isn't a directory throw an exception
+            if (!Files.exists(source) || !Files.isDirectory(source)) {
+                throw new IOException(source + " doesn't exist or isn't a directory. Unable to copy");
+            }
+            // If the target exists but isn't a directory throw an error
+            if (Files.exists(target) && !Files.isDirectory(target)) {
+                throw new IOException(target + " exists but isn't a directory");
+            }
+            // If the target doesn't exist try to create it
+            if (!Files.exists(target)) {
+                Files.createDirectories(target);
+            }
+
             this.source = source;
             this.target = target;
         }
--- a/genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
@@ -18,7 +18,9 @@
 package com.netflix.genie.web.agent.launchers.impl;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.netflix.genie.common.internal.util.GenieHostInfo;
 import com.netflix.genie.web.agent.launchers.AgentLauncher;
+import com.netflix.genie.web.data.services.JobSearchService;
 import com.netflix.genie.web.dtos.ResolvedJob;
 import com.netflix.genie.web.exceptions.checked.AgentLaunchException;
 import com.netflix.genie.web.properties.LocalAgentLauncherProperties;
@@ -51,6 +53,11 @@ public class LocalAgentLauncherImpl implements AgentLauncher {
     private static final String JOB_ID_OPTION = "--jobId";
     private static final String FULL_CLEANUP_OPTION = "--full-cleanup";
 
+    private static final Object MEMORY_CHECK_LOCK = new Object();
+
+    private final JobSearchService jobSearchService;
+    private final LocalAgentLauncherProperties launcherProperties;
+    private final String hostname;
     private final ExecutorFactory executorFactory;
     private final MeterRegistry registry;
     private final CommandLine commandTemplate;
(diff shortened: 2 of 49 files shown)


genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
```diff
@@ -51,29 +53,42 @@
     private static final String JOB_ID_OPTION = "--jobId";
     private static final String FULL_CLEANUP_OPTION = "--full-cleanup";
 
+    private static final Object MEMORY_CHECK_LOCK = new Object();
```
mprimi: Why static?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
```diff
@@ -102,10 +117,43 @@ public LocalAgentLauncherImpl(
...
+                "Unable to launch job as the requested job memory ("
+                    + jobMemory
+                    + "MB) exceeds the maximum allowed by the configuration of the system ("
+                    + this.launcherProperties.getMaxJobMemory()
+                    + "MB)"
+            );
+        }
+
+        // One at a time to ensure we don't overflow configured max
+        synchronized (MEMORY_CHECK_LOCK) {
```
mprimi: Maybe leave a TODO or note here as a reminder that this does not guarantee enforcement of the limit.

Also after reading this, i think `getUsedMemoryOnHost` should include more than `RUNNING` as per comment on previous commit.

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
```diff
@@ -173,6 +189,39 @@ public void launchAgent(@Valid final ResolvedJob resolvedJob) throws AgentLaunch
         }
     }
 
+    /**
```
mprimi: Since it's all data methods, why merging the two in here, as opposed to having a standalone indicator?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
```diff
@@ -45,18 +43,7 @@
...
-    static final String NUMBER_RUNNING_JOBS_KEY = "numRunningJobs";
-    @VisibleForTesting
-    static final String ALLOCATED_MEMORY_KEY = "allocatedMemory";
-    @VisibleForTesting
-    static final String USED_MEMORY_KEY = "usedMemory";
-    @VisibleForTesting
-    static final String AVAILABLE_MEMORY = "availableMemory";
-    @VisibleForTesting
-    static final String AVAILABLE_MAX_JOB_CAPACITY = "availableMaxJobCapacity";
+public class LocalAgentLauncherImpl implements AgentLauncher {
```
mprimi: Ah, this answers my previous comment about splitting them

genie-common-internal/src/main/java/com/netflix/genie/common/internal/services/impl/FileSystemJobArchiverImpl.java
```diff
@@ -83,7 +83,20 @@ public boolean archiveDirectory(final Path directory, final URI target) throws J
         private final Path source;
         private final Path target;
 
-        JobDirectoryCopier(final Path source, final Path target) {
+        JobDirectoryCopier(final Path source, final Path target) throws IOException {
+            // If the source doesn't exist or isn't a directory throw an exception
```
mprimi: I don't know that i agree with checking these here, since by the time you get to the actual copy they may no longer be true, so why bother.

genie-web/src/main/java/com/netflix/genie/web/data/services/jpa/JpaJobSearchServiceImpl.java
```diff
@@ -436,6 +439,30 @@ public JobMetadata getJobMetadata(@NotBlank final String id) throws GenieExcepti
...
+    public long getAllocatedMemoryOnHost(final String hostname) {
+        return this.jobRepository.getTotalMemoryUsedOnHost(hostname, JobStatus.getActiveStatuses());
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public long getUsedMemoryOnHost(final String hostname) {
+        return this.jobRepository.getTotalMemoryUsedOnHost(hostname, RUNNING_JOB_SET);
```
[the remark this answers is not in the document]
tgianos: Are you saying it's a bad variable name or that it's missing a state? I didn't include `INIT` cause that means the job isn't actually using any memory on the box. I thought we'd discussed this in person

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
tgianos: Because though in practice this will likely be a singleton instance in our app I wanted to ensure that even if it wasn't the memory check would be atomic across all instance and that wouldn't happen if the lock was an instance variable

genie-common-internal/src/main/java/com/netflix/genie/common/internal/services/impl/FileSystemJobArchiverImpl.java
tgianos: I actually added these checks after some integration tests failed due to some of these conditions

genie-web/src/main/java/com/netflix/genie/web/data/services/jpa/JpaJobSearchServiceImpl.java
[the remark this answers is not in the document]
tgianos: Per conversation I'll change the set of states to be `CLAIMED`, `INIT` and `RUNNING`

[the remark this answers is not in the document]
tgianos: Done

PR #899, opened 2019-08-12
Job completion summary via SNS

mprimi:


genie-web/src/main/java/com/netflix/genie/web/events/SNSNotificationsPublisher.java
```diff
@@ -89,6 +89,11 @@ public SNSNotificationsPublisher(
      */
     @Override
     public void onApplicationEvent(final JobStateChangeEvent event) {
+        if (!this.properties.isEnabled()) {
```
tgianos: you've got this `!fixup` commit still in here

genie-web/src/main/java/com/netflix/genie/web/events/AbstractSNSPublisher.java
```diff
@@ -37,81 +35,46 @@
...
- * Currently, only job status changes trigger a notification.
+ * Abstract SNS publisher that adds fields common to all payloads, records metrics, etc.
  *
  * @author mprimi
  * @since 4.0.0
  */
 @Slf4j
-public class SNSNotificationsPublisher implements ApplicationListener<JobStateChangeEvent> {
-
+public class AbstractSNSPublisher {
```
tgianos: should you make this abstract and implement `ApplicationListener<JobStateChangedEvent>` or is there a thought that maybe others could publish off different internal events?

genie-web/src/main/java/com/netflix/genie/web/events/JobFinishedSNSPublisher.java
```diff
@@ -0,0 +1,211 @@
...
+    ) {
+        super(properties, registry, snsClient, mapper);
+        this.jobPersistenceService = jobPersistenceService;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onApplicationEvent(final JobStateChangeEvent event) {
```
tgianos: Would it be more valuable for the system to send an internal `JobFinishedEvent` (I believe we have something like this in v3 already) so that you don't have to do additional filtering in here? Plus in that event you could include information already and maybe save db calls?

genie-web/src/main/java/com/netflix/genie/web/events/AbstractSNSPublisher.java
mprimi: I should make this class abstract. But I'd rather not make it specific to `JobStateChangedEvent`

genie-web/src/main/java/com/netflix/genie/web/events/JobFinishedSNSPublisher.java
mprimi: The DB call is unavoidable. 
Publishing that different event on the bus does indeed simplify this publisher but puts the same complexity upstream so overall i kinda like the current approach of listening to job status changes and doing whatever necessary

genie-web/src/main/java/com/netflix/genie/web/events/AbstractSNSPublisher.java
tgianos: Ok then you could make it `implements ApplicationListener<? extends ApplicationEvent>`

PR #903, opened 2019-08-29
URI property parsing

tgianos:


genie-test-web/src/main/resources/application.yml
```diff
@@ -29,7 +29,7 @@ cloud:
 genie:
   file:
     cache:
-      location: file:///tmp/genie/cache
+      location: file://${java.io.tmpdir}genie/cache
```
ajoymajumdar: Are you assuming that the java.io.tmpdir will not contain the scheme?

tgianos: In every example and test i've run it doesn't but yeah it's an assumption.

PR #905, opened 2019-09-05
Add job timeout to agent protocol. Add isAPIJob flag to database

tgianos:
Other changes related to these two core changes.

Reverts changing working directory of agent as well


genie-web/src/main/java/com/netflix/genie/web/services/JobResolverService.java
```diff
@@ -54,8 +54,9 @@
      *
      * @param id         The id of the job
      * @param jobRequest The job request containing all details a user wants to have for their job
+     * @param apiJob     {@literal true} if this job was submitted via the REST API. {@literal false} otherwise.
      * @return The completely resolved job information within a {@link ResolvedJob} instance
      */
     @Nonnull
-    ResolvedJob resolveJob(String id, @Valid JobRequest jobRequest);
+    ResolvedJob resolveJob(String id, @Valid JobRequest jobRequest, boolean apiJob);
```
mprimi: Seems strange we have to pass this boolean in order to resolve... It doesn't actually influence the resolution so seems off to have it here

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/MonitorJobAction.java
```diff
@@ -72,12 +73,30 @@ protected Events executeStateAction(final ExecutionContext executionContext) {
 
         log.info("Job process completed with final status {}", finalJobStatus);
 
+        // TODO: Likely want to clean this up as it only contains a few cases.
```
mprimi: Alternative, have constant format strings in addition to constant strings.
i.e. this one could be `"Job process completed with final status: %s"`

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
```diff
@@ -185,6 +176,16 @@ protected void executeStateActionCleanup(final ExecutionContext executionContext
         fileManifestService.stop();
     }
 
+    private void relocateAgentLogFile(final File jobDirectory) {
```
mprimi: Moving pieces of code around really makes it hard to review because I need to check line by line to see if something actually changed or the block was moved down as-is. 
This PR is full of such changes.
Could you please submit a PR that re-formats the entire codebase so we can stop having these changes mixed in?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/converters/JobServiceProtoConverter.java
```diff
@@ -21,6 +21,7 @@
 import com.google.common.collect.ImmutableSet;
```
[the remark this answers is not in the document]
tgianos: Split into 3 commits

genie-proto/src/main/proto/genie.proto
```diff
@@ -85,6 +86,7 @@ message ExecutionResourceCriteria {
 message AgentConfig {
     bool is_interactive = 1;
     string job_directory_location = 2;
+    google.protobuf.Int32Value timeout = 3; // Requested timeout in seconds
```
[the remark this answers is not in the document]
tgianos: This message is a sub-message of the `ReserverJobIdRequest` and the `DryRunJobSpecificationRequest` so it is used as input to the job request. It's DTO representation is a member of the V4 `JobRequest` DTO as well. So this is the requested timeout and I just left it named `timeout` to be consistent with the other members of this proto which don't have `requested` in their names.

genie-web/src/main/java/com/netflix/genie/web/data/entities/JobEntity.java
```diff
@@ -121,7 +121,9 @@
         "resolved",
         "claimed",
         "v4",
-        "requestedArchiveLocationPrefix"
+        "requestedArchiveLocationPrefix",
+        "timeoutUsed",
```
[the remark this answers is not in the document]
tgianos: I agree with your sentiment here. As of right now the convention in the table is `requested_X` (e.g. `requested_memory`, `requested_timeout`) and then the final fields have been `X_used` (e.g. `memory_used`). It is likely better to change all of the `used` to something like `resolved_x` etc.

Renaming columns is a metadata only operation so it is easy to do (other than hard to go backwards between versions) but if you want we can change or do it in a subsequent commit

genie-common-internal/src/main/java/com/netflix/genie/common/internal/dto/v4/JobRequestMetadata.java
```diff
@@ -45,19 +45,32 @@
...
         final int numAttachments,
         final long totalSizeOfAttachments
-    ) {
+    ) throws IllegalArgumentException {
+        if (apiClientMetadata == null && agentClientMetadata == null) {
+            throw new IllegalArgumentException(
+                "Both apiClientMetadata and agentClientMetadata are missing. One is required."
+            );
+        }
+        if (apiClientMetadata != null && agentClientMetadata != null) {
```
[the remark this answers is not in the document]
tgianos: I'm pretty sure.

This is within `JobRequestMetadata` and is only used to capture/return data about the initial request not subsequent execution metadata. There's no way (that I can think of right now) that a Job Request could come from both the API and some other mechanism.

Doing this allows us to not have to pass in an `isApi` `boolean` to this object as well as we can derive it based on which one of the metadata objects is passed in. I've considered other options like Template class on one of these metadata types but it would be a pain.

If you can think of a good reason we'd need both at initial job submission time we can remove this commit.

I looked at the job service and perhaps you're mixing this with `claimJob` which sends the `AgentMetadata` again and we persist it within `JobPersistenceService.claimJob`

genie-web/src/main/java/com/netflix/genie/web/data/entities/projections/JobApiProjection.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.netflix.genie.web.data.entities.projections;
+
+/**
+ * A projection which allows the system to pull back whether the job was submitted via the REST API or other mechanism.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+public interface JobApiProjection {
```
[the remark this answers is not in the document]
tgianos: It's added to this so that there can be an API that just determines if a given job was submitted via API or not. The current implementation of the `JobResolverService` already had a whole bunch of queries run (instead of some huge payload query as it may not even be possible) so there was no real clean place to put it other than making a single call which is captured in the service api `isApiJob` in this commit.

We can combine this projection with other projections via interface extension (we already do this in many places) however that doesn't really solve `search for just API jobs` as these projections are only for what data is returned from a given query not the criteria of the query. That would be either a new repository api or `criteria` dynamic query. Unless you want to add this to other results and then filter the results within the java code which I doubt.

genie-web/src/main/java/com/netflix/genie/web/services/JobResolverService.java
tgianos: This `resolveJob` signature is the three argument one. It is used in three places (currently). The dry run job specification, the resolve job specification call by the agent (will no longer be the case once my other work for agent execution is incorporated) and within the `JobCoordinatorServiceImpl` which is the V3 code path. 2 of these 3 are non-api but the other is only API. There was no field in the `JobRequest` which captured this data as it is in `JobRequestMetadata` as it's more data about the request then something actually provided by the user (which is what's in the job request DTO)

It is required/influences the resolution because of the cascading rules we discussed which is implemented below:
```
        final Integer timeout;
        if (jobRequest.getRequestedAgentConfig().getTimeoutRequested().isPresent()) {
            timeout = jobRequest.getRequestedAgentConfig().getTimeoutRequested().get();
        } else if (apiJob) {
            // For backwards V3 compatibility
            timeout = com.netflix.genie.common.dto.JobRequest.DEFAULT_TIMEOUT_DURATION;
        } else {
            timeout = null;
        }
```

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/MonitorJobAction.java
tgianos: It works but won't match what is in the current V3 responses (if we want to maintain backwards capability) also that string doesn't capture things like "Killed due to timeout" or "Killed due to std out too large"

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/SetUpJobAction.java
tgianos: sure. commit added to front of commit.
Could you please ensure you run the formatter before you commit from now on so that it doesn't drift?

mprimi: 👍

PR #907, opened 2019-09-10
Job process manager changes

tgianos:
This PR contains changes to improve the `JobProcessManager` class primarily through providing better return types as well as extending it to support killing a job via timeout.


genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/MonitorJobAction.java
```diff
@@ -139,6 +139,11 @@ protected Events executeStateAction(final ExecutionContext executionContext) {
         return Events.MONITOR_JOB_COMPLETE;
     }
 
+    @Override
```
mprimi: There must be a setting in your IDE that re-sorts methods and i don't have. I'm interested in finding out what it is so we can avoid this kind of changes in unrelated commits.

genie-agent/src/main/resources/META-INF/spring.factories
```diff
@@ -1,13 +1,13 @@
...
   com.netflix.genie.agent.cli.JCommanderAutoConfiguration,\
-  com.netflix.genie.agent.configs.AgentAutoConfiguration,\
-  com.netflix.genie.agent.configs.ProcessAutoConfiguration,\
   com.netflix.genie.agent.execution.ExecutionAutoConfiguration,\
   com.netflix.genie.agent.execution.services.impl.grpc.GRpcServicesAutoConfiguration,\
   com.netflix.genie.agent.execution.services.impl.ServicesAutoConfiguration,\
   com.netflix.genie.agent.execution.statemachine.StateMachineAutoConfiguration,\
   com.netflix.genie.agent.execution.statemachine.actions.StateMachineActionsAutoConfiguration,\
-  com.netflix.genie.agent.rpc.GRpcAutoConfiguration
+  com.netflix.genie.agent.rpc.GRpcAutoConfiguration,\
```
mprimi: All the classes mentioned in this file are Spring autoconfigurations. Why move just some of them?

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -195,11 +198,14 @@ public JobStatus waitFor() throws InterruptedException {
...
-            return JobStatus.SUCCEEDED;
+            return new JobProcessResult.Builder(JobStatus.KILLED, JobStatusMessages.JOB_KILLED_BY_USER).build();
+        } else if (exitCode == SUCCESS_EXIT_CODE) {
+            return new JobProcessResult.Builder(
+                JobStatus.SUCCEEDED,
+                JobStatusMessages.JOB_FINISHED_SUCCESSFULLY
+            ).build();
         } else {
-            return JobStatus.FAILED;
+            return new JobProcessResult.Builder(JobStatus.FAILED, JobStatusMessages.JOB_FAILED).build();
```
mprimi: Seems like this loses the exit code. I'm guessing you probably want to save it.
(I would also put it in the final status message, users do look at it and special codes have meaning that people need to know).

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/MonitorJobAction.java
tgianos: Perhaps it's having the checkstyle plugin tied to our checkstyle file? Not sure if that impacts it but this looks like some rule that moves based on the scope of the methods and ordering. It didn't change when I did it via the default IDE reformat from the previous PR which I had included code cleanup checked. 🤷‍♂

genie-agent/src/main/resources/META-INF/spring.factories
tgianos: I moved the two that were in the original `spring` package. The rest are sprinkled throughout the codebase due the classes they're creating all being package protected. Moving all of them would be a big change. I'd like to do it so that it's consistent but didn't feel like taking it on right now.

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
tgianos: Yeah we probably want exit code in the return that's true. I'll add it in a subsequent commit

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -197,8 +227,32 @@ public JobProcessResult waitFor() throws InterruptedException {
...
-            return new JobProcessResult.Builder(JobStatus.KILLED, JobStatusMessages.JOB_KILLED_BY_USER).build();
+        // If for whatever reason the timeout thread is currently running or if it is scheduled to be run, cancel it
+        final ScheduledFuture timeoutThreadFuture = this.timeoutKillThread.get();
+        if (timeoutThreadFuture != null) {
+            timeoutThreadFuture.cancel(true);
+        }
+
+        // TODO: This doesn't seem quite right to me and is confirmed by an existing test. If the job completes
+        //       successfully but then the system calls kill before this method then the final job status will be
+        //       killed instead of successful. We should discuss this - TJG 9/9/2019
```
tgianos: @mprimi I just wanted to call out this `TODO` to get your thoughts. We can discuss in person if easier

mprimi: I don't think it's a super big deal. The user called for a kill (or the timeout kicked in) just as the job was completing, then the final status is `killed`.
It's very unlikely to ever happen, but even if it does, it's not too incorrect, no? The job was killed before it had time to finish and finalize cleanly.

PR #910, opened 2019-09-13
Enable ability to disable jobs submitted from agent

tgianos:


genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobServiceImpl.java
```diff
@@ -113,6 +120,21 @@ public void reserveJobId(
         final StreamObserver<ReserveJobIdResponse> responseObserver
     ) {
         try {
+            // TODO: This is quick and dirty and we may want to handle it better overall for the system going forward
+            //       e.g. should it be in an filter that we can do for more APIs?
+            //            should value be cached rather than checking every time?
+            // TODO: While this currently shares code with JobRestController I couldn't find a good place to put it
+            //       that was shared and I suspect they'll diverge or change so I'm going to leave it for now
+            //       - TJG 9/13/2019
+            if (!this.environment.getProperty(JobConstants.JOB_SUBMISSION_ENABLED_PROPERTY_KEY, Boolean.class, true)) {
```
mprimi: I would move this logic to the `HandshakeService`, which can be the one place where we reject new clients

tgianos: My thought here was to limit the blast radius as much as possible. I thought if we move it to handshake it would block more actions than maybe we intend. (e.g. dry run job specification resolution). Putting it here prevents new jobs from being run which was the original intent and matches what the API is doing. We can change it though if you really think it makes more sense there given that caveat.

mprimi: Handshake is intended to be the choke point where an agent says "Hi I'm here to run a job" and the server can allow/reject.
Putting it there does not prevent anything but agent running a job.

mprimi: (also, the protocol has room for a custom message, which in this case would be: server is currently rejecting all new jobs" or something)

mprimi:
Closing, re-implemented within the existing agent filtering framework in: https://github.com/Netflix/genie/pull/911

PR #911, opened 2019-09-13
Reject agent via fast property

mprimi:


genie-web/src/main/java/com/netflix/genie/web/agent/inspectors/impl/RejectAllJobsAgentMetadataInspector.java
```diff
@@ -0,0 +1,75 @@
...
+        final boolean jobSubmissionEnabled = this.environment.getProperty(
+            JobConstants.JOB_SUBMISSION_ENABLED_PROPERTY_KEY,
+            Boolean.class,
+            true
+        );
+
+        if (jobSubmissionEnabled) {
+            return new InspectionReport(
+                InspectionReport.Decision.ACCEPT,
+                "Job submission is enabled"
```
tgianos: This probably should be a constant

genie-web/src/test/java/com/netflix/genie/web/spring/autoconfigure/agent/inspectors/AgentInspectorsAutoConfigurationTest.java
```diff
@@ -55,10 +56,12 @@ public void testContextIfEnabled() {
             .run(
                 (context) -> {
                     Assertions.assertThat(context).hasSingleBean(AgentFilterProperties.class);
-                    Assertions.assertThat(context).getBeans(AgentMetadataInspector.class).hasSize(3);
+                    Assertions.assertThat(context).getBeans(AgentMetadataInspector.class).hasSize(4);
                     Assertions.assertThat(context).hasSingleBean(WhitelistedVersionAgentMetadataInspector.class);
                     Assertions.assertThat(context).hasSingleBean(BlacklistedVersionAgentMetadataInspector.class);
                     Assertions.assertThat(context).hasSingleBean(MinimumVersionAgentMetadataInspector.class);
+                    Assertions.assertThat(context).hasSingleBean(RejectAllJobsAgentMetadataInspector.class);
+
```
tgianos: Extra line here. NBD just pointing it out

genie-agent/src/main/resources/logback-spring.xml
```diff
@@ -11,7 +11,7 @@
 
     <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
 
-    <property name="LOG_FILE" value="/tmp/genie-agent-${PID}.log"/>
+    <property name="LOG_FILE" value="${GENIE_AGENT_TEMPORARY_LOG_FILE:-/tmp/genie-agent-${PID}.log}"/>
```
tgianos: Not sure the convention here but there is a `:-`. Not sure if it should just be `:`

mprimi: This is `default value` notation like bash or other: https://www.tldp.org/LDP/abs/html/parameter-substitution.html
i.e. use `GENIE_AGENT_TEMPORARY_LOG_FILE` value if set, if not default to `/tmp/genie-agent-${PID}.log`

tgianos: Sounds good. Got it mixed up with Spring parameter substitution

PR #912, opened 2019-09-18
Accept Command 'executable' field as tokenized array

mprimi:
- Update V3 API so that 'executable' can be sent as array of strings (while retaining backward compatibility with existing clients and agent)
 - Update gRPC protocol between server and agent protocol to send executable and its arguments separately from user-provided, job-specific arguments

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
@@ -136,22 +136,28 @@ public class JobProcessManagerImpl implements JobProcessManager {
         final List<String> commandLine = Lists.newArrayList(commandArguments);
         commandLine.addAll(jobArguments);
 
-        // Configure arguments
-        log.info("Job command-line: {}", Arrays.toString(commandLine.toArray()));
+        log.info(
+            "Job command-line: {} arguments: {}",
+            Arrays.toString(commandArguments.toArray()),
+            Arrays.toString(jobArguments.toArray())
+        );
 
-        final List<String> expandedCommandLine;
+        // Configure arguments
+        final List<String> expandedCommandArguments;
         try {
-            expandedCommandLine = expandCommandLineVariables(
-                commandLine,
+            expandedCommandArguments = expandCommandLineVariables(
+                commandArguments,
                 Collections.unmodifiableMap(currentEnvironmentVariables)
             );
         } catch (final EnvUtils.VariableSubstitutionException e) {
-            throw new JobLaunchException("Job command-line arguments variables could not be expanded");
+            throw new JobLaunchException("Command executable and arguments variables could not be expanded");
         }
 
-        if (!commandLine.equals(expandedCommandLine)) {
-            log.info("Job command-line with variables expanded: {}", Arrays.toString(expandedCommandLine.toArray()));
-        }
+        final List<String> expandedCommandLine = Lists.newArrayList();
+        expandedCommandLine.addAll(expandedCommandArguments);
+        expandedCommandLine.addAll(jobArguments);
+
+        log.info("Command-line after expansion: {}", expandedCommandLine);
 
         processBuilder.command(expandedCommandLine);
 
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
@@ -17,6 +17,7 @@
  */
 package com.netflix.genie.agent.execution.process.impl;
 
+import com.google.common.collect.Lists;
 import com.netflix.genie.agent.cli.UserConsole;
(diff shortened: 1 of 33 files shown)


genie-common/src/main/java/com/netflix/genie/common/dto/Command.java
```diff
@@ -93,9 +106,10 @@ protected Command(@Valid final Builder builder) {
     public static class Builder extends ExecutionEnvironmentDTO.Builder<Builder> {
 
         private final CommandStatus bStatus;
-        private final String bExecutable;
         private final long bCheckDelay;
+        private String bExecutable;
         private Integer bMemory;
+        private ImmutableList<String> bExecutableAndArguments;
```
tgianos: Not much point in making this an immutable list if you're allowing it to be overwritten. Patterns in other places in codebase is to make this `final List<String> bExecutableAndArguments = Lists.newArrayList()` and then `clear` and copy elements into it as needed then in the `build` copy into the concrete DTO `ImmutableList` instance.

genie-common/src/main/java/com/netflix/genie/common/dto/Command.java
```diff
@@ -107,18 +121,61 @@ protected Command(@Valid final Builder builder) {
...
+        ) {
+            super(name, user, version);
+            this.bStatus = status;
+            this.bExecutableAndArguments = ImmutableList.copyOf(executableAndArguments);
+            this.bCheckDelay = checkDelay;
+        }
+
+        /*
+         * This constructor is provided just for JSON deserialization and considers both the old 'executable' and the
+         * new 'executableAndArguments' fields optional for API backward compatibility.
```
tgianos: I don't see any of these constructors marked with `@JsonCreator` so which one is used by Jackson by default?

It seems like the second public on you added really isn't necessary and you can just have this one marked `@JsonCreator` and it will use the `with` methods and then get exceptions on check in the `build`

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -125,12 +127,15 @@ public void launchProcess(
         });
 
         // Validate arguments
-        if (commandLine == null) {
+        if (commandArguments == null) {
             throw new JobLaunchException("Job command-line arguments is null");
-        } else if (commandLine.isEmpty()) {
+        } else if (commandArguments.isEmpty()) {
```
tgianos: Can't you just combine these two into an `||` and say "Job command executable is missing".

I mean technically all these checks are redundant with our validation of the DTO's and whatnot upstream but I can understand being safe and making this unit self contained.

With `@ParametersAreNonNullByDefault` all these fields should be non-null warnings in the compiler but obviously that doesn't enforce runtime so if you want to be really secure in it I guess this works.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/LaunchJobAction.java
```diff
@@ -74,17 +73,16 @@ protected Events executeStateAction(final ExecutionContext executionContext) {
...
             this.jobProcessManager.launchProcess(
                 jobDirectory,
                 jobEnvironment,
-                jobCommandLine,
+                jobSpec.getExecutableArgs(),
+                jobSpec.getJobArgs(),
                 interactive,
-                jobSpec.getTimeout().orElse(null)
-            );
+                jobSpec.getTimeout().orElse(null));
```
tgianos: Nitpick: Somehow the closing `);` got moved on to this line.

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
mprimi: I agree with this but i was trying not to change the logic beyond the strictly necessary.

genie-common/src/main/java/com/netflix/genie/common/dto/Command.java
mprimi: It wasn't marked before either because Jackson is smart enough to figure it out

mprimi: Also, the second constructor i added is the new "proper" constructor with all required arguments.
Once we phase out the old one, we can remove the `withExecutableAndArgument(List)` maybe i should mark it already deprecated

genie-common/src/main/java/com/netflix/genie/common/dto/Command.java
```diff
@@ -51,6 +54,8 @@
     @NotEmpty(message = "An executable is required")
     @Size(max = 1024, message = "Executable path can't be longer than 1024 characters")
     private final String executable;
+    @NotEmpty(message = "An executable is required")
+    private final List<@NotEmpty String> executableAndArguments;
```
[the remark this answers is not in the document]
mprimi: Making the member Immutable is a convention of the new/V4 DTOs only.

PR #913, opened 2019-09-18
Exception handling modifications

tgianos:
Attempts to somewhat unify exceptions particularly those thrown out of `JobResolverService` to enable behavior between V3 and V4 execution to expose the same contract going forward.


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -151,6 +151,78 @@ public File createJobDirectory(final JobSpecification jobSpecification) throws S
         return setupFiles;
     }
 
+    public Map<String, String> setupJobEnvironment(
```
mprimi: Our IDEs are still on different style settings i need to copy yours.

genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/GenieExceptionMapper.java
```diff
@@ -69,32 +68,28 @@ public GenieExceptionMapper(final MeterRegistry registry) {
...
-     * @throws IOException on error in sending error
+     * @param e The exception to handle
+     * @return An {@link ResponseEntity} instance
      */
     @ExceptionHandler(GenieException.class)
-    public void handleGenieException(
-        final HttpServletResponse response,
-        final GenieException e
-    ) throws IOException {
+    public ResponseEntity<Object> handleGenieException(final GenieException e) {
```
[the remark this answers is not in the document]
tgianos: Transparent to clients. It just makes it easier to work with what we want to send back and abstracts away the details of working with the primitive `HttpServletResponse` directly.

The `JobRestController` integration tests don't use the java client they send actual http requests and read the response so they validate that the HTTP responses haven't changed.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
tgianos: ok lets look at it tomorrow

PR #914, opened 2019-09-23
Minor updates to API

mprimi:


genie-common/src/main/java/com/netflix/genie/common/dto/Application.java
```diff
@@ -81,10 +81,10 @@ protected Application(final Builder builder) {
          * @param status  The status of the Application
          */
         public Builder(
-            @JsonProperty("name") final String name,
-            @JsonProperty("user") final String user,
-            @JsonProperty("version") final String version,
-            @JsonProperty("status") final ApplicationStatus status
+            @JsonProperty(value = "name", required = true) final String name,
```
tgianos: While I understand the intent behind this change it seems more like a shortcoming of `GenieExceptionMapper` then anything specific with where it is enforced. The plain/text likely comes to how we're building the `ResponseEntity`.

This logic is somewhat conflicting now cause the error messages embedded in the `javax.annotation` ones are more specific. e.g. `A user is required and must be at most 255 characters` in your example vs. `Missing required creator property 'user' `.

Changing the response code from `412` to `400` doesn't feel like a minor change either if clients have specific handling for different 4xx codes.

mprimi: I agree but 400 is more specific and more correct, and i'm expecting any client to be already handling 'bad request'.

PR #915, opened 2019-09-23
Changes to continue supporting V3 embedded execution but allow for Agent execution

tgianos:
Lots of changes that enable a job to be executed with the traditional V3 execution but allow framework for executing a job via the Agent instead.

The agent path was tested with an old commit that parsed string command arguments into a list and it worked but that commit broke internal users. The Agent path won't work properly till this is resolved.

Opening pull request to get this large amount of changes in the shared code base to allow iteration and not diverge a great deal.


genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
```diff
@@ -48,11 +49,12 @@
...
     private static final String EXEC_COMMAND = "exec";
     private static final String SERVER_HOST_OPTION = "--serverHost";
-    private static final String SERVER_HOST_VALUE = "localhost";
+    @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
+    private static final String SERVER_HOST_VALUE = "127.0.0.1";
     private static final String SERVER_PORT_OPTION = "--serverPort";
     private static final String API_JOB_OPTION = "--api-job";
     private static final String JOB_ID_OPTION = "--jobId";
-    private static final String FULL_CLEANUP_OPTION = "--full-cleanup";
+    private static final String FULL_CLEANUP_OPTION = "--no-cleanup";
```
mprimi: Variable name doesn't match value. Maybe a leftover from a test?

genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/dtos/DtosAutoConfiguration.java
```diff
@@ -0,0 +1,102 @@
+/*
```
mprimi: I'd rename this to something else.. `ServerBinding` maybe? `ServerAddressInfo`?
Because it's not really configuring anything about DTOs, it's just creating 2 objects that happen to be DTOs (but could also not be..)

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/LocalAgentLauncherImpl.java
tgianos: Yeah I think this is a remnant of debugging but I want to leave it for now cause I do want `no-cleanup` so we can check the job directory etc as we do dev.

Eventually we probably want these to be some sort of properties but that's part of ongoing improvement

genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/dtos/DtosAutoConfiguration.java
tgianos: Yeah. I've generally tried to name the auto configurations to match the "module" that they're tied to but you bring up a good point about these two objects. They probably should be moved as they're more "introspection" or "metadata" type classes meant for runtime information about the web server. Let me add another commit that moves them and this config somewhere else.

PR #917, opened 2019-09-27
Customize Amazon SNS client so it automatically retries in case of error

mprimi:


genie-web/src/test/java/com/netflix/genie/web/spring/autoconfigure/aws/AWSAutoConfigurationTest.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    @Test
+    public void testExpectedContextWhenSNSDisabled() {
+        this.contextRunner
+            .withPropertyValues(
+                "genie.notifications.sns.enabled=false"
+            )
+            .run(
+            (context) -> {
+                System.out.println(context);
```
tgianos: Probably didn't mean to leave this in here?

mprimi: Whoops.. 👍

mprimi:
@tgianos I had to change this since your review. 
Deployment was failing due to multiple beans of type `AmazonSNS` being found by some Spring Cloud configuration that expects a single one.
The only way i found of resolving this was to rename the custom SNS client so it overrides the one created by Spring Cloud. This results in a single client bean the custom one defined here.

PR #918, opened 2019-10-04
More work to enable agent execution behind V3 API

mprimi:
- Solve race condition encountered in some tests attempting to download agent files immediately after they're created
 - Utilities to perform job arguments tokenization
 - Execute V3 job request in V4 (agent) mode
 - Update Job controller tests


genie-web/src/integTest/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestControllerIntegrationTest.java
```diff
@@ -267,12 +267,7 @@ public void testForTooManyCommandArgs() throws Exception {
 
     private void submitAndCheckJob(final int documentationId, final boolean archiveJob) throws Exception {
         Assume.assumeTrue(SystemUtils.IS_OS_UNIX);
-        final List<String> commandArgs;
-        if (this.agentExecution) {
-            commandArgs = Lists.newArrayList("-c", "echo hello world");
-        } else {
-            commandArgs = Lists.newArrayList("-c", "'echo hello world'");
-        }
+        final List<String> commandArgs = Lists.newArrayList("-c", "echo hello world");
```
tgianos: Would we prefer the tests to match how they were in the past? e.g. `-c 'echo hello world'`?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -137,6 +137,21 @@ public synchronized void stop() {
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public synchronized void forceServerSync() {
```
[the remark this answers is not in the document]
mprimi: If we invalidate every time we ask for a manifest, then there's no point in having caching logic there at all.
I think caching is good in general, as it avoid excessive I/O.

genie-web/src/integTest/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestControllerIntegrationTest.java
mprimi: I don't think that's possible because we lose the quotes in the process.
So we could keep the two execution modes the same without quotes, or we need split logic for the two cases and keep the quotes for the legacy path.

genie-web/src/integTest/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestControllerIntegrationTest.java
```diff
@@ -726,7 +721,7 @@ private void checkJobExecution(final int documentationId, final String id) {
         if (this.agentExecution) {
             validatableResponse
                 .body(HOST_NAME_PATH, Matchers.notNullValue())
-                .body(PROCESS_ID_PATH, Matchers.notNullValue())
```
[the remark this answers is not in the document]
mprimi: The agent does send its PID as part of `ReserveJob` or `ClaimJob`, at least according to the protocol. I'm gonna dig a bit and see why they're not coming back here.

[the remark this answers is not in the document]
mprimi: There's a bug somewhere. When transitioning to `CLAIMED` the agent sends its PID and it is persisted, but later it gets overwritten to `NULL` by some other persistence call. Gonna keep digging..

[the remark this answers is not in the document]
mprimi: @tgianos mystery solved.. took me a minute because I wasn't really expecting this :-) 

Snippet from `JobEntity.java`:
```java

    @Basic
    @Column(name = "agent_pid")
    @Min(0)
    private Integer agentPid;

    @Basic
    @Column(name = "process_id")
    private Integer processId;
```

There are 2 entity fields (and table columns) to store PID!

`processId` is set by the V3 codepath, and then exposed in the `JobExecution` DTO.
`agent_pid` is set by the V4 codepath, but it's not exposed through any DTO or API.

The new column was introduced here: https://github.com/Netflix/genie/commit/88cfe13

Do you remember what's the rationale for adding a separate column?

We should either expose it through the 'JobExecution` DTO or stop using it and keep the original `processId`.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/jobs/JobConstants.java
```diff
@@ -292,6 +292,16 @@
...
 
+    /**
+     * Environment variable for the Genie username the job request.
+     */
+    public static final String GENIE_USER_ENV_VAR = "GENIE_USER";
+
+    /**
+     * Environment variable for the Genie user group the job request.
+     */
+    public static final String GENIE_USER_GROUP_ENV_VAR = "GENIE_USER_GROUP";
```
tgianos: do you think this is useful to anyone? I think we only really need the user

genie-web/src/integTest/resources/com/netflix/genie/web/apis/rest/v3/controllers/JobRestControllerIntegrationTests/non-linux-runsh.txt
```diff
@@ -1,132 +0,0 @@
-#!/usr/bin/env bash
```
tgianos: Shouldn't the V3 execution tests still be comparing against these files?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/jobs/JobConstants.java
mprimi: Adding both for consistency, since they're tightly related (agreed internally nobody is using this..)

genie-web/src/integTest/resources/com/netflix/genie/web/apis/rest/v3/controllers/JobRestControllerIntegrationTests/non-linux-runsh.txt
mprimi: I found no reference to these files and as you can see tests pass without them...
I assumed they got removed some time ago, tho i don't remember specifically.

tgianos: It looks like I removed it in https://github.com/Netflix/genie/commit/3cc8ee9c799ae868789771d61e2c9976e0d027dc during all those massive changes but forgot to put it back in once the dual path was supported with a flag. We probably should put it back in

```// check that the generated run file is correct
        final String runShFileName = SystemUtils.IS_OS_LINUX ? "linux-runsh.txt" : "non-linux-runsh.txt";

        final String runShFile = this.resourceLoader
            .getResource(BASE_DIR + runShFileName)
            .getFile()
            .getAbsolutePath();
        final String runFileContents = new String(Files.readAllBytes(Paths.get(runShFile)), StandardCharsets.UTF_8);

        final String jobWorkingDir = this.jobDirResource.getFile().getCanonicalPath() + FILE_DELIMITER + id;
        final String expectedRunScriptContent = this.getExpectedRunContents(runFileContents, jobWorkingDir, id);

        RestAssured
            .given(this.getRequestSpecification())
            .filter(fileResultFilter)
            .when()
            .port(this.port)
            .get(JOBS_API + "/{id}/output/{filePath}", id, "run")
            .then()
            .statusCode(Matchers.is(HttpStatus.OK.value()))
            .body(Matchers.is(expectedRunScriptContent));```

mprimi: I'll take this commit out and add this missing code back in.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/jobs/JobConstants.java
tgianos: I think this concept will go away in agent execution so perhaps we shouldn't do it? It doesn't harm anything I just don't want anyone to start relying on it and then pull it out

mprimi: We could keep this as a concept/model and API level as a way to express the user belongs to a specific group of users.

PR #919, opened 2019-10-18
Archived job retrieval improvements

tgianos:
Primarily this PR contains improvements to how retrieving the manifests of archived jobs are handled.

Previously it used an embedded Guava `LoadingCache` but this cache wasn't configurable nor where there retries. When the manifests were stored in an eventually consistent system (like `S3`) there is a chance the manifest won't show up on a request that is made quickly after job is completed.

This PR changes this a bit. It breaks the retrieval of Archived job metadata into a separate service that allows proxying of beans. It then enables `retry` and `cache` features from Spring Boot on this method. Reconfiguring the project a bit to make `retry` and `caching` available via annotations will likely save us from redoing work like this in the future.

Other minor changes are included as separate commits.


genie-web/src/main/java/com/netflix/genie/web/dtos/ArchivedJobMetadata.java
```diff
@@ -0,0 +1,42 @@
...
+ * @since 4.0.0
+ */
+@Getter
+@AllArgsConstructor
+@EqualsAndHashCode(doNotUseGetters = true)
+@ToString(doNotUseGetters = true)
+public class ArchivedJobMetadata {
+    private final String jobId;
+    private final DirectoryManifest manifest;
+    private final URI jobDirectoryRoot;
```
mprimi: Maybe `archiveBaseURL` would be more clear?

Either way, this makes assumptions about how the manifest is stored.
So maybe just `url` or `location`?

genie-web/src/main/java/com/netflix/genie/web/services/impl/ArchivedJobServiceImpl.java
```diff
@@ -0,0 +1,149 @@
...
+
+            final URI manifestLocation;
+            if (StringUtils.isBlank(JobArchiveService.MANIFEST_DIRECTORY)) {
+                manifestLocation = jobDirectoryRoot.resolve(JobArchiveService.MANIFEST_NAME).normalize();
+            } else {
+                manifestLocation = jobDirectoryRoot
+                    .resolve(JobArchiveService.MANIFEST_DIRECTORY + SLASH)
+                    .resolve(JobArchiveService.MANIFEST_NAME)
+                    .normalize();
+            }
```
mprimi: Since this is coupled with JobArchiveService anyway (statically), why not just delegate this logic to it (eg. through a static helper method)?

Or maybe even better extract into a "manifest convention" thing that is used by both

genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/RetryAutoConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Auto configuration which enables {@link org.springframework.retry.annotation.Retryable} for method calls on beans.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+@Configuration
+@EnableRetry
+public class RetryAutoConfiguration {
+}
```
mprimi: I'm assuming this won't have any effect on the existing `retryTemplate` we have around?

genie-web/src/main/resources/genie-web-defaults.yml
```diff
@@ -118,6 +118,9 @@ spring:
     name: genie
   banner:
     location: genie-banner.txt
+  cache:
+    caffeine:
+      spec: maximumSize=100,expireAfterAccess=600s
```
mprimi: Quick back-of-the-envelope calculation specific to our internal deployment/usage.

We peak at 1.2k jobs daily in tests, with a 10 nodes cluster.
If clients are polling for outputs (which they are), the manifest cache may reach 120 'hot' elements.
Depending on the frequency of such requests, this may lead to some trashing in the cache.

i.e. this default value is a bit low for our own deployment.
Just something to keep in mind, we may want to double it when rolling this out.

genie-web/src/main/java/com/netflix/genie/web/dtos/ArchivedJobMetadata.java
```diff
@@ -0,0 +1,42 @@
...
+ * A simple POJO for a compound value of related information to a job archived location and files.
+ *
+ * @author tgianos
+ * @since 4.0.0
+ */
+@Getter
+@AllArgsConstructor
+@EqualsAndHashCode(doNotUseGetters = true)
+@ToString(doNotUseGetters = true)
+public class ArchivedJobMetadata {
```
mprimi: After going through the rest of the commits in this PR, i'm not sure what is the value of exposing this, as opposed to just returning the manifest.
The job id is the key used to look this up, so it's redundant probably.
And the URL is only used by implementations to retrieve, no need to expose it.

I'm not asking to change this, it's fine. But maybe i'm missing something because this DTO seems redundant.

genie-web/src/main/java/com/netflix/genie/web/dtos/ArchivedJobMetadata.java
tgianos: Fair I'll change it this is the result of a copy paste from a reuse location

genie-web/src/main/java/com/netflix/genie/web/services/impl/ArchivedJobServiceImpl.java
tgianos: Yeah I'm not thrilled with this. I believe I left it in here purposely to refine it somehow in the future. right now with the code being shared still if I put it in a utility method it won't break anything when we move the `JobArchiveService` only into the Agent code base and we'll forget about this weak coupling. We need to do something better. Possibly some new field in the protocol that tells the agent where to store the archive metadata.

genie-web/src/main/resources/genie-web-defaults.yml
tgianos: Yeah this is just the default we can change it in internal

genie-web/src/main/java/com/netflix/genie/web/dtos/ArchivedJobMetadata.java
tgianos: The way that `JobDirectoryServerServiceImpl` is currently written needs the baseURL of wherever the job directory currently resides in order to build up the correct redirection urls. As this `ArchivedJobMetadataService` now abstracts details of where that resides it has to return it somehow or the shared code between active and archived job directory serving will break. This whole thing needs some cleanup after the code split happens.

PR #920, opened 2019-10-21
Job INIT kill bug, client bug, verbosity adjustments

mprimi:


genie-web/src/main/java/com/netflix/genie/web/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -152,11 +151,6 @@ public void executeTask(@NotNull final Map<String, Object> context) throws Genie
                 .redirectOutput(new File(jobExecEnv.getJobWorkingDir() + JobConstants.GENIE_LOG_PATH))
                 .redirectError(new File(jobExecEnv.getJobWorkingDir() + JobConstants.GENIE_LOG_PATH));
 
-            //
-            // Check if file can be executed. This is to fix issue where execution of the run script fails because
```
tgianos: We're sure this isn't an issue still?

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -574,7 +574,7 @@ public JobStatus waitForCompletion(final String jobId, final long blockTimeout,
             if (System.currentTimeMillis() - startTime < blockTimeout) {
                 Thread.sleep(pollTime);
             } else {
-                throw new GenieTimeoutException("Timed out waiting for job to finish");
+                throw new GenieTimeoutException("Timed out waiting for job to finish: " + jobId);
```
tgianos: This happens in the logs but the code is on the client side. Are you sure this is the error log message you think it is?

genie-web/src/main/java/com/netflix/genie/web/jobs/workflow/impl/JobKickoffTask.java
mprimi: I really don't see how `touch` would have made any difference.

But i did deploy to my cluster and run the smoke tests.

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
mprimi: Yes, smoke tests and gradle plugin test for example print this but no idea what the corresponding job is.

PR #923, opened 2019-11-06
Add option to change job working directory on launch

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/cli/RuntimeConfigurationArgumentsImpl.java
```diff
@@ -0,0 +1,38 @@
...
+ * @since 4.0.0
+ */
+@Getter
+class RuntimeConfigurationArgumentsImpl implements ArgumentDelegates.RuntimeConfigurationArguments {
+
+    @Parameter(
+        names = {"--launchInJobDirectory"},
+        description = "Whether the job process should be forked within the job directory"
+    )
+    private boolean launchInJobDirectory;
```
tgianos: Should the default be `true` and then have the bdp-cli just always pass in true or vice versa?

mprimi: Default `true` would be a change of default behavior so I don't think it should be.
But this option should be in the default launch argument of the agent launcher, which i forgot.

PR #925, opened 2019-11-14
Launch job by invoking Bash explicitly, and minor other improvements

mprimi:


genie-web/src/main/java/com/netflix/genie/web/jobs/workflow/impl/JobKickoffTask.java
```diff
@@ -151,6 +151,10 @@ public void executeTask(@NotNull final Map<String, Object> context) throws Genie
                 command.add(user);
             }
 
+            // Launch via bash rather than executing the run script directly to avoid 'Text file busy' error.
+            // https://bugs.openjdk.java.net/browse/JDK-8068370
+            command.add("bash");
```
tgianos: Could make `bash` a constant but whatever.

PR #928, opened 2019-12-10
Generalize mechanism to load script extensions (and more)

mprimi:


genie-agent-app/src/smokeTest/java/com/netflix/genie/GenieAgentApplicationSmokeTest.java
```diff
@@ -70,15 +70,15 @@ public void smokeTestCommands() throws Exception {
         // Test exec
         runner.run(
             "exec",
-            "--clusterCriterion",
+            "--cluster-criterion",
```
tgianos: It might be good to mix use of camel and dash in here

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -368,6 +368,11 @@ of 2.0
 |5
 |no
 
+|genie.scripts-manager.refresh-interval
```
tgianos: is there any desire to have this on a per script basis instead of globally refresh all scripts at the same interval

genie-web/src/main/java/com/netflix/genie/web/scripts/ManagedScript.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @PostConstruct
+    void loadPostConstruct() {
+        final URI scriptUri = this.properties.getSource();
+        if (this.properties.isAutoLoadEnabled() && scriptUri != null) {
+            this.scriptManager.manageScript(scriptUri);
+        }
+    }
+
+    protected Object evaluateScript(
```
tgianos: this may not be pertinent to this PR but part of me wonders if we should standardize on one type of script (`groovy` for example) rather than just supporting any kind of script. This might allow us to provide a better contract here for scripts. I'm a little concerned that this abstracts it to such a level that there's really no contract that we can enforce. It might be nice in the future to have say "ok if you want a `ClusterSelector` script here are your input object types and the expected output type is a `Cluster` or something concrete instead of just Object.

I mean I guess I could see it being super generic here (kind of pythonic) but we do lose something by doing that while gaining the reusability. Maybe templetize `ManagedScript` somehow?

genie-web/src/main/java/com/netflix/genie/web/scripts/ManagedScriptBaseProperties.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Base abstract properties for individual script classes to extend.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Getter
+@Setter
+@Validated
+public abstract class ManagedScriptBaseProperties {
```
tgianos: Are these documented somewhere? I don't see it in this commit in the doc changes

genie-web/src/main/java/com/netflix/genie/web/scripts/ManagedScriptBaseProperties.java
```diff
@@ -0,0 +1,43 @@
...
+    protected static final String SCRIPTS_PREFIX = "genie.scripts";
+    protected static final String SOURCE_PROPERTY_SUFFIX = ".source";
+    @Nullable
+    private URI source;
+    private long timeout = 5_000L;
+    private boolean autoLoadEnabled = true;
```
tgianos: Can't we add the refresh rate here so it's on a per script basis?

genie-web/src/main/java/com/netflix/genie/web/scripts/ScriptManager.java
```diff
@@ -0,0 +1,284 @@
...
+    protected Object evaluateScript(
+        final URI scriptUri,
+        final Bindings bindings,
+        final long timeout
+    ) throws ScriptNotConfiguredException, ScriptExecutionException {
+
+        final AtomicReference<CompiledScript> compiledScriptReference = this.scriptsMap.get(scriptUri);
+
+        if (compiledScriptReference == null) {
+            throw new ScriptNotConfiguredException("Unknown script: " + scriptUri);
```
tgianos: Wouldn't it be good to wrap these in the try finally so there are metrics collected on these exceptions being thrown?

genie-web/src/main/java/com/netflix/genie/web/scripts/ScriptManager.java
```diff
@@ -0,0 +1,284 @@
...
+                // We want a compilable engine so we can cache the script
+                throw new ScriptLoadingException(
+                    "Script engine for file extension: " + scriptExtension + " is not " + Compilable.class.getName()
+                );
+            }
+
+            final Compilable compilable = (Compilable) engine;
+
+            final InputStream scriptInputStream;
+            try {
```
tgianos: This should probably be a try with resources

genie-web/src/main/java/com/netflix/genie/web/scripts/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+/**
+ * Package for script loaded at runtime to provide custom extensions and behavior.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+
```
tgianos: extra space

genie-web/src/test/groovy/com/netflix/genie/web/scripts/ScriptManagerSpec.groovy
```diff
@@ -0,0 +1,469 @@
...
+        1 * meterRegistry.timer(
+            ScriptManager.SCRIPT_EVALUATE_TIMER_NAME,
+            {
+                Set<Tag> tags -> tags.containsAll(evalSuccessTags)
+            }
+        ) >> timer
+        1 * timer.record({ it > 0 }, TimeUnit.NANOSECONDS)
+        result == evaluationResult
+    }
+//
```
tgianos: A lot of commented out code here. If you have a test you want to skip there's a way to do that with spock

genie-web/src/integTest/java/com/netflix/genie/web/scripts/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+/**
+ * Integration tests for script components.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+
```
tgianos: Extra line

genie-web/src/main/java/com/netflix/genie/web/scripts/ClusterLoadBalancerScript.java
```diff
@@ -0,0 +1,114 @@
...
+ * <p>
+ * The contract between the script and the Java code is that the script will be supplied global variables
+ * {@code clusters} and {@code jobRequest} which will be JSON strings representing the list (array) of clusters
+ * matching the cluster criteria tags and the job request that kicked off this evaluation. The code expects the script
+ * to either return the id of the cluster if one is selected or null if none was selected.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public class ClusterLoadBalancerScript extends ManagedScript {
```
tgianos: Since you're making this a new class anyway I'd planned to rename this class / feature `ClusterSelector` instead of `ClusterLoadBalancer` as part of some of the resource data model changes we've been discussing. Might be able to kill two birds with one stone here

genie-web/src/test/groovy/com/netflix/genie/web/spring/autoconfigure/scripts/ScriptsAutoConfigurationSpec.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+import com.netflix.genie.web.properties.ClusterLoadBalancerScriptProperties
+import com.netflix.genie.web.properties.ScriptManagerProperties
+import com.netflix.genie.web.scripts.ScriptManager
+import io.micrometer.core.instrument.MeterRegistry
+import org.springframework.core.io.ResourceLoader
+import org.springframework.scheduling.TaskScheduler
+import spock.lang.Specification
+
+class ScriptsAutoConfigurationSpec extends Specification {
```
tgianos: There is a much better way to test this which actually tests it in a spring environment. You can look at some of my other test classes for AutoConfigurations

mprimi:
Still working on this. Temporarily closing PR to avoid spamming.

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: As discussed in person: per-script refresh would high-effort, low-reward, so leaving it out, until we have a strong use-case for it.

genie-web/src/main/java/com/netflix/genie/web/scripts/ManagedScript.java
mprimi: Right now that contract is one layer above in concrete `*Script` class that extends the base `ManagedScript`. Users of the latter pass typed parameters and receive a result of the expected type.

But i agree we if we could reduce support to just Groovy, we could do some nicer things.

genie-web/src/test/groovy/com/netflix/genie/web/scripts/ScriptManagerSpec.groovy
mprimi: Deleted, didn't intend to commit commented out code.

genie-web/src/main/java/com/netflix/genie/web/scripts/ClusterLoadBalancerScript.java
mprimi: So there's `ClusterLoadBalancer` interface with its implementations, and there's `ClusterLoadBalancerScript` (this newly added class that wraps a script resource).

Which one you want to rename? @tgianos

tgianos: I'd like to rename `ClusterLoadBalancer` to `ClusterSelector` as it's clearer the purpose and then as fallout of that we probably should rename the implementations as well.

genie-web/src/test/groovy/com/netflix/genie/web/spring/autoconfigure/scripts/ScriptsAutoConfigurationSpec.groovy
mprimi: 👍

genie-web/src/main/java/com/netflix/genie/web/scripts/ClusterLoadBalancerScript.java
mprimi: Ok will do (as last commit of this PR, to avoid making other commits hard to read)

PR #930, opened 2019-12-17
Agent execution script filter (and more)

mprimi:


mprimi:
Major differences from previous iteration:
* Last 2 commits add then use a new script to filter agent execution (the main purpose of this PR).
* Addressed comments from previous review

New commits with minor changes:
* `Apply test settings to all Gradle Java test configurations`
* `Lower verbosity of UserMetricsTask`

@tgianos

genie-web/src/integTest/resources/application-integration.yml
```diff
@@ -29,6 +29,9 @@ genie:
     cluster-load-balancer:
       source: file:///tmp/genie/loadBalancers/script/source/loadBalance.js
       auto-load-enabled: true
+    execution-mode-filter:
+      source: file:///tmp/genie/loadBalancers/script/source/executionModeFilter.js
```
tgianos: when we change the name of `ClusterLoadBalancer` to `ClusterSelector` we probably want to change these paths too.

genie-web/src/main/java/com/netflix/genie/web/scripts/ExecutionModeFilterScript.java
```diff
@@ -0,0 +1,91 @@
...
+        super(scriptManager, properties, mapper, registry);
+    }
+
+    /**
+     * Evaluate the script and return true if this job should be forced to execute via agent, false if it should be
+     * forced to execute in embedded mode, null if the script decides not explicitly flag this job for one or the other
+     * execution mode.
+     *
+     * @param jobRequest the job request
+     * @return An optional boolean valua
```
tgianos: typo here for `value`

genie-web/src/main/java/com/netflix/genie/web/scripts/ExecutionModeFilterScript.java
```diff
@@ -0,0 +1,91 @@
...
+     * @throws ScriptNotConfiguredException if the script is notyet successfully loaded and compiled
+     * @throws ScriptExecutionException     if the script evaluation produces an error
+     */
+    public Optional<Boolean> forceAgentExecution(
```
tgianos: Why does it return an `Optional`? Sorry if this is in another part of this PR but wouldn't it just be true/false or exception? Are you using `Optional.empty()` to represent no preference?

genie-web/src/test/resources/com/netflix/genie/web/scripts/execution-mode-filter.groovy
```diff
@@ -0,0 +1,32 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
```
tgianos: year is off

genie-web/src/main/java/com/netflix/genie/web/scripts/ExecutionModeFilterScript.java
```diff
@@ -0,0 +1,91 @@
...
+ * The contract between the script and the Java code is that the script will be supplied global variables
+ * {@code jobRequest} which will be JSON strings representing job request that kicked off this evaluation.
+ * The code expects the script to either return the true (to force agent execution), false (to force legacy/embedded
+ * execution) or null (for no preference).
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public class ExecutionModeFilterScript extends ManagedScript {
```
tgianos: Shouldn't this be implementing some other interface reflecting that it is some sort of `ExecutionModeSelector`? Not sure if we have that for the Headers or Property based paths?

genie-web/src/integTest/resources/application-integration.yml
mprimi: Ah yeah, actually not even sure why these are there given that these files won't exist. Test that the app starts and work regardless?

genie-web/src/main/java/com/netflix/genie/web/scripts/ExecutionModeFilterScript.java
mprimi: Yep

> The code expects the script to either return the true (to force agent execution), false (to force legacy/embedded execution) or null (for no preference).

genie-web/src/main/java/com/netflix/genie/web/scripts/ExecutionModeFilterScript.java
mprimi: This is just a script object. 
The execution mode is a different thing and uses this.

PR #932, opened 2019-12-23
Bug: REST API produces 406 for file not found

mprimi:
This PR addresses an observed misbehavior of the REST API.

When a file is requested that does not exist (or is not yet present in the manifest), the API produces a 406 error (content-negotiation failure), rather than the expected 404.

This PR adds tests and does some refactoring to solve this problem.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/data/services/JobPersistenceService.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/data/services/JobPersistenceService.java
@@ -273,11 +273,11 @@ public interface JobPersistenceService {
      *
      * @param id The id of the job
      * @return true if its a v4 job
-     * @throws GenieJobNotFoundException If no job with {@code id} exists
+     * @throws GenieNotFoundException If no job with the given {@code id} exists
      */
     boolean isV4(
         @NotBlank(message = "Id is missing and is required") String id
-    );
+    ) throws GenieNotFoundException;;
 
     /**
      * Get the status for a job with the given {@code id}.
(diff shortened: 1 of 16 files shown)


genie-web/src/main/java/com/netflix/genie/web/data/services/JobPersistenceService.java
```diff
@@ -273,11 +273,11 @@ void updateJobStatus(
      *
      * @param id The id of the job
      * @return true if its a v4 job
-     * @throws GenieJobNotFoundException If no job with {@code id} exists
+     * @throws GenieNotFoundException If no job with the given {@code id} exists
```
tgianos: Technically this was a new API and the other APIs are supporting old v3 job execution paths that were going to be deleted when migration was complete. I know you're adamant about your love for checked exceptions so I'm not going to argue that again cause it's not worth it but we should at least make this a checked "JobNotFoundException" so that we can be clear about what isn't found for APIs at higher levels that compound multiple actions. `GenieNotFoundException` as it bubbles up will be lost as to whether it was a job, cluster, command, file, etc.

genie-web/src/main/java/com/netflix/genie/web/services/JobDirectoryServerService.java
```diff
@@ -17,12 +17,11 @@
  */
 package com.netflix.genie.web.services;
 
+import com.netflix.genie.common.exceptions.GenieException;
```
tgianos: I'd like to get away from using the exceptions in the old common module as we move towards hopefully deleting it at the completion of v4 development. It doesn't have to be part of this commit but we should move to exceptions in either `common-external`, `common-internal`, `web` or `agent` so they're scoped properly.

PR #933, opened 2020-01-06
Decouple resource status enums from database representation

tgianos:
The primary purpose of this PR is to remove the direct dependency between the various DTO status enums (e.g. `JobStatus`) from the database tier (i.e. using the status enumerations in the entity classes).

This tight coupling made it hard to migrate or support removing enum values

Additionally this PR contains many refactoring changes to prep for the (hopefully) final V4 development thrust. Moving V4 DTO classes into the `common-external` module will allow us to use them in an eventual V4 client as well as in the eventual REST APIs

Other various improvements and cleanup discovered while doing these two changes


build.gradle
```diff
@@ -315,6 +315,9 @@ configure(javaProjects) {
     spotbugs {
         excludeFilter = new File(project.parent.projectDir, "codequality/spotbugs/excludeFilter.xml")
     }
+    spotbugsTest.enabled = false
```
mprimi: I think there's value in keeping this. How much does it shave off the build time?

tgianos:
> Some of these changes are a little scary... 🤞

Ha I agree especially since I did some of them during some awesome bouts of insomnia

build.gradle
tgianos: On successful builds it's ~2 minutes on my local machine however i ran into multiple instances where despite increasing the heap size of the gradle JVM the spotbugs process would just get stuck infinitely before completely still blowing away the heap. For now I've turned it off for the test files. I tried upgrading spotbugs to the latest version but that ran into other plugin classpath issues. @chali was looking at that and maybe if he gets it resolved we can put this back in and see if it's still a problem

PR #935, opened 2020-01-07
Upgrade to Gradle 6 with minimum required versions of plugins

chali:


build.gradle
```diff
@@ -17,20 +17,15 @@ buildscript {
         classpath "com.netflix.nebula:gradle-aggregate-javadocs-plugin:3.0.1"
         classpath "com.google.protobuf:protobuf-gradle-plugin:0.8.8"
         classpath "io.franzbecker:gradle-lombok:3.2.0"
-        /*
-         * TODO: Upgrading to 1.3.1 caused problems
-         * A problem was found with the configuration of task ':genie-ui:npmInstall'.
-         * > File '{somePath}/genie/genie-ui/package-lock.json' specified for property '$2' does not exist.
-         */
-        classpath "com.moowork.gradle:gradle-node-plugin:1.2.0"
+        classpath "com.netflix.nebula:nebula-node-plugin:1.3.0"
```
chali: This is a fork of original `gradle-node-plugin` which is not maintained anymore and it is not compatible with Gradle 6.

codequality/checkstyle/checkstyle.xml
```diff
@@ -164,12 +164,6 @@
         <!-- Checks the placement of left curly braces on types, methods and other blocks. -->
         <module name="LeftCurly"/>
 
-        <!-- Checks for long lines. -->
```
chali: This change is required by breaking change in checkstyle which is upgraded in Gradle.

build.gradle
tgianos: Will this be maintained or should we be looking into some other options?

build.gradle
```diff
@@ -17,20 +17,15 @@ buildscript {
...
         /*
          * TODO: Failed when upgrading to 2.0.0. Looks like a Guava version problem (surprise)
          * Execution failed for task ':genie-agent:spotbugsIntegTest'.
          * > Failed to run Gradle SpotBugs Worker
          *   > Could not create an instance of type com.github.spotbugs.internal.spotbugs.SpotBugsExecutor.
          *      > com.google.common.collect.ImmutableList.builderWithExpectedSize(I)Lcom/google/common/collect/ImmutableList$Builder;
          */
-        classpath "gradle.plugin.com.github.spotbugs:spotbugs-gradle-plugin:1.6.9"
+        classpath "com.github.spotbugs:spotbugs-gradle-plugin:3.0.0"
```
tgianos: If this fixes the issue we can probably get rid of the big comment block above this.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0-bin.zip
```
tgianos: the latest appears to be `6.0.1`. should we just go ahead and do that as part of this?

chali: sounds good. Let me update that.

build.gradle
chali: Yes, we plan to maintain compatibility. We don't plan any feature development.

chali:
@tgianos `npm-shrinkwrap.json` has now its original content.

PR #936, opened 2020-01-09
Published pom.xml doesn't have proper metadada when using deprecated runtime gradle configuration

chali:


genie-web/build.gradle
```diff
@@ -85,10 +85,10 @@ dependencies {
      * Runtime Dependencies
      *******************************/
 
-    runtime("com.h2database:h2")
-    runtime("mysql:mysql-connector-java")
-    runtime("org.codehaus.groovy:groovy-all")
-    runtime("org.postgresql:postgresql")
+    runtimeOnly("com.h2database:h2")
```
tgianos: Did you check if this is the only place we use `runtime` in the project anymore by chance?

chali: I did and now I see I haven't saved one file. It is for tests though so it would have affect publishing, but I will add it for consistency's sake.

PR #937, opened 2020-01-10
Add cluster criteria to commands

tgianos:
This PR is the first step into switching how `clusters` and `commands` are resolved for a job. This adds a new optional priority ordered list of `criterion` to the classic V3 `command` resource. Eventually this list will be used but for now this PR adds all the database tables, CRUD APIs, model changes, etc to support storing and manipulating this list.


genie-web/src/main/java/com/netflix/genie/web/data/entities/CommandEntity.java
```diff
@@ -274,6 +289,68 @@ protected void setClusters(@Nullable final Set<ClusterEntity> clusters) {
...
+     * Add a new cluster criterion with the given priority.
+     *
+     * @param criterion The new criterion to add
+     * @param priority  The priority with which this criterion should be considered. {@literal 0} would be the highest
+     *                  priority and anything greater than the current size of the existing criteria will just be added
+     *                  to the end of the list. If a priority of {@code < 0} is passed in the criterion is added
+     *                  as the highest priority ({@literal 0}).
+     */
+    public void addClusterCriterion(final CriterionEntity criterion, final int priority) {
+        if (priority <= HIGHEST_CRITERION_PRIORITY) {
```
mprimi: This only works if priority is 0, and throws if it's negative.
Maybe the condition should be `< size`?

genie-web/src/integTest/java/com/netflix/genie/web/apis/rest/v3/controllers/RestControllerIntegrationTestBase.java
```diff
@@ -196,6 +196,8 @@ public void setup() throws Exception {
                         Preprocessors.modifyUris().host(URI_HOST).scheme(URI_SCHEME).removePort()
                     )
             )
+//            .addFilter(new RequestLoggingFilter())
```
mprimi: Remove these commented out lines?

tgianos: I left them in on purpose so that if we need help debugging something we don't have to continually look up what it is

genie-web/src/main/java/com/netflix/genie/web/data/entities/CommandEntity.java
tgianos: Not sure I follow what you're saying here. Unless I'm missing something the constant `HIGHEST_CRITERION_PRIORITY` is `0` so this is `<=0` it will just use 0 so it won't throw exception. If it's `>=` the current size it appends it at the end. any other case it inserts it. Do you think this is wrong?

PR #940, opened 2020-01-13
Support s3n:// and s3a:// dependency URLs

mprimi:


genie-common-internal/src/main/java/com/netflix/genie/common/internal/aws/s3/S3ProtocolResolver.java
```diff
@@ -63,11 +63,20 @@ public S3ProtocolResolver(
     @Override
     public Resource resolve(final String location, final ResourceLoader resourceLoader) {
         log.debug("Attempting to resolve if {} is a S3 resource or not", location);
+
+        final String normalizedLocation;
+        // Rewrite s3n:// and s3a:// URIs as s3:// for backward compatibility
+        if (location.startsWith("s3n:") || location.startsWith("s3a:")) {
```
tgianos: You probably want these strings to be constants

PR #954, opened 2020-02-18
Implement V3-compatible execution in Agent

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionContext.java
```diff
@@ -165,4 +150,19 @@ void addStateActionError(
      * @param jobId the job id
      */
     void setClaimedJobId(@NotBlank String jobId);
+
```
tgianos: Extra line here

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/LaunchJobAction.java
```diff
@@ -32,7 +32,6 @@
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.File;
-import java.util.Map;
 
 /**
  * Action performed when in state LAUNCH_JOB.
```
tgianos: Somewhere, now that you're relying on bash, don't we need to assert that bash is available or short circuit the agent?

genie-agent/src/test/groovy/com/netflix/genie/agent/cli/JobRequestConverterSpec.groovy
```diff
@@ -131,6 +132,35 @@ class JobRequestConverterSpec extends Specification {
...
+
+        then:
+        1 * jobRequestArgs.getCommandArguments() >> providedJobArguments
+        1 * validator.validate(_ as AgentJobRequest) >> Sets.newHashSet()
+
+        expect:
+        jobRequest.getCommandArgs() == expectedJobArguments
+
+        where:
+        providedJobArguments | expectedJobArguments
```
tgianos: You might want to run the formatter here or this will change later

genie-agent/src/test/resources/JobSetupServiceImplSpec_run1.test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+export GENIE_JOB_DIR="<JOB_DIR_PLACEHOLDER>"
+
+export GENIE_APPLICATION_DIR="${GENIE_JOB_DIR}/genie/applications"
+
+export GENIE_COMMAND_DIR="${GENIE_JOB_DIR}/genie/command/<COMMAND_ID_PLACEHOLDER>"
+
+export GENIE_CLUSTER_DIR="${GENIE_JOB_DIR}/genie/cluster/<CLUSTER_ID_PLACEHOLDER>"
+
+export __GENIE_SETUP_LOG_FILE="${GENIE_JOB_DIR}/genie/logs/setup.log"
```
tgianos: any reason to name these specially with the `__`?

genie-agent/src/test/resources/JobSetupServiceImplSpec_run1.test.sh
```diff
@@ -0,0 +1,53 @@
...
+set -o errexit
+# Error out if any command in a pipeline fails
+set -o pipefail
+# Error out if unknown variable is used
+set -o nounset
+
+
+export GENIE_JOB_DIR="<JOB_DIR_PLACEHOLDER>"
+
+export GENIE_APPLICATION_DIR="${GENIE_JOB_DIR}/genie/applications"
```
tgianos: Should we export variables like `GENIE_APPLICATION_0_ID=` `GENIE_APPLICATION_1_ID=` etc?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/jobs/JobConstants.java
```diff
@@ -422,6 +432,11 @@
     public static final String JOB_SUBMISSION_DISABLED_DEFAULT_MESSAGE
         = "Job submission is currently disabled. Please try again later.";
 
+    /**
+     * The name of the setup file (for entities that have one) when downloaded in the job directory.
+     */
+    public static final String GENIE_ENTITY_SETUP_SCRIPT_FILE_NAME = "genie_setup.sh";
```
tgianos: One thing we want to be careful of is files appended with `.sh` cause the browser to force a download of said file which people have complained about in the past. They'd prefer it was returned as `text/plain` so that it opens within the browser window. If we're going to start putting the proper extension on these files we'll have to modify the file server code on the server to make sure `.sh` is remapped to `text/plain`

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/actions/LaunchJobAction.java
mprimi: It's non trivial... we use `env` in the shebang to ask the shell to run the script with the most appropriate `bash` available. It could be at a non-default location, be named something else other than `bash` and a million other things.

In general, seems to me like trying figure out whether bash is installed is not worth it. 

Moreover, if bash was indeed not installed, a user would get a message in `stdout` (terminal or file depending on `--interactive`.

genie-agent/src/test/groovy/com/netflix/genie/agent/cli/JobRequestConverterSpec.groovy
mprimi: 👍

genie-agent/src/test/resources/JobSetupServiceImplSpec_run1.test.sh
mprimi: My own notation for environment variables that are private in the script.

genie-agent/src/test/resources/JobSetupServiceImplSpec_run1.test.sh
mprimi: The server is generating and sending these over as part of the specification.
As you can see in the test files we compare against (for example `JobControllerIntegrationTest`, everything is there).

genie-common-internal/src/main/java/com/netflix/genie/common/internal/jobs/JobConstants.java
mprimi: This is a browser behavior. Shouldn't we tell those users Genie has no control over browser behavior, but they do?

Moreover, take a look at our current configurations. All setup script i looked at are already named `.sh`. So in practice there is no change of behavior here.

tgianos: Technically it can control the behavior by setting the `Content-Type`.

This is more pertinent for the `run` file as previously it was named `run.sh` till there were many complaints about the behavior and we changed it to `run` and allowed it to be `text/plain`

mprimi: Given how much time we already spend dealing with content-type negotiation corner cases, I'm not a big fan of adding more complexity.
But if we do really want to do it, I'd do it as interceptor that modifies headers when appropriate, rather than as hack in here by giving weird names to thing.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -530,7 +530,28 @@ String composeScript() {
 
             // Script Environment Section
 
-            this.environmentVariables.forEach(
+            sb.append("# Locally-generated environment variables").append(NEWLINE);
+
+            sb.append(NEWLINE);
+
+            this.localEnvironmentVariables.forEach(
```
tgianos: You might want to move the above `sb.append(NEWLINE)` into the loop as the first step and remove the trailing `NEWLINE` from the loop. That will make it so the new lines are balanced. As it is now after the last ENV variable there will be 2 blank lines. Obviously this is a tiny detail just pointing it out.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -547,6 +550,21 @@ String composeScript() {
 
             sb.append(NEWLINE);
 
+            sb
+                .append("# During setup, redirect stdout and stderr of this script to a log file").append(NEWLINE)
```
tgianos: I know we have some comments in the run script today. Do we want to keep them or just remove them and leave them as comments in the java code? It would save serialization/bytes/etc. Could also add some flag to the job request that would enable this eventually optionally. Just a thought.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
mprimi: The double linebreak there was intentional for readability (of the script) since it's between 2 sections of the script.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
mprimi: I like the fact that the script is self-contained and someone looking at it (not us) can understand what these obscure things are, e.g.:
```
exec 1>&6 6>&-
exec 2>&7 7>&-
```

mprimi:
@tgianos added one more commit, the rest you already reviewed and didn't change except for formatting or things you suggested

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -616,6 +632,12 @@ String composeScript() {
 
             sb.append(NEWLINE);
 
+            sb
+                .append("# Setup completed successfully, delete marker file").append(NEWLINE)
+                .append("rm ${").append(SETUP_ERROR_FILE_ENV_VAR).append("}").append(NEWLINE);
```
tgianos: Ha what if this delete fails? (this is a rabbit hole not actually expecting to handle every single little thing)

Not sure if `rm` behaves nicely if a file doesn't exist but if it doesn't you might want to wrap in an `if`

mprimi: I could do `rm ... || true` to make it proceed.

As is, if removal fails, the error will be captured in the setup log, and the job fails.
This seems like the correct behavior to me.

tgianos: Not sure if you’re saying if this fails the job should fail but if so I disagree cause this is just a genie internal cleanup mechanism and if it fails that really shouldn’t affect the users job run as that run can still continue fine right?

PR #955, opened 2020-02-19
Initial implementation of V4 command and cluster resolution logic

tgianos:


genie-web/src/main/java/com/netflix/genie/web/data/repositories/jpa/specifications/JpaCommandSpecs.java
```diff
@@ -125,4 +130,40 @@ private JpaCommandSpecs() {
             return cb.and(predicates.toArray(new Predicate[0]));
         };
     }
+
+    /**
+     * Get the specification for the query which will find the commands which match the given criterion.
```
mprimi: Couple of things that would be nice to comment:
 - Implicitly looks for active
 - Cannot get a query that looks for commands in any state using this (i.e. all commands with tag `foo` regardless of state
 - An idea of the resulting SQL, especially the join/count part 

Just for future readability/understandability.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -92,6 +95,28 @@
     private static final String CLUSTER_COMMAND_QUERY_TIMER_NAME
         = "genie.services.jobResolver.clusterCommandQuery.timer";
 
+    /**
+     * How long it takes to resolve a cluster for a job given the resolved command and the request criteria.
+     */
+    private static final String RESOLVE_CLUSTER_TIMER_NAME = "genie.services.jobResolver.resolveCluster.timer";
```
mprimi: Very minor: only 2 of the constants have a `_NAME` suffix.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -672,4 +724,214 @@ private int resolveJobMemory(final JobRequest jobRequest, final Command command)
...
+
+    private Command resolveCommand(final JobRequest jobRequest) throws GenieJobResolutionException {
+        final long start = System.nanoTime();
+        final Set<Tag> tags = Sets.newHashSet();
+        try {
+            final Criterion criterion = jobRequest.getCriteria().getCommandCriterion();
+            final Set<Command> commands = this.commandPersistenceService.findCommandsMatchingCriterion(criterion);
+            final Command command;
+            if (commands.isEmpty()) {
+                throw new GenieJobResolutionException("No command matching command criterion found");
```
mprimi: Error message here and on 746 are kinda the same (i like the latter more), maybe use exactly the same for sake of error classification systems?
Or make the 2 more different/precise.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -672,4 +724,214 @@ private int resolveJobMemory(final JobRequest jobRequest, final Command command)
...
+            } else if (commands.size() == 1) {
+                command = commands
+                    .stream()
+                    .findFirst()
+                    .orElseThrow(() -> new GenieJobResolutionException("No command matching criterion found."));
+                log.debug("Found single command {} matching criterion {}", command.getId(), criterion);
+            } else {
```
mprimi: I'm wondering if we should always defer to the selector even if the matching set contains a single element.
2 reasons:
 - Selector may make "smart" decision and actually not like the only match
 - Metrics and other things would be more uniform (vs. now we short-circuit in some cases).

There is a performance cost, but it's probably negligible..

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -672,4 +724,214 @@ private int resolveJobMemory(final JobRequest jobRequest, final Command command)
...
+        final String jobId
+    ) throws GenieJobResolutionException {
+        final long start = System.nanoTime();
+        final Set<Tag> tags = Sets.newHashSet();
+        int queryCount = 0;
+        int criteriaCombinationCount = 0;
+        try {
+            Cluster cluster = null;
+
+            // This makes it so that if the command has no Cluster Criteria no cluster will possibly
```
mprimi: I think the current behavior you describe is the right one.
If a command wants to match everything and just delegate to job criteria, it can always do so by setting a cluster criteria with `status=UP`.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -672,4 +724,214 @@ private int resolveJobMemory(final JobRequest jobRequest, final Command command)
...
+     * - If only one criterion has any of these fields {@literal id, name, version, status} then that value is present
+     * in the resulting criterion
+     * - Any {@literal tags} present in either criterion are merged into the super set of both sets of tags
+     *
+     * @param one The first {@link Criterion}
+     * @param two The second {@link Criterion}
+     * @return A merged {@link Criterion} that can be used to search the database
+     * @throws IllegalArgumentException If the criteria can't be merged due to the described assumptions
+     */
+    private Criterion mergeCriteria(final Criterion one, final Criterion two) throws IllegalArgumentException {
```
mprimi: I feel like this belong somewhere else, i.e. in the criterion class, maybe as static helper/factory

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -672,4 +724,214 @@ private int resolveJobMemory(final JobRequest jobRequest, final Command command)
...
+            if (t instanceof GenieJobResolutionException) {
+                throw (GenieJobResolutionException) t;
+            } else {
+                throw new GenieJobResolutionException(t);
+            }
+        } finally {
+            this.registry
+                .counter(RESOLVE_CLUSTER_CRITERIA_COMBINATION_COUNTER, tags)
+                .increment(criteriaCombinationCount);
+            this.registry
```
mprimi: The 2 counter metrics are a little weird.
The number is in the end the number of attempts made.
However they are grouped/tagged by cluster if successful.

So in the end i cannot think of a way to extract things like average/max/min number of attempts.
It's just a global counter with some tags that i cannot think of a use for.

Let's talk in person, it's easier.

PR #956, opened 2020-02-21
Add API which allows resolving clusters from a given command criterion

tgianos:
`/api/v3/commands/{id}/clusterCiteria/{priority}/clusters` will execute the criterion resolution against the clusters in the database to see what would be selected if a job was submitted and evaluated against that criterion


mprimi:
I'm not in favor of indroducing new API that solves a thin sliver of a problem (but that we will have to maintain or worry about breaking going forward).

It seems to me that this exposes something that we can already kinda do via search APIs.

I'm in favor of introducing API if it solves the whole problem.
In my head, solving this entire problem means
- Take (job) command criteria and (job) cluster criteria
- Expose all steps and intermediate matching result

This is so that when something is not working, we can take a job, run it through this API and have a full debug trace.

tgianos:
@mprimi I believe you're discussing a different problem than this is targeted at. Your vision for helping to debug execution details is valid and useful but this specific PR is intended for helping administrators more so than end users.

The admins of our deployment asked for the ability to tell which clusters a command can run on and which commands a cluster can use during our meeting last week. I circled back with those admins and asked specifically about this PR and they re-iterated that this seems to solve what they're asking for as they work on the migration from old to new resource resolution mechanisms. Yes, this can partially be addressed with the search API but it is not abstracted in the same manner and isn't as specifically targeted to recreate the environment the clusters are resolved in as this API is.

It builds on existing APIs in the service tier developed for this brand new cluster/command resolution feature so maintenance going forward shouldn't be an issue as it is where we're heading in the near term.

tgianos:
After discussion on design I'm going to revise this a little bit and open a new PR.

PR #960, opened 2020-02-28
Add Resolved Clusters API for Commands

tgianos:
- `/api/v3/commands/{id}/resolvedClusters` will execute the criteria resolution against the clusters in the database to see which would be selected for each criterion at that point in time
- Update gRPC dependencies to match internal versions


genie-common-external/src/main/java/com/netflix/genie/common/external/dtos/v4/ResolvedResources.java
```diff
@@ -0,0 +1,67 @@
...
+    private final Criterion criterion;
+    private final ImmutableSet<R> resources;
+
+    /**
+     * Constructor.
+     *
+     * @param criterion The {@link Criterion} that was used to resolve {@literal resources}
+     * @param resources The resources that were resolved based on the {@literal criterion}
+     */
+    @JsonCreator
```
mprimi: Why is `@JsonCreator` required on this?
Just consistency with the rest of the DTOs?

tgianos: That's one reason but the main one here is that I bind these objects in the integration tests using an `ObjectMapper` for validation.

PR #964, opened 2020-03-25
Command and Application Lifecycle Support

tgianos:
Primarily adding support for cleaning unused commands and applications from the database similar to how we've done jobs, clusters, files and tags in the past.

Made all database cleanup related properties dynamic to reduce operational burden so clusters don't have to be re-deployed for changes to be picked up.

Additionally technical debt and other improvements found along the way

### DIFF
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -451,7 +451,7 @@ See: `genie.tasks.database-cleanup.expression`
 |genie.tasks.database-cleanup.expression
 |The cron expression for how often to run the database cleanup task
 |0 0 0 * * *
-|no
+|yes
 
 |genie.tasks.database-cleanup.maxDeletedPerTransaction
 |The number of job records (across multiple tables) to delete from the database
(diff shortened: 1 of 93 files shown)


genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -442,11 +442,38 @@ lost and failed by the Genie leader
...
 |Skip the Clusters table when performing database cleanup
 |false
 |yes
 
+|genie.tasks.database-cleanup.command-cleanup.skip
+|Skip the Commands table when performing database cleanup
+|false
+|yes
+
+|genie.tasks.database-cleanup.command-deactivation.commandCreationThreshold
```
mprimi: Minor: this property is named `retention` for jobs. Shouldn't they be named the same?

genie-web/src/main/java/com/netflix/genie/web/data/repositories/jpa/JpaCommandRepository.java
```diff
@@ -16,11 +16,53 @@
...
  * Command repository.
  *
  * @author tgianos
  */
 public interface JpaCommandRepository extends JpaBaseRepository<CommandEntity> {
+
+    /**
+     * The query used to set commands to a given status given input parameters.
+     */
+    String SET_UNUSED_STATUS_QUERY =
```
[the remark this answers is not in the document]
tgianos: so i'm sure you've picked up on the fact I had a hard time naming things in here as they could mean a bunch of stuff... in this case the `UNUSED` refers to the command being unused in any job moreso than the status we're setting

genie-docs/src/docs/asciidoc/_properties.adoc
tgianos: I thought about this when I was doing it. `retention` was a hard line in the sand where everything before time x was just going to be deleted so we were keeping a floating window of y days. In this case even if the command was created before this time it could be left unaffected as it's still used "recently" by a job. Since this is the deactivation setting and not the full delete setting I left it this way so I could distinguish it from the job threshold

tgianos:
> Wondering if it's worth pulling all the common code up into a superclass or reusable component.

I think this is probably worthwhile at some point. I'm just too tired of refactoring/cleaning up this old code right now to tackle it now. A lot of this cruft would be good to get rid of though I agree.

PR #965, opened 2020-03-26
Fix bug where command cluster table not taken into account for command deletion

tgianos:
Forgot to factor in the `clusters_commands` table when considering which commands to delete. This caused referential integrity failure when run on real data.
This table is going to be deleted once V4 resource selection migration is completed so it slipped my mind to include it in the original query. This will have to be ripped out again once we remove that table.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/data/repositories/jpa/JpaCommandRepository.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/data/repositories/jpa/JpaCommandRepository.java
@@ -57,7 +57,8 @@ public interface JpaCommandRepository extends JpaBaseRepository<CommandEntity> {
             + "SELECT DISTINCT(command_id)"
             + " FROM jobs"
             + " WHERE command_id IS NOT NULL"
-            + ");";
+            + ")"
+            + " AND id NOT IN (SELECT DISTINCT(command_id) FROM clusters_commands);";
 
     /**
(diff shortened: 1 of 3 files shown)


genie-web/src/main/java/com/netflix/genie/web/data/repositories/jpa/JpaCommandRepository.java
```diff
@@ -57,7 +57,8 @@
             + "SELECT DISTINCT(command_id)"
             + " FROM jobs"
             + " WHERE command_id IS NOT NULL"
-            + ");";
+            + ")"
+            + " AND id NOT IN (SELECT DISTINCT(command_id) FROM clusters_commands);";
```
mprimi: Indentation is different from similar statement a couple of lines above.

tgianos:
> So clusters must be deleted first and they will cascade remove entries from `cluster_commands`, i'm guessing?

yes, the cluster deletion is done first in `DatabaseCleanupTask`

genie-web/src/main/java/com/netflix/genie/web/data/repositories/jpa/JpaCommandRepository.java
tgianos: yeah it's cause of if you flattened it out the spaces need to be there for the SQL to be valid. I could put extra space between `NULL` and `)` but figured this was the "best" I could do

PR #967, opened 2020-03-27
Add actuator endpoint to force leadership shuffle

mprimi:
When deploying a new release of Genie, it is sometimes desirable to have one the nodes running the latest version become leader.
The old nodes may have long-running jobs and be around for a while.

The new RejoinLeaderElection actuator addresses this problem by forcing a restart of the leader election service.
In the current (zookeeper based implementation), leadership is awarded in order of arrival.
Therefore any node that receive this command will re-enqueue last, and newly added nodes will be ahead in the queue to become leader.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/spring/actuators/package-info.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/spring/actuators/package-info.java
@@ -1,0 +1,28 @@
+/*
+ *
+ *  Copyright 2020 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+/**
+ * Actuator endpoints.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@ParametersAreNonnullByDefault
+package com.netflix.genie.web.spring.actuators;
+
+import javax.annotation.ParametersAreNonnullByDefault;
--- a/genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/tasks/leader/LeaderAutoConfiguration.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/tasks/leader/LeaderAutoConfiguration.java
@@ -26,6 +26,9 @@ import com.netflix.genie.web.properties.DatabaseCleanupProperties;
 import com.netflix.genie.web.properties.LeadershipProperties;
 import com.netflix.genie.web.properties.UserMetricsProperties;
 import com.netflix.genie.web.properties.ZookeeperLeaderProperties;
+import com.netflix.genie.web.services.ClusterLeaderService;
+import com.netflix.genie.web.services.impl.ClusterLeaderServiceCuratorImpl;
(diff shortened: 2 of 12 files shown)


genie-web/src/main/java/com/netflix/genie/web/services/impl/ClusterLeaderServiceImpl.java
```diff
@@ -0,0 +1,73 @@
+/*
+ *
+ *  Copyright 2017 Netflix, Inc.
```
tgianos: year

genie-web/src/main/java/com/netflix/genie/web/services/impl/ClusterLeaderServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+package com.netflix.genie.web.services.impl;
+
+import com.netflix.genie.web.services.ClusterLeaderService;
+import org.springframework.integration.zookeeper.leader.LeaderInitiator;
+
+/**
+ * Implementation of {@link ClusterLeaderService} using Spring's {@link LeaderInitiator}.
+ *
+ * @since 4.0.0
+ * @author mprimi
```
tgianos: Nit. These two are backwards to our usual order

genie-web/src/main/java/com/netflix/genie/web/spring/actuators/RejoinLeaderElectionActuator.java
```diff
@@ -0,0 +1,76 @@
...
+ * An actuator endpoint that exposes leadership status.
+ * Allows leaving and rejoining the election. Useful when a specific set of nodes should be given priority to win the
+ * leader election (e.g., because they are running newer code).
+ *
+ * @since 4.0.0
+ * @author mprimi
+ */
+@Endpoint(id = "rejoinLeaderElection")
+@Slf4j
+public class RejoinLeaderElectionActuator {
```
tgianos: Why not just `LeaderActuator`? this makes it seem like a verb more than a noun

genie-web/src/main/java/com/netflix/genie/web/spring/actuators/package-info.java
```diff
@@ -0,0 +1,28 @@
...
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+/**
+ * Actuator endpoints.
+ *
+ * @author tgianos
```
tgianos: Look at me go! :-)

genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/tasks/leader/LeaderAutoConfiguration.java
```diff
@@ -241,4 +245,32 @@ public AgentJobCleanupTask agentJobCleanupTask(
...
+     * Create a {@link ClusterLeaderService} bean if one is not already defined and if {@link LeaderInitiator} is
+     * available.
+     *
+     * @param leaderInitiator the Spring Zookeeper/Curator based leader election component
+     * @return a {@link ClusterLeaderService}
+     */
+    @Bean
+    @ConditionalOnBean(LeaderInitiator.class)
+    @ConditionalOnMissingBean(ClusterLeaderService.class)
+    public ClusterLeaderService clusterLeaderService(final LeaderInitiator leaderInitiator) {
```
tgianos: Seems like it might be good to have one based on local leader that forces it not to be the leader?

genie-web/src/main/java/com/netflix/genie/web/spring/actuators/RejoinLeaderElectionActuator.java
mprimi: This way of creating an actuator using `@WriteOperation` is very nice, but the downside is that there's only 1 possible action associated to a `POST` request. In this case it's 'rejoin'.

If we wanted to add other actions like leave permanently, it would have to be a similar but different controller. Hence the name.

Let me see what i can do in this regards..

PR #973, opened 2020-04-10
Fix agent state machine occasional hangup

mprimi:
* Replace Spring StateMachine implementation with a simple homebrew implementation.
 * Tweak forced manifest refresh stage
 * Rename and reformat


genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
```diff
@@ -18,102 +18,266 @@
...
+        }
+    }
+
+    private void run() {
+        this.listener.stateMachineStarted();
+        for (final ExecutionStage executionStage : executionStages) {
+            final States state = executionStage.getState();
+            log.debug("Job execution stage: {} ({})", state, executionStage.getClass().getSimpleName());
+
+            this.currentState.set(state);
```
tgianos, line 97: should this be a compare and set to prevent some unexpected transition?

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
```diff
@@ -18,102 +18,266 @@
...
+            this.listener.stateEntered(state);
+
+            int retriesLeft = state.getTransitionRetries();
+            while (true) { // Not actually used, loop is broken out from inside.
```
tgianos, line 101: The contents of this loop are pretty large. Might be good to break out some of the sections into methods for readability/clarity

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -145,7 +146,7 @@ public synchronized void forceServerSync() {
         if (started.get()) {
             try {
                 this.jobDirectoryManifestCreatorService.invalidateCachedDirectoryManifest(jobDirectoryPath);
-                this.pushManifest();
+                this.taskScheduler.schedule(this::pushManifest, Instant.now());
```
tgianos, line 149: Perhaps if we want to force this it should be synchronous on the current thread instead of putting it on a scheduler?  I think even though you say `now` that's just a best effort to schedule it as soon as possible.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
```diff
@@ -18,102 +18,266 @@
...
-            .put(ExecutionStage.EXECUTION_CONTEXT_CONTEXT_KEY, executionContext);
-        this.executionCompletionListener = new StateMachineListenerAdapter<States, Events>() {
-            @Override
-            public void stateMachineStopped(final StateMachine<States, Events> sm) {
-                log.debug("State machine execution complete");
-                completionLatch.countDown();
-            }
-        };
+
+        this.taskExecutor = new TaskExecutorBuilder()
```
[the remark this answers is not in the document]
mprimi: I really wanted to be explicit on the fact that this runs on its own (single) thread.
Maybe rather than injecting it i could replace `start()` and `stop()` with a `run()` call which makes the executor superfluous.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
```diff
@@ -18,102 +18,266 @@
...
-        // Kick off first transition
-        stateMachine.sendEvent(Events.START);
+
+        final boolean startedNow = runFuture.compareAndSet(
+            null,
+            this.taskExecutor.submit(this::run)
+        );
+
+        if (!startedNow) {
+            throw new IllegalStateException("State machine already running");
```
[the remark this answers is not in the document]
mprimi: It doesn't shut it down, it just throws back to the tread that called start on an already started machine.

genie-agent/src/main/java/com/netflix/genie/agent/execution/statemachine/JobExecutionStateMachineImpl.java
mprimi: Indeed. I am in the process of writing tests, once i have those i might move things around to make it more readable while still being sure i didn't break it.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
mprimi: I understand scheduling it `now` is just a best effort. 
I don't want the calling thread to get tangled in gRPC, but i could make this synchronous by waiting on the future returned by `schedule`.

genie-agent/src/main/java/com/netflix/genie/agent/cli/ExecCommand.java
```diff
@@ -69,20 +65,15 @@ public ExitCode run() {
...
-
-        final States finalState;
+        log.info("Starting job execution");
         try {
-            finalState = stateMachine.waitForStop();
+            this.stateMachine.run();
         } catch (final Exception e) {
-            log.warn("Job state machine execution failed", e);
-            throw new RuntimeException("Job execution error", e);
+            log.error("Job state machine execution failed: {}", e.getMessage());
```
tgianos, line 72: are you purposely not adding the exception to this log message?

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionAutoConfiguration.java
```diff
@@ -92,114 +88,44 @@
      */
     @Bean
     @Lazy
-    @ConditionalOnMissingBean
-    public LoggingListener loggingListener() {
+    @ConditionalOnMissingBean(LoggingListener.class)
+    public JobExecutionListener loggingListener() {
```
tgianos, line 92: You should make these the concrete classes as the return types. You don't need to downcast them for them to be picked up in collection injections and using the concrete classes lets them be recognized easier for conditionals.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentFileStreamService.java
```diff
@@ -42,6 +44,10 @@
     /**
      * Request the service perform an update outside of the regular schedule to make sure the server has the most
      * up-to-date view of local files.
+     *
+     * @return a scheduled future, in case the caller wants to synchronously wait or verify the operation is successful
+     * or null if the task was not scheduled.
      */
-    void forceServerSync();
+    @Nullable
```
tgianos, line 51: Typically in the rest of the codebase we've generally used `Optional` for this kind of thing

genie-agent/src/main/java/com/netflix/genie/agent/cli/ExecCommand.java
mprimi: Yes, this is happening pretty much in `main()` which prints the full trace anyway right after.

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionAutoConfiguration.java
mprimi: I'll change them, but `@ConditionalOnMissingBean` (without arguments) already does the right thing: create a bean of that concrete type if it doesn't already exist.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentFileStreamService.java
mprimi: Switched to Optional, and in the calling code (`RefreshManifestStage`) i'm doing a `get` with timeout, so it waits for a bit but does not fail if it doesn't complete in time.

PR #976, opened 2020-04-21
Kill service retries and better agent connection tracking

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillServiceImpl.java
```diff
@@ -17,117 +17,165 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
- * Implementation of the {@link AgentJobKillService}.
+ * Implementation of the {@link AgentJobKillService}, listens for kill coming from server using long-polling.
  *
- * @author standon
+ * Note: this implementation still suffers from a serious flaw: because it is implemented with a unary call,
```
tgianos, line 38: Is there something stopping us from doing this now rather than putting something in we think won't really be right?

genie-web/src/main/java/com/netflix/genie/web/agent/services/AgentConnectionTrackingService.java
```diff
@@ -0,0 +1,41 @@
...
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.web.agent.services;
+
+/**
+ * Tracks active connections and heartbeats coming from agents actively executing a job.
```
tgianos, line 21: Not a big deal but missing our typical `@author` `@since` metadata

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentConnectionTrackingServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            claimedJobId,
+            jobId -> {
+                isNew.set(true);
+                return new JobStreamsRecord(jobId);
+            }
+        );
+
+        log.debug(
+            "Received heartbeat for {} job {} using stream {}",
+            isNew.get() ? "new" : "existing",
```
tgianos, line 96: Couple of minor things. Could make strings constants and `isNew.get()` is called multiple. times in this method. You could just unwrap it once after the compute if absent call

genie-web/src/test/groovy/com/netflix/genie/web/agent/services/impl/AgentConnectionTrackingServiceImplSpec.groovy
```diff
@@ -0,0 +1,222 @@
...
+import java.util.concurrent.ScheduledFuture
+import java.util.function.Supplier
+
+class AgentConnectionTrackingServiceImplSpec extends Specification {
+
+    AgentConnectionTrackingServiceImpl service
+    AgentRoutingService agentRoutingService
+    TaskScheduler taskScheduler
+    Supplier<Instant> timeSupplier
+    long streamTTL = AgentConnectionTrackingServiceImpl.STREAM_EXPIRATION_PERIOD
```
tgianos, line 34: Why not just use the constant directly? Just shorter name?

genie-web/src/test/java/com/netflix/genie/web/spring/autoconfigure/agent/services/package-info.java
```diff
@@ -0,0 +1,28 @@
+/*
+ *
+ *  Copyright 2019 Netflix, Inc.
```
tgianos, line 3: Year

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobKillServiceImpl.java
mprimi: Just doing cost/benefit, there are more urgent things to work on.
A small accumulation of parked requests that take a long time to timeout on the server hasn't created problems.
Fixing this would require a protocol change which along with migration of existing clients makes the cost of fixing this pretty high.

genie-web/src/test/groovy/com/netflix/genie/web/agent/services/impl/AgentConnectionTrackingServiceImplSpec.groovy
mprimi: Inlined, it was used only in a couple of places after all

PR #978, opened 2020-04-22
Implement regex-based filter for manifest creation

mprimi:
Allow specifying via properties which files and directories are excluded from the manifest


genie-common-internal/src/main/java/com/netflix/genie/common/internal/configs/CommonServicesAutoConfiguration.java
```diff
@@ -129,32 +134,45 @@ public JobDirectoryManifestCreatorServiceImpl jobDirectoryManifestCreatorService
         return new DirectoryManifest.Factory(directoryManifestFilter);
     }
 
+//    /**
```
tgianos, line 137: Purposely leaving this in commented out?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/properties/RegexDirectoryManifestProperties.java
```diff
@@ -0,0 +1,50 @@
...
+@Setter
+@Validated
+public class RegexDirectoryManifestProperties {
+    /**
+     * Properties prefix.
+     */
+    public static final String PROPERTY_PREFIX = "genie.jobs.files.filter";
+
+    private boolean caseSensitiveMatching = true;
+    private Collection<@NotBlank String> fileRejectPatterns = Lists.newArrayList();
```
tgianos, line 46: Any value in enforcing a set to remove duplicates? Particularly if profiles are applied i'm not sure if collections are overridden or appended

genie-common-internal/src/main/java/com/netflix/genie/common/internal/properties/package-info.java
```diff
@@ -0,0 +1,28 @@
+/*
+ *
+ *  Copyright 2019 Netflix, Inc.
```
tgianos, line 3: year. Not sure if you're copying and pasting or your intellij template isn't working right?

PR #980, opened 2020-04-29
Track agent connections using Curator's Service Discovery 

mprimi:
* Generalize Zookeeper Leader properties for other uses of Zookeeper
 * Split out Zookeeper components into a new AutoConfiguration
 * Implement AgentRoutingService on top of Curator's Service Discovery 
 * Remove AgentMetricsService

N.B.: this PR does not remove the database-based routing service (and related persistence APIs).
This implementation is retained as fallback in case Zookeeper is disabled (via Spring Integration).


genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/ZookeeperAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+    {
+        ZookeeperProperties.class
+    }
+)
+@AutoConfigureAfter(
+    {
+        org.springframework.cloud.zookeeper.ZookeeperAutoConfiguration.class
+    }
+)
+public class ZookeeperAutoConfiguration {
```
tgianos, line 52: Perhaps this whole configuration should be conditional on the presence of a `CuratorFramework` bean?

genie-web/src/test/java/com/netflix/genie/web/spring/autoconfigure/ZookeeperAutoConfigurationTest.java
```diff
@@ -0,0 +1,97 @@
...
+        new ApplicationContextRunner()
+            .withConfiguration(
+                AutoConfigurations.of(
+                    ZookeeperAutoConfiguration.class
+                )
+            );
+
+    /**
+     * Test expected beans when Zookeeper is not enabled.
+     */
```
tgianos, line 50: Just an FYI that you no longer need these stupid javadocs on the test methods if you don't want to

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentRoutingServiceCuratorDiscoveryImpl.java
```diff
@@ -0,0 +1,306 @@
...
+/**
+ * Implementation of {@link AgentRoutingService} that relies on Curator's Discovery extension.
+ * Rather than the traditional use of this recipe (register a service for the node itself, this class registers one
+ * service instance for each agent locally connected.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public class AgentRoutingServiceCuratorDiscoveryImpl implements AgentRoutingService {
```
tgianos, line 59: It might be nice eventually for either this bean directly or maybe just a leader task to implement `InfoContributor` and dump a mapping of which job id's are connected to which nodes

mprimi:
@tgianos 3 new commits
 - Hostname resolution (you already looked at it in the separate PR)
 - InfoContributor for agent jobs connected locally
 - Stop relying on agent connections stored in persistence where applicable

genie-web/src/main/java/com/netflix/genie/web/data/services/PersistenceService.java
```diff
@@ -1031,6 +1024,13 @@ void updateJobStatus(
      * @see JobStatus#getActiveStatuses()
      */
     long getActiveJobCountOnHost(@NotBlank String hostname);
+
+    /**
+     * Get the list of active agent jobs.
```
tgianos, line 1029: FYI You say `list` but return `set`

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
```diff
@@ -1904,6 +1891,16 @@ public long getActiveJobCountOnHost(@NotBlank final String hostname) {
         log.debug("[getActiveJobCountOnHost] Called for hostname {}", hostname);
         return this.jobRepository.countByAgentHostnameAndStatusIn(hostname, ACTIVE_STATUS_SET);
     }
+
```
tgianos, line 1894: I'm surprised this didn't make you add a javadoc

PR #982, opened 2020-05-06
Ignore commands from being considered if they have no cluster criteria

tgianos:
Upgrade gradle and plugins as well


build.gradle
```diff
@@ -57,7 +57,7 @@ configure(javaProjects) {
     apply plugin: "java"
     apply plugin: "checkstyle"
     apply plugin: "com.github.spotbugs"
-    apply plugin: "pmd"
+//    apply plugin: "pmd"
```
mprimi, line 60: I haven't noticed this difference.

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/queries/specifications/JpaCommandSpecs.java
```diff
@@ -114,5 +114,8 @@ private JpaCommandSpecs() {
                 CommandEntity_.id,
                 criterion
             );
+
+            return cb.and(predicate, cb.isNotEmpty(root.get(CommandEntity_.clusterCriteria)));
```
mprimi, line 118: Would this be more efficient if the order was swapped?

build.gradle
tgianos: For now I pulled this out

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/queries/specifications/JpaCommandSpecs.java
tgianos: As discussed offline this is unknown to me due to the layers the path would go through to generate the sql as well as if it's better to short circuit on a join clause vs an in row clause first. For now we'll try this and if we see a performance problem we can try switching it.

PR #984, opened 2020-05-08
Switch to log4j2 and tweak logging

mprimi:
- Switch from logback to log4j2 due to the former deadlocking on shared output streams
- Refactoring around agent logfile and console logging
- Tweak log messages


genie-agent/src/main/java/com/netflix/genie/agent/cli/UserConsole.java
```diff
@@ -27,91 +28,26 @@
...
  * @since 4.0.0
  */
+@Slf4j
 public final class UserConsole {
 
     /**
      * The name of this logger must match the one explicitly whitelisted in the underlying logger configuration
-     * consumed by Spring.
+     * consumed by Spring. It is treated differently than other logs.
+     * consumed by Spring. It is treated differently than other logs.
```
tgianos: Looks like an accidental double paste here.

genie-agent/src/main/java/com/netflix/genie/agent/cli/CliAutoConfiguration.java
```diff
@@ -18,12 +18,17 @@
...
 import com.netflix.genie.agent.AgentMetadata;
+import com.netflix.genie.agent.cli.logging.AgentLogManager;
+import com.netflix.genie.agent.cli.logging.AgentLogManagerLog4j2Impl;
 import com.netflix.genie.agent.execution.services.AgentHeartBeatService;
 import com.netflix.genie.agent.execution.services.AgentJobService;
 import com.netflix.genie.agent.execution.services.DownloadService;
 import com.netflix.genie.agent.execution.services.KillService;
 import com.netflix.genie.agent.execution.statemachine.JobExecutionStateMachine;
 import com.netflix.genie.proto.PingServiceGrpc;
+import org.apache.logging.log4j.LogManager;
```
tgianos, line 29: Since you made the bean conditional on a class shouldn't this direct import not be here as it will fail to start if someone (for some reason) ripped out log4j?

genie-agent/src/main/java/com/netflix/genie/agent/execution/ExecutionAutoConfiguration.java
```diff
@@ -221,13 +222,15 @@ CreateJobDirectoryStage createJobDirectoryStage(final JobSetupService jobSetupSe
 
     /**
      * Create a {@link RelocateLogFileStage} bean if one is not already defined.
+     *
+     * @param agentLogManager the agent log manager
      */
     @Bean
     @Lazy
     @Order(70)
```
tgianos, line 230: This is minor but you may want to make these orders public static variables that people can reference if they want to ensure their bean is higher or lower precedence.

genie-agent/src/main/java/com/netflix/genie/agent/cli/CliAutoConfiguration.java
mprimi: I'm not 100% sure, but I don't think so, it will fail to compile if log4j2 is not in the classpath, but we for example bring in plenty of spring autoconfigurations that depend on classes not in the classpath, and there's no runtime error (example: `RedisHealthContributorAutoConfiguration`).

Beside, this is a CLI application and it's not supposed to be included as a library.

I *think* (but it may be wrong) this allows a wrapper project to provide a substitute bean for a substitute logging library.

PR #986, opened 2020-05-19
Separate timeouts for starting vs. disconnected agent jobs

mprimi:
Separate the two cases of
 * Agent not yet started (job in unclaimed state)
 * Agent started then disconnected (job in active/claimed state)
and allow setting different timeout values for the two.

### DIFF
--- a/genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImplJobsIntegrationTest.java
+++ b/genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImplJobsIntegrationTest.java
@@ -1189,6 +1189,16 @@ class JpaPersistenceServiceImplJobsIntegrationTest extends JpaPersistenceService
             .containsExactlyInAnyOrder(AGENT_JOB_1, AGENT_JOB_2);
     }
 
+    @Test
+    @DatabaseSetup("persistence/jobs/unclaimed.xml")
+    void canGetUnclaimedAgentJobs() {
+
+        Assertions
+            .assertThat(this.service.getUnclaimedAgentJobs())
+            .hasSize(2)
+            .containsExactlyInAnyOrder(AGENT_JOB_1, AGENT_JOB_2);
+    }
+
     private void validateJobRequest(final com.netflix.genie.common.dto.JobRequest savedJobRequest) {
         Assertions.assertThat(savedJobRequest.getId()).isPresent().contains(UNIQUE_ID);
         Assertions.assertThat(savedJobRequest.getName()).isEqualTo(NAME);
--- a/genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
@@ -187,6 +187,16 @@ public class JpaPersistenceServiceImpl implements PersistenceService {
         .collect(Collectors.toSet());
 
     /**
+     * The a set containing statuses that come before CLAIMED.
+     */
+    @VisibleForTesting
+    static final Set<String> UNCLAIMED_STATUS_SET = JobStatus
+        .getStatusesBeforeClaimed()
+        .stream()
+        .map(Enum::name)
+        .collect(Collectors.toSet());
+
+    /**
      * The set of job statuses which are considered to be using memory on a Genie node.
      */
     @VisibleForTesting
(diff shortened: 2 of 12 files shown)


genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImplJobsIntegrationTest.java
```diff
@@ -1185,8 +1186,18 @@ void canGetAgentActiveJobs() {
...
         Assertions
             .assertThat(this.service.getActiveAgentJobs())
-            .hasSize(2)
-            .containsExactlyInAnyOrder(AGENT_JOB_1, AGENT_JOB_2);
+            .hasSize(3)
+            .containsExactlyInAnyOrder(AGENT_JOB_1, AGENT_JOB_2, AGENT_JOB_3);
+    }
+
+    @Test
+    @DatabaseSetup("persistence/jobs/search.xml")
```
tgianos, line 1194: One reason I broke it apart like this is so that individual tests can use their own setup file to prevent the trickle you're seeing here where you had to change the counts in other test methods cause you added this test and changed this file.

Instead you could copy this file into some other scoped file and limit it down to just what you want to test.

You don't have to do this obviously but it keeps the tests a little more isolated/focussed imo.

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
```diff
@@ -185,6 +185,16 @@
         .map(Enum::name)
         .collect(Collectors.toSet());
 
+    /**
+     * The a set containing statuses that come before CLAIMED.
+     */
+    @VisibleForTesting
+    static final Set<String> UNCLAIMED_STATUS_SET = Sets.newHashSet(
```
tgianos, line 192: Theres a few of these constants in here. Would it be better to locate them in the job status file?

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/AgentJobCleanupTask.java
```diff
@@ -43,7 +43,8 @@
  */
 @Slf4j
 public class AgentJobCleanupTask extends LeaderTask {
-    private static final String STATUS_MESSAGE = "Agent AWOL for too long";
+    private static final String AWOL_STATUS_MESSAGE = "Agent AWOL for too long";
+    private static final String NEVER_CLAIMED_STATUS_MESSAGE = "No agent claimed the job for too long";
```
tgianos, line 47: General question. What happens if these jobs are marked failed and the agent comes up and tries to claim them?

genie-web/src/main/java/com/netflix/genie/web/properties/AgentCleanupProperties.java
```diff
@@ -48,5 +48,7 @@
 
     private long refreshInterval = 10_000;
 
-    private long timeLimit = 120_000;
+    private long reconnectTimeLimit = 120_000;
```
tgianos, line 51: Something i've wanted to do for a while and maybe you can start doing it here is to leverage the better property conversion stuff for more readable/flexible durations. See [here](https://docs.spring.io/spring-boot/docs/2.2.7.RELEASE/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration)

They also have converters for [data sizes](https://docs.spring.io/spring-boot/docs/2.2.7.RELEASE/reference/html/spring-boot-features.html#boot-features-external-config-conversion-datasize)

genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImplJobsIntegrationTest.java
mprimi: Makes sense, didn't even occur to me but i should use a separate file.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/AgentJobCleanupTask.java
mprimi: It will fail to claim and exit with non-zero exit code.

genie-web/src/main/java/com/netflix/genie/web/properties/AgentCleanupProperties.java
mprimi: Cool! Yes i'll convert this and use these going forward.

PR #987, opened 2020-05-21
Implementation of New Algorithm for V4 Resource Resolution

tgianos:
After internal testing determined some gaps in the previous implementation of a V4 resource resolution algorithm additional design and discussion went into a new algorithm to fill those gaps. This PR contains that implementation as well as a bunch of improvements/supporting work to hopefully make the system easier to modify going forward.

Algorithm Steps (simplified a bit):

1. Take `command criterion` from user` job request` and query database for all possible matching commands (candidate commands)
2. Take `cluster criteria` from `job request` and `cluster criteria` from each `command` candidate and create uber query which finds ALL clusters that match at least one of the resulting merged criterion (merged meaning combining a job and command cluster criterion)
3. Iterate through commands from step 1 and evaluate job/command cluster criterion against resulting set of clusters from step 2. Filter out any commands that don't match any clusters. Save resulting cluster set for each command in map command -> Set<Cluster>
4. Pass `set<command>`, `jobRequest`, `jobId`, `map<command<set<Cluster>>`  to command selector which will return single command
5. Using command result pass previously computed `Set<Cluster>` to cluster selector to select single cluster
6. Save results and run job

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
@@ -87,26 +87,32 @@ import java.util.stream.Collectors;
 @Slf4j
 @Validated
 public class JobResolverServiceImpl implements JobResolverService {
+    //region Metric Constants
     /**
      * How long it takes to completely resolve a job given inputs.
      */
     private static final String RESOLVE_JOB_TIMER = "genie.services.jobResolver.resolve.timer";
 
     /**
-     * How long it takes to query the database for cluster command combinations matching supplied criteria.
+     * How long it takes to resolve a command for a job given the supplied command criterion.
      */
-    private static final String CLUSTER_COMMAND_QUERY_TIMER = "genie.services.jobResolver.clusterCommandQuery.timer";
+    private static final String RESOLVE_COMMAND_TIMER = "genie.services.jobResolver.resolveCommand.timer";
+
+    /**
+     * How long it takes to resolve a cluster for a job given the resolved command and the request criteria.
+     */
+    private static final String RESOLVE_CLUSTER_TIMER = "genie.services.jobResolver.resolveCluster.timer";
+
+    /**
+     * How long it takes to resolve the applications for a given command.
+     */
+    private static final String RESOLVE_APPLICATIONS_TIMER = "genie.services.jobResolver.resolveApplications.timer";
 
     /**
      * How long it takes to resolve a cluster for a job given the resolved command and the request criteria.
      */
     private static final String GENERATE_CRITERIA_PERMUTATIONS_TIMER
         = "genie.services.jobResolver.generateClusterCriteriaPermutations.timer";
-
-    /**
-     * How long it takes to resolve a cluster for a job given the resolved command and the request criteria.
-     */
-    private static final String RESOLVE_CLUSTER_TIMER = "genie.services.jobResolver.resolveCluster.timer";
 
     /**
      * The number of criteria combinations attempted while resolving a cluster.
@@ -120,11 +126,6 @@ public class JobResolverServiceImpl implements JobResolverService {
     private static final String RESOLVE_CLUSTER_QUERY_COUNTER = "genie.services.jobResolver.resolveCluster.query.count";
 
     /**
-     * How long it takes to resolve a command for a job given the supplied command criterion.
-     */
-    private static final String RESOLVE_COMMAND_TIMER = "genie.services.jobResolver.resolveCommand.timer";
-
-    /**
      * How long it takes to select a cluster from the set of clusters returned by database query.
      */
     private static final String SELECT_CLUSTER_TIMER = "genie.services.jobResolver.selectCluster.timer";
@@ -133,11 +134,6 @@ public class JobResolverServiceImpl implements JobResolverService {
      * How long it takes to select a command for a given cluster.
      */
     private static final String SELECT_COMMAND_TIMER = "genie.services.jobResolver.selectCommand.timer";
-
-    /**
-     * How long it takes to select the applications for a given command.
-     */
-    private static final String SELECT_APPLICATIONS_TIMER = "genie.services.jobResolver.selectApplications.timer";
 
     /**
      * How many times a cluster selector is invoked.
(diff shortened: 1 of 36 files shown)


genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -915,6 +925,62 @@ private Cluster resolveCluster(
...
+                final ImmutableList.Builder<Criterion> listBuilder = ImmutableList.builder();
+                for (final Criterion commandClusterCriterion : command.getClusterCriteria()) {
+                    for (final Criterion jobClusterCriterion : jobRequest.getCriteria().getClusterCriteria()) {
+                        try {
+                            // Failing to merge the criteria is equivalent to a round-trip DB query that returns
+                            // zero results. This is an in memory optimization which also solves the need to implement
+                            // the db query as a join with a subquery.
+                            listBuilder.add(this.mergeCriteria(commandClusterCriterion, jobClusterCriterion));
+                        } catch (final IllegalArgumentException e) {
+                            log.debug(
```
mprimi, line 816: This doesn't deserve at least a `warn`?

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -968,4 +1034,63 @@ private String mergeCriteriaStrings(
...
+                        matchedClustersBuilder.add(candidateCluster);
+                    }
+                }
+
+                final ImmutableSet<Cluster> matchedClusters = matchedClustersBuilder.build();
+                if (!matchedClusters.isEmpty()) {
+                    // If we found some clusters the evaluation for this command is done
+                    matrixBuilder.put(command, matchedClusters);
+                    log.debug("For command {} matched clusters {}", command, matchedClusters);
+                    // short circuit further criteria evaluation for this command
```
mprimi: Seems like this short-circuiting could create a matrix where only a subset of the potential matches for a commands are present. Is this ok?

Wouldn't the lack of some clusters (that actually ARE a match, but not present in this set) influence the command selector?

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -696,30 +668,48 @@ private String tagsToString(final Set<String> tags) {
...
+    private void resolveJobMemory(final JobResolutionContext context) {
+        context.setJobMemory(
+            context.getJobRequest()
+                .getRequestedJobEnvironment()
+                .getRequestedJobMemory()
+                .orElse(
+                    context
+                        .getCommand()
+                        .orElseThrow(
+                            () -> new IllegalStateException(
```
mprimi: Seems to me like this and similar unchecked exception will bomb resolution, rather than throwing `GenieJobResolutionException`

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -379,95 +386,167 @@ private ResolvedJob resolve(final JobResolutionContext context) throws GenieJobR
...
-            MetricsUtils.addFailureTagsWithException(tags, t);
-            throw new GenieJobResolutionException(t);
-        } finally {
+                .counter(RESOLVE_CLUSTER_CRITERIA_COMBINATION_COUNTER)
+                .increment(criteriaCombinationCount);
             this.registry
-                .timer(SELECT_COMMAND_TIMER, tags)
+                .counter(RESOLVE_CLUSTER_QUERY_COUNTER)
+                .increment(queryCount);
+            tags.add(Tag.of(RESOLVE_CLUSTER_CRITERIA_COMBINATION_TAG, String.valueOf(criteriaCombinationCount)));
```
mprimi: I think the proper way to make this info available (and easy to query) is a distribution/histogram: https://micrometer.io/docs/concepts#_distribution_summaries

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
tgianos: The top level public API implementations both have the following catch blocks:
```
        } catch (final GenieJobResolutionException e) {
            MetricsUtils.addFailureTagsWithException(tags, e);
            throw e;
        } catch (final Throwable t) {
            MetricsUtils.addFailureTagsWithException(tags, t);
            throw new GenieJobResolutionException(t);
        }
```
 ensuring the checked contract is held up outside the class implementation.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
tgianos: I don't think so in this case. The constructor for `Criterion` does indeed throw an exception for invalid configuration but at this point both the input criterion where valid so there's no real case where they can be invalid due to invalid data. The `IllegalArgumentException` is thrown by the helper method due to a case where the comment above the `add` line states because the query to the db can never succeed (e.g. conflicting `id` fields in the criterion can never succeed in matching anything)

It was a way to just signal back to this caller that the resulting criterion would be invalid. We could instead pass back `Optional` with empty representing that the two can't be merged if that makes the intent clearer.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
tgianos: Cool. I'll take a look at this for future stuff. In this case this logic is deleted later in the PR anyway as we don't have these nested queries.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
tgianos: This is a legit question. Basically you're asking if the matrix should instead be `Command -> List<Set<cluster>>` where we fully explore the combinations. As of now this logic matches what v3 did in terms of short circuiting the evaluation at the highest priority combination found to have matches. The selectors for clusters are guaranteed to select at least one if the random selector remains the last fallback.

The question here is if the command selector will make different decision based on presence of more clusters at a lower priority. Perhaps. @rdblue or Dan (couldn't find his git handle) may have an opinion however the only way I see this working is if the `CommandSelector` and `ClusterSelector` are combined as there (currently at least) is no way for the command selector to pass back information like "ok use this command but use the 3rd highest priority set of clusters not the first". We'd wnat it to select the command and the cluster.

For now I think this is something we could modify in the future if desired unless Ryan or Dan believe it critical to continuing with their work.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
rdblue: While I think it is safe to do this, it would probably be a better practice to log this as a warning like @mprimi suggests. That way if this is ever happening for an unexpected reason, we can clearly see what conflicted and whether that was correct or the result of a bug. I'd opt for visibility, though possibly at an info level rather than warn.

genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
rdblue: I think this is correct, and a good way to avoid computing the entire intersection. We only need the highest priority match, just like before.

PR #988, opened 2020-05-22
Archive files if present in job manifest

mprimi:


genie-common-internal/src/main/java/com/netflix/genie/common/internal/services/JobArchiver.java
```diff
@@ -37,10 +39,11 @@
...
      *
      * @param directory The directory to archive
+     * @param filesList The list of files to archive
      * @param target    The root of a writable location to archive to.
      * @return {@code false} if this implementation doesn't support archiving to {@code target}. {@code true} if does
      * support archiving to {@code target} and the archival was successful
      * @throws JobArchiveException If an exception happened during archival
      */
-    boolean archiveDirectory(Path directory, URI target) throws JobArchiveException;
+    boolean archiveDirectory(Path directory, List<File> filesList, URI target) throws JobArchiveException;
```
tgianos, line 48: Nit: I think this variable name should be just `files`

tgianos, line 48: Could we just pass the manifest? Or are you trying to keep it decoupled?

mprimi: If we pass the manifest, all implementations would need to do some duplicate work of going through it. This approach with the file list seems more generic. The archivers just upload what they're told to archive explicitly.

PR #992, opened 2020-05-28
Fill in missing gaps in metric reporting for agent api jobs

tgianos:
Previously the jobs launched with the agent weren't being considered in various metrics

This PR also improves the queries used for some metric generation and hopefully the performance of the system by caching aggregate data for health and metric reporting and periodically updating it


genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/health/HealthAutoConfiguration.java
```diff
@@ -49,22 +46,6 @@
...
-    /**
-     * Provide a health indicator related to job memory usage if one hasn't already been provided elsewhere.
-     *
-     * @param jobMetricsService Implementation of {@link JobMetricsService} to use
-     * @param jobsProperties    The {@link JobsProperties} to use
-     * @return An instance of {@link GenieMemoryHealthIndicator}
-     */
-    @Bean
-    @ConditionalOnMissingBean(GenieMemoryHealthIndicator.class)
-    public GenieMemoryHealthIndicator genieMemoryHealthIndicator(
```
mprimi, line 61: This indicator is listed in the property `genie.tasks.cluster-checker.healthIndicatorsToIgnore` value.
Doesn't hurt to leave it, but should be removed.

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/AgentLauncher.java
```diff
@@ -19,14 +19,15 @@
...
+import org.springframework.boot.actuate.health.HealthIndicator;
 
 /**
  * A interface which implementations will launch instances of an agent in some manner in order to run a job.
  *
  * @author tgianos
  * @since 4.0.0
  */
-public interface AgentLauncher {
+public interface AgentLauncher extends HealthIndicator {
```
mprimi, line 30: I believe you need to update `genie.tasks.cluster-checker.healthIndicatorsToIgnore` with the new name that this indicator will get.
Otherwise the leader is gonna mark jobs on an host failed when that host is out of memory.

genie-web/src/main/java/com/netflix/genie/web/properties/LocalAgentLauncherProperties.java
```diff
@@ -109,4 +112,20 @@
...
+    @DurationUnit(ChronoUnit.SECONDS)
+    private Duration hostInfoExpireAfter = Duration.ofMinutes(1L);
+
+    /**
+     * How long after the job information for this host is written should it be automatically refreshed from
+     * the underlying data source.
+     *
+     * Should be lower than {@link #getHostInfoExpireAfter()}
+     */
+    private Duration hostInfoRefreshAfter = Duration.ofSeconds(30L);
```
mprimi, line 130: The one above is decorated with `@DurationUnit`, but this one is not

genie-web/src/main/java/com/netflix/genie/web/data/services/PersistenceService.java
```diff
@@ -107,12 +107,12 @@
      * @return The page of found applications
      */
     Page<Application> findApplications(
-        @Nullable String name,
-        @Nullable String user,
-        @Nullable Set<ApplicationStatus> statuses,
-        @Nullable Set<String> tags,
-        @Nullable String type,
-        Pageable page
+            @Nullable String name,
```
mprimi, line 110: Hmm.. all of these got 8 spaces of indentation. Is that right..?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/AgentLauncher.java
tgianos: I double checked by running the OSS one and comparing it to our current deployment. Both get the key `localAgentLauncher` so I think we're ok here.

genie-web/src/main/java/com/netflix/genie/web/properties/LocalAgentLauncherProperties.java
tgianos: Yeah. The `@DurationUnit` is only to tell spring what to consider a value without a qualifier like `s` or `m`. Anything not annotated the default is milliseconds. So in this case I just put it to `SECONDS` on the expiry one as another catch condition for people not to accidentally make expiry < refresh

genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/health/HealthAutoConfiguration.java
tgianos: Done

genie-web/src/main/java/com/netflix/genie/web/data/services/PersistenceService.java
tgianos: Good catch. Not sure why this changed all of a sudden but the IDE `continuation indent` setting was 8 spaces.

Fixed

genie-web/src/main/java/com/netflix/genie/web/properties/LocalAgentLauncherProperties.java
mprimi: I think having these 2 right next to eachother with 2 different units is a little confusing.
Personally I'd leave everything in milliseconds which is "convention" in this project.

One larger than the other, means 100% cache miss, which is bad but not terrible.

A worst mistake that could happen here is:
`host-info-expire-after: 3000`

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/AgentLauncher.java
mprimi: Weird... that's not the bean name, nor the class name, nor is a constant i could find anywhere. Kind of an incredible coincidence that they get the same name 🤷‍♂️

genie-web/src/main/java/com/netflix/genie/web/properties/LocalAgentLauncherProperties.java
tgianos: Fair enough I'll remove the first one before merging this.

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/AgentLauncher.java
tgianos: Probably some convention where they rip off `Impl` or other `HealthIndicator` to get to the core class name. I tried looking for something in the API which would allow us to dictate the name but I couldn't find anything which was surprising.

PR #993, opened 2020-05-28
[PREVIEW] Replace hardcoded agent values with properties

mprimi:
I'm replacing hardcoded values in the agent with Properties classes.
Eventually the values will come from the server and get bound.


genie-common-internal/src/main/java/com/netflix/genie/common/internal/properties/ExponentialBackOffTriggerProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ * since it's not a root property class.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+@NoArgsConstructor
+public class ExponentialBackOffTriggerProperties {
```
tgianos, line 41: You could still tag it unless you're just concerned about a proliferation of objects in the JVM

PR #996, opened 2020-06-02
Agent Runtime Config

mprimi:
Allow the server to override some agent runtime configuration.
For example, slow down the rate of heartbeat if the server is under excessive load.
But could also be used to remotely turn off agent features or perform A/B experiments.

 * Remove hardcoded configuration in Agent with Properties classes
 * Create protocol to retrieve properties overrides from the server
 * Introduce new execution stage that binds server-provided values into loaded ones

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/ConfigureException.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/ConfigureException.java
@@ -1,0 +1,46 @@
+/*
+ *
+ *  Copyright 2020 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.agent.execution.exceptions;
+
+/**
+ * Failure to obtain configuration properties from server.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public class ConfigureException extends Exception {
+
+    /**
+     * Constructor with message.
+     *
+     * @param message a message
+     */
+    public ConfigureException(final String message) {
+        super(message);
+    }
+
+    /**
+     * Constructor with message and cause.
+     *
+     * @param message a message
+     * @param cause   a cause
+     */
+    public ConfigureException(final String message, final Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobService.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobService.java
@@ -18,6 +18,7 @@
 package com.netflix.genie.agent.execution.services;
 
 import com.netflix.genie.agent.execution.exceptions.ChangeJobStatusException;
+import com.netflix.genie.agent.execution.exceptions.ConfigureException;
 import com.netflix.genie.agent.execution.exceptions.HandshakeException;
 import com.netflix.genie.agent.execution.exceptions.JobIdUnavailableException;
 import com.netflix.genie.agent.execution.exceptions.JobReservationException;
@@ -30,6 +31,7 @@ import org.springframework.validation.annotation.Validated;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotBlank;
+import java.util.Map;
 
 /**
  * Agent side job specification service for resolving and retrieving job specifications from the server.
@@ -50,6 +52,17 @@ public interface AgentJobService {
     void handshake(
         @Valid AgentClientMetadata agentClientMetadata
     ) throws HandshakeException;
+
+    /**
+     * Obtain server-provided configuration properties.
+     *
+     * @param agentClientMetadata metadata about the client making this request
+     * @throws ConfigureException if the server properties cannot be obtained
+     * @return a map of properties
+     */
+    Map<String, String> configure(
+        @Valid AgentClientMetadata agentClientMetadata
+    ) throws ConfigureException;
 
     /**
      * Request a given job id to be reserved for this job, send along the job details, to be persisted by the server.
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
@@ -17,8 +17,10 @@
  */
 package com.netflix.genie.agent.execution.services.impl.grpc;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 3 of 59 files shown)


genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/ConfigureException.java
```diff
@@ -0,0 +1,46 @@
+/*
+ *
+ *  Copyright 2019 Netflix, Inc.
```
tgianos, line 3: Year

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobService.java
```diff
@@ -51,6 +53,17 @@ void handshake(
...
 
+    /**
+     * Obtain server-provided configuration properties.
+     *
+     * @param agentClientMetadata metadata about the client making this request
+     * @throws ConfigureException if the server properties cannot be obtained
+     * @return a map of properties
+     */
+    Map<String, String> configure(
+        @Valid AgentClientMetadata agentClientMetadata
```
tgianos, line 64: Is this sending the same payload as the `handshake`? I assume it's separate as the handshake can happen multiple times with different servers but this should only happen once per job?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
```diff
@@ -117,6 +122,26 @@ public void handshake(
         }
     }
 
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<String, String> configure(
```
tgianos, line 130: You might want to think about preemptively making this return type some sort of `AgentConfig` object or something so that you can add other things to it if you need to other than just key value pairs?

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentJobServiceImpl.java
```diff
@@ -124,6 +126,17 @@ public void handshake(
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<String, String> getAgentProperties(
+        @Valid final AgentClientMetadata agentClientMetadata
+    ) {
+        // TODO: currently NOOP. Implement.
```
tgianos: I'm going in order of commits so I'll assume this is completed in another commit?

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentConfigurationServiceImpl.java
```diff
@@ -0,0 +1,138 @@
...
+        );
+
+        // Use a cache to re-compute agent properties periodically, rather than for every request.
+        this.cache = Caffeine
+            .newBuilder()
+            .expireAfterWrite(this.agentConfigurationProperties.getCacheExpirationInterval())
+            .refreshAfterWrite(this.agentConfigurationProperties.getCacheRefreshInterval())
+            .initialCapacity(1)
+            .build(this::loadProperties);
+    }
```
tgianos, line 83: Ha this looks familiar! We might want to refactor it out into something reusable?

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentConfigurationServiceImpl.java
```diff
@@ -0,0 +1,138 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<String, String> getAgentProperties() {
+        return this.cache.get(AGENT_PROPERTIES_CACHE_KEY);
+    }
+
+
+    private Map<String, String> loadProperties(@NonNull final String propertiesKey) {
+        if (AGENT_PROPERTIES_CACHE_KEY.equals(propertiesKey)) {
```
tgianos, line 95: Why so many levels of indirection here between the getter above and the load function in the cache and the reload properties below?

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentJobServiceImpl.java
```diff
@@ -133,8 +138,21 @@ public void handshake(
     public Map<String, String> getAgentProperties(
         @Valid final AgentClientMetadata agentClientMetadata
     ) {
-        // TODO: currently NOOP. Implement.
-        return Maps.newHashMap();
+        final HashSet<Tag> tags = Sets.newHashSet(
+            Tag.of(AGENT_VERSION_METRIC_TAG_NAME, agentClientMetadata.getVersion().orElse("null")),
+            Tag.of(AGENT_HOST_METRIC_TAG_NAME, agentClientMetadata.getHostname().orElse("null"))
```
tgianos, line 143: I think this is a bad idea for cardinality. It makes sense in theory but in reality we're going to get a lot of UUID's for host names on containers

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentConfigurationServiceImpl.java
```diff
@@ -0,0 +1,138 @@
...
+        final PropertySources propertySources;
+        if (environment instanceof ConfigurableEnvironment) {
+            propertySources = ((ConfigurableEnvironment) environment).getPropertySources();
+        } else {
+            propertySources = new StandardEnvironment().getPropertySources();
+        }
+
+        // Create set of all properties that match the filter pattern
+        final Set<String> filteredPropertyNames = Sets.newHashSet();
+        for (final PropertySource<?> propertySource : propertySources) {
```
tgianos, line 112: Is this the only way to flatten and list all the properties? I'd have hoped somewhere in the environment there was a better API

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/AgentJobService.java
mprimi: Yes it's the same payload.
Basically allowing the server to send down different properties based on agent version, if that ever becomes necessary.

(Handshake is also executed once per job, not once per connection)

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentJobServiceImpl.java
mprimi: Indeed, that is the proper thing to do here.
I cut a corner since this is relatively easy to change later (server-side only), and didn't want to deal with DTO, converters, etc to just get a map of strings..

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentConfigurationServiceImpl.java
mprimi: I suppose i could inline the function below here.

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentJobServiceImpl.java
mprimi: Indeed, i added this for consistency with the rest of this class, but i meant to then add a commit that removes all hostname tags).
I'll add it in this same PR.

genie-web/src/main/java/com/netflix/genie/web/agent/services/impl/AgentConfigurationServiceImpl.java
mprimi: I looked at Spring's `EnvironmentEndpoint` to see how they "list all properties" and this is what they do (in their case it's a lot more complex because they want to keep them separate and annotated with origin, etc). But same basic approach.

mprimi:
@tgianos removed the `hostname` tagging with an extra commit at the end. Fixed a couple of other things you pointed out.

PR #999, opened 2020-06-05
Implement fast and efficient range requests when serving agent files

mprimi:
Improve the implementation of range requests when the file is served by a live agent by only streaming the requested bytes, as opposed to sending the entire file then letting the server seek and discard.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/agent/resources/AgentFileProtocolResolver.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/agent/resources/AgentFileProtocolResolver.java
@@ -24,12 +24,15 @@ import org.apache.http.client.utils.URIBuilder;
 import org.springframework.core.io.ProtocolResolver;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
+import org.springframework.http.HttpRange;
 
+import javax.annotation.Nullable;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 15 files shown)


genie-web/src/main/java/com/netflix/genie/web/agent/resources/AgentFileProtocolResolver.java
```diff
@@ -107,12 +121,44 @@ public Resource resolve(final String location, final ResourceLoader resourceLoad
...
-            agentFileStreamService.getResource(jobId, relativePath, uri).orElse(null);
+            agentFileStreamService.getResource(jobId, relativePath, uri, rangeStart, rangeEnd).orElse(null);
 
         if (resourceOrNull != null) {
             log.debug("Returning resource: {}", resourceOrNull.getDescription());
         }
         return resourceOrNull;
     }
+
+    private Pair<Integer, Integer> parseRangeHeader(final String rangeHeader) {
```
tgianos, line 137: I thought (in looking at some of this in the past) that Spring had a bunch of utilities for working with headers and handling range requests. I know there is a [HttpRange](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpRange.html) abstraction that exists already. Not sure if it would help to get rid of all this error checking and whatnot but thought I'd point it out to save you from possibly reinventing the wheel.

mprimi: Yes! This does pretty much everything in the PR i just wrote. I got busy rewriting the missing commit and forgot to do some research to see if something was available off the shelf..
I'll replace with this, should be pretty quick to do so.

PR #1002, opened 2020-06-13
Allow retrieval of arbitrary files via Java client

vickyuec:


genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -482,7 +482,33 @@ public InputStream getJobStderr(
...
+    /**
+     * Method to fetch an output file for a job from Genie.
+     *
+     * @param jobId The id of the job whose output file is desired.
+     * @param outputFilePath The path to the file within output directory.
+     * @return An input stream to the output file contents.
+     * @throws GenieClientException If the response received is not 2xx.
+     * @throws IOException          For Network and other IO issues.
+     */
+    public InputStream getJobOutputFile(
```
mprimi, line 499: Should `getJobStdErr` and `getJobStdOut` be re-implemented using this, rather than duplicate?

tgianos:
@vickyuec As @mprimi said thanks for taking the time to add this. Aside from the things Marco already mentioned I have a higher level concern about the behavior of this API. One of the reasons we originally (way back when) had the two defined APIs was because the backing REST endpoint can serve as something of a file browser.

By this I mean a request to `/api/v3/{jobId}/output` will return a manifest of that directory (html or JSON depending on accept headers) whereas a request to `/api/v3/{jobId}/output/stdout` will give back a file. It was designed this way to support the UI file browser originally.

Within this PR I see you're just blindly returning the stream of bytes which in cases of directory requests is odd and I'm not sure if it's the best possible solution.

vickyuec:
@mprimi I've done the refactoring and updated the IT.

Also had discussion with @tgianos offline and he suggested that the best we can do right now is to warn the client about directory vs file behavior. So I've updated the method javadoc. Tom also mentioned that v4 API will make this distinction better for clients.

genie-client/src/main/java/com/netflix/genie/client/apis/JobService.java
```diff
@@ -144,6 +144,24 @@
     @GET(JOBS_URL_SUFFIX + "/{id}/output/stderr")
     Call<ResponseBody> getJobStderr(@Path("id") String jobId);
 
+    /**
```
mprimi, line 147: The two methods above this one should now be removed, or is there some reason to keep them?

vickyuec: They can be removed now. Fixed in following comming.

vickyuec:
> Please get rid of the merge commit before merging this.

Done.

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -476,13 +469,37 @@ public InputStream getJobStdout(final String jobId) throws IOException, GenieCli
...
+        final String jobId, final String outputFilePath
     ) throws IOException, GenieClientException {
         if (StringUtils.isEmpty(jobId)) {
             throw new IllegalArgumentException("Missing required parameter: jobId.");
         }
-        final ResponseBody body = this.jobService.getJobStderr(jobId).execute().body();
+        if (StringUtils.isEmpty(outputFilePath)) {
```
tgianos, line 496: @vickyuec I think this can be empty as it would return the root of the output directory. At least if you want to be complete.

vickyuec: Fixed this, and updated IT to test it.

PR #1007, opened 2020-06-20
Improvements to agent file streaming

mprimi:
Refactoring of FileStream service that addresses concurrency issues and bugs surfaced during real-world testing.

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
@@ -117,6 +117,7 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
         this.jobId = claimedJobId;
         this.jobDirectoryPath = jobDirectoryRoot;
         this.scheduledTask = this.taskScheduler.schedule(this::pushManifest, trigger);
+        log.debug("Started");
     }
 
     /**
@@ -124,6 +125,7 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
      */
     @Override
     public synchronized void stop() {
+        log.debug("Stopping");
 
         if (this.started.compareAndSet(true, false)) {
             if (!this.activeFileTransfers.isEmpty()) {
@@ -133,6 +135,7 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
             this.scheduledTask = null;
             this.discardCurrentStream(true);
             while (!this.activeFileTransfers.isEmpty()) {
+                log.debug("{} transfers are still active after waiting", this.activeFileTransfers.size());
                 try {
                     final FileTransfer fileTransfer = this.activeFileTransfers.iterator().next();
                     if (this.activeFileTransfers.remove(fileTransfer)) {
@@ -143,6 +146,7 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
                 }
             }
         }
+        log.debug("Stopped");
     }
 
     /**
@@ -152,6 +156,7 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
     public synchronized Optional<ScheduledFuture<?>> forceServerSync() {
         if (started.get()) {
             try {
+                log.debug("Forcing a manifest refresh");
                 this.jobDirectoryManifestCreatorService.invalidateCachedDirectoryManifest(jobDirectoryPath);
                 return Optional.of(this.taskScheduler.schedule(this::pushManifest, Instant.now()));
             } catch (Exception e) {
@@ -210,6 +215,7 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
             }
 
             if (this.controlStreamObserver == null) {
+                log.debug("Creating new control stream");
                 this.controlStreamObserver = fileStreamServiceStub.sync(this.responseObserver);
                 if (this.controlStreamObserver instanceof ClientCallStreamObserver) {
                     ((ClientCallStreamObserver) this.controlStreamObserver)
@@ -217,24 +223,27 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
                 }
             }
 
+            log.debug("Sending manifest via control stream");
             this.controlStreamObserver.onNext(jobFileManifest);
         }
     }
 
     private void handleControlStreamError(final Throwable t) {
-        log.warn("Manifest stream error: {}", t.getMessage(), t);
+        log.warn("Control stream error: {}", t.getMessage(), t);
         this.trigger.reset();
         this.discardCurrentStream(false);
     }
 
     private void handleControlStreamCompletion() {
-        log.debug("Manifest stream completed");
+        log.debug("Control stream completed");
         this.discardCurrentStream(false);
     }
 
     private synchronized void discardCurrentStream(final boolean sendStreamCompletion) {
         if (this.controlStreamObserver != null) {
+            log.debug("Discarding current control stream");
             if (sendStreamCompletion) {
+                log.debug("Sending control stream completion");
                 this.controlStreamObserver.onCompleted();
             }
             this.controlStreamObserver = null;
@@ -284,11 +293,13 @@ public class GRpcAgentFileStreamServiceImpl implements AgentFileStreamService {
         );
         this.activeFileTransfers.add(fileTransfer);
         fileTransfer.start();
+        log.debug("Created and started new file transfer: {}", fileTransfer.streamId);
     }
 
     private void handleTransferComplete(final FileTransfer fileTransfer) {
         this.activeFileTransfers.remove(fileTransfer);
         this.concurrentTransfersSemaphore.release();
+        log.debug("File transfer completed: {}", fileTransfer.streamId);
     }
 
     private static class ServerControlStreamObserver implements StreamObserver<ServerControlMessage> {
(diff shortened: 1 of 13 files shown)


genie-web/src/main/java/com/netflix/genie/web/util/StreamBuffer.java
```diff
@@ -90,28 +90,44 @@ public void closeForCompleted() {
...
+     * Try to append a chunk of data for consumption.
+     * If the previous buffer is still not drained, then does not block and returns false.
+     *
+     * @param data the data to write into the buffer
+     * @return true if the data was added to the buffer, false otherwise
+     * @throws IllegalStateException if writing is attempted after the buffer has been closed
+     */
+    public boolean tryWrite(final ByteString data) {
+        synchronized (this.lock) {
+            if (this.closed) {
```
tgianos, line 114: Should `this.closed` be atomic? It looks unprotected at the moment

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -379,30 +410,34 @@ private void completeTransfer(final boolean shutdownStream, @Nullable final Exce
         private void sendChunk() throws IOException {
 
             if (this.watermark < this.endOffset - 1) {
```
tgianos, line 412: Does this class need to be thread safe? If so should these variables be atomic?

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -142,333 +151,630 @@ public GRpcAgentFileStreamServiceImpl(
...
-    private void unregisterControlStream(final String jobId, final ControlStreamObserver controlStreamObserver) {
-        final boolean removed = this.jobIdControlStreamMap.remove(jobId, controlStreamObserver);
-        if (!removed) {
-            log.warn("Could not remove observer for job: {}, not found in map", jobId);
-        }
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void contribute(final Info.Builder builder) {
```
tgianos: 👍

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -142,333 +151,630 @@ public GRpcAgentFileStreamServiceImpl(
...
+            final String jobId,
+            final String fileTransferId,
+            final String relativePath,
+            final int startOffset,
+            final int endOffset
+        ) throws GenieNotFoundException {
+
+            final ControlStreamObserver controlStreamObserver = this.controlStreamMap.get(jobId);
+            if (controlStreamObserver == null) {
+                throw new GenieNotFoundException("No active stream control stream for job: " + jobId);
```
tgianos, line 238: On the server side I've been trying to use the newer exceptions. In this case there's [NotFoundException](https://github.com/Netflix/genie/blob/master/genie-web/src/main/java/com/netflix/genie/web/exceptions/checked/NotFoundException.java).

Reason being eventually we need to get rid of the old `genie-common` and those exceptions are outdated anyway as they tie the HTTP error code to the exception rather than delegating to outside mapper which has more context.

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -142,333 +151,630 @@ public GRpcAgentFileStreamServiceImpl(
...
-    private void handleFileTransferError(
-        final FileTransferStreamObserver fileTransferStreamObserver,
-        @Nullable final String streamId,
-        final Throwable t
-    ) {
-        log.error("Error in file transfer stream: {}: {}", streamId, t.getMessage(), t);
+        private StreamObserver<AgentManifestMessage> handleNewControlStream(
+            final StreamObserver<ServerControlMessage> responseObserver
+        ) {
+            log.info("New agent control stream established");
```
tgianos, line 261: Is there a way in the information that is available to log which job this stream is for? Could be helpful for debugging.

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -142,333 +151,630 @@ public GRpcAgentFileStreamServiceImpl(
...
+            }
         }
 
-        final StreamBuffer inProgressTransferBuffer = this.inProgressTransferBuffersMap.remove(streamId);
-        if (inProgressTransferBuffer != null) {
-            inProgressTransferBuffer.closeForCompleted();
+        @Override
+        public synchronized void contribute(final Info.Builder builder) {
+            builder
+                .withDetail("file_control_streams", this.controlStreamMap.size())
```
tgianos: Keys should probably be constants. Also since this renders as JSON in browser I think it's preferred to used `-` as delimiter instead of `_`

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -142,333 +151,630 @@ public GRpcAgentFileStreamServiceImpl(
...
+                    this.activeTransfers.values()
+                        .stream()
+                        .map(FileTransfer::toString)
+                        .collect(Collectors.toList())
+                )
+                .withDetail("unclaimed_transfer_streams", this.unclaimedTransferStreams.size());
+        }
+    }
+
+    private static final class FileTransfer {
```
tgianos, line 636: There's a bunch of private classes in this file. Might help readability to break them out into perhaps package protected independent classes?

genie-web/src/main/java/com/netflix/genie/web/properties/AgentFileStreamProperties.java
```diff
@@ -0,0 +1,74 @@
...
+@ConfigurationProperties(prefix = AgentFileStreamProperties.PROPERTY_PREFIX)
+@Getter
+@Setter
+@Validated
+public class AgentFileStreamProperties {
+
+    /**
+     * Prefix for all properties related to the agent file transfer service.
+     */
+    public static final String PROPERTY_PREFIX = "genie.agent.filestream";
```
tgianos, line 42: Nit: should it be `file-stream`?

genie-web/src/main/java/com/netflix/genie/web/util/StreamBuffer.java
mprimi: Members that need synchronization are protected by `this.lock`. Maybe i'm misunderstanding your comment?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImpl.java
mprimi: The parent class is synchronized where appropriate.
This inner `FileTransfer` class should only be used by one (gRPC) thread at the time so adding further synchronization should not be necessary.

genie-web/src/main/java/com/netflix/genie/web/util/StreamBuffer.java
tgianos: I didn't closely read the whole file but is every possible mutation of `closed` protected by the same lock?

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
mprimi: I had to pull this out in the next commit (leaving it in a separate commit because i hope to restore it at some point).

Something with having this here creates a circular dependency in Spring initialization and the app doesn't start.

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
mprimi: I'll update

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
mprimi: No, at this point it's just a new "connection". As soon as the first message is sent along in that connection, i can associate the stream to an ongoing transfer and then we do have the rest of the context. But not here.

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
mprimi: Oh, i intentionally avoided `-` because i vaguely remember it's a problem for JS.

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
mprimi: Yeah this file is long and this class is complex. It's juggling gRPC streams and reacting to different asynchronous events (stream errors, messages, download cancelled, timeouts, ...).

genie-web/src/main/java/com/netflix/genie/web/properties/AgentFileStreamProperties.java
mprimi: We're not super consistent. Some have dashes, other don't, e.g.: `genie.s3filetransfer`

tgianos: we could at least be consistent about any new ones but if you'd rather sweep through and standardize them all later then that's fine

PR #1009, opened 2020-06-23
File transfers from agent return random bytes

mprimi:
[This recent change](https://github.com/Netflix/genie/commit/86a35d831b299bae5121673b7c2c200fe616a8f8#diff-940e18974ceae4a15af1275a9378170dR413) introduced a serious regression where sometimes the agent-streamed job files consist of random bytes.

Also improve gRPC random stream cancellation we sometime observe.

### DIFF
--- a/genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImplSpec.groovy
+++ b/genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImplSpec.groovy
@@ -19,7 +19,6 @@ package com.netflix.genie.agent.execution.services.impl.grpc
 
 import com.google.common.collect.Maps
 import com.netflix.genie.agent.execution.services.AgentFileStreamService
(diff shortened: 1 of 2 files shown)


genie-agent/src/test/groovy/com/netflix/genie/agent/execution/services/impl/grpc/GRpcAgentFileStreamServiceImplSpec.groovy
```diff
@@ -467,20 +573,20 @@ class GRpcAgentFileStreamServiceImplSpec extends Specification {
 
                 @Override
                 void onNext(final AgentManifestMessage value) {
-                    printf("Received manifest")
+                    println("Received manifest")
```
tgianos, line 576: Shouldn't these be `log` messages instead of straight `println`?

mprimi: It could, but at the moment all messages in groovy spock tests are printed with println

PR #1010, opened 2020-06-25
File downloads hanging for agent transfer timeout

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobDirectoryLimitsServiceImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    public void stop() {
+        if (this.scheduledCheck != null) {
+            this.scheduledCheck.cancel(true);
+        }
+    }
+
+    private void checkFilesSize(final Path jobDirectory) {
+        final DirectoryManifest manifest;
+        try {
+            manifest = this.manifestCreatorService.getDirectoryManifest(jobDirectory);
```
tgianos, line 82: Rather than this can we not just use the same manifest we're periodically sending to the server?

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobDirectoryLimitsServiceImpl.java
```diff
@@ -0,0 +1,125 @@
...
+        if (files > maxFiles) {
+            log.warn("Limit exceeded, too many files: {}/{}", files, maxFiles);
+            this.killService.kill(KillService.KillSource.FILES_LIMIT);
+            return;
+        }
+
+        final DataSize totalSize = DataSize.ofBytes(manifest.getTotalSizeOfFiles());
+        final DataSize maxTotalSize = this.properties.getMaxTotalSize();
+        if (totalSize.toBytes() > maxTotalSize.toBytes()) {
+            log.warn("Limit exceeded, job directory too large: {}/{}", totalSize, maxTotalSize);
```
tgianos, line 99: Since you're about to kill it I'd think we should make these `error` level messages

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobDirectoryLimitsService.java
```diff
@@ -0,0 +1,42 @@
...
+import java.nio.file.Path;
+
+/**
+ * Service that monitors the job directory and may decide to kill the job if some limit is exceeded.
+ * For example, if a file grows larger than some amount.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public interface JobDirectoryLimitsService {
```
tgianos, line 29: I might lose the plural on `Limit` for the name or maybe just make it `JobDirectoryCheckerService` therefore more generic than just checking limits if we decide to add some other sanity checking

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobDirectoryLimitsServiceImpl.java
mprimi: The creator service internally caches and recomputes periodically.
So this call does not force the creation of a new manifest from scratch.

tgianos: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobDirectoryLimitsServiceImpl.java
mprimi: Done

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/JobDirectoryLimitsService.java
mprimi: Done, went with `JobMonitorService`

PR #1012, opened 2020-06-28
Database Changes

tgianos:
- Increase date field precision to max 6 decimal points
- Remove unused `requested_archive_location_prefix` field from jobs table
- Add `archive_status` column to jobs table
- Add` launcher_ext` column to jobs table
- Add `requested_launcher_ext` column to jobs table
- Add `launcher_ext` column to commands table
- Add attribute converter for JSON columns to get rid of custom code embedded in other converter classes
- Unpin H2 version

Note: These changes only exist within the data tier and the associated DTOs are not yet plumbed through until we're sure how/what/where we want to do them


genie-web/src/main/resources/db/migration/mysql/V4_0_0_6__Change_Timestamp_Precision.sql
```diff
@@ -0,0 +1,49 @@
+/*
```
mprimi, line 1: No indexing for the new columns. Just checking this is deliberate and not forgotten.
(makes sense to me that we don't need it, at least for the current/expected usage)

tgianos: Yeah we could put one on the `archive_status` field before we forget. Good call.

As for the TEXT/JSON fields. I tried to move some of the columns to the native `json` types in each of the databases we currently support which would allow indexing to make more sense on those fields. However I got into all kinds of problems (most of which I forget at the moment) so I punted and just made them TEXT fields. In mysql at least indexing these makes no sense really. These fields will all hold some arbitrary structured data so unless we enable full text search I doubt we'll be searching through them. They're more likely to be point queries from the app level at least unless you can think fo something else?

mprimi: I agree indexing those JSON fields is probably a waste of resources.

PR #1014, opened 2020-06-29
Stop agent running a job if job was marked failed

mprimi:


genie-web/src/main/java/com/netflix/genie/web/properties/HeartBeatProperties.java
```diff
@@ -0,0 +1,46 @@
...
+@Setter
+@Validated
+public class HeartBeatProperties {
+
+    /**
+     * Prefix for all properties related to the agent heart beat service.
+     */
+    public static final String PROPERTY_PREFIX = "genie.agent.heart-beat";
+
+    private Duration sendInterval = Duration.ofSeconds(5);
```
tgianos, line 44: As a FYI if you put a javadoc comment on these properties the spring boot configuration properties jar should help generate metadata that people can use in IDE's for autocompleting the properties.

genie-agent/src/main/resources/genie-agent-defaults.yml
```diff
@@ -64,6 +64,7 @@ genie:
         max-files: 64000
         max-file-size: 8GB
         max-total-size: 16GB
+        check-remote-job-status: true
```
tgianos, line 67: Should we roll this out in two phases just to avoid any chance of a client sitting out there released if we have to roll the server back for some reason?

mprimi: Not sure what situation you're hinting at.

But we can roll this code out with this feature disabled by setting a fast property on the server.

PR #1017, opened 2020-07-08
Refactor agent shutdown and fix Zookeeper bug

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/spring/autoconfigure/AgentAutoConfiguration.java
```diff
@@ -90,46 +92,93 @@ public FileLockFactory fileLockFactory() {
...
+     *
+     * @param agentProperties the agent properties
+     * @return a customizer for the task executor
+     */
+    @Bean
+    TaskExecutorCustomizer taskExecutorCustomizer(final AgentProperties agentProperties) {
+        return taskExecutor -> {
+            taskExecutor.setWaitForTasksToCompleteOnShutdown(true);
+            taskExecutor.setAwaitTerminationSeconds(
+                (int) agentProperties.getShutdown().getSystemExecutorLeeway().getSeconds()
```
tgianos, line 164: You may want to see if you can set a `@min` in your property class for `Duration` objects or some sort of validation because if the person sets like `400ms` as the value (totally hypothetical) I think this will round it down to 0 seconds

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -976,4 +976,24 @@ The server also has a chance to override them during the 'Agent Configuration' e
 | 1.2
 |
 
+| `genie.agent.runtime.shutdown.internal-executors-leeway`
```
tgianos, line 984: There are enough agent properties at this point that we might want to think about breaking them out into two separate docs for agent and server properties

genie-agent/src/main/java/com/netflix/genie/agent/cli/ExecCommand.java
```diff
@@ -45,26 +49,52 @@
...
     public ExitCode run() {
-        for (final sun.misc.Signal s : signalsToIntercept) {
-            sun.misc.Signal.handle(s, signal -> handleTerminationSignal());
-        }
+
+        // Lock-free since the only other thread accessing this has not been registered yet
+        this.isRunning = true;
+
+        // Before execution starts, add shutdown hooks
+        Runtime.getRuntime().addShutdownHook(this.threadFactory.newThread(this::waitForCleanShutdown));
```
tgianos, line 96: General comment: if we can ever get the Agent to java 9+ we can leverage improvements in [Process API](https://www.baeldung.com/java-9-process-api) that may help with things beyond just the shutdown hook.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/KillServiceImpl.java
```diff
@@ -33,54 +38,54 @@
...
+     * Hopefully, execution terminates due to regular shutdown procedures before this mechanism kicks-in.
+     */
+    @SuppressFBWarnings("DM_EXIT") // For calling System.exit
+    private void emergencyStop() {
+        try {
+            Thread.sleep(this.agentProperties.getEmergencyShutdownDelay().toMillis());
+        } catch (InterruptedException e) {
+            log.warn("Emergency shutdown thread interrupted");
         }
+        System.exit(ExitCode.EXEC_ABORTED.getCode());
```
tgianos, line 89: May be overcomplicating things here but there are [other ways to shut down a spring application](https://www.baeldung.com/spring-boot-shutdown#exit) that don't necessarily involve explicitly calling `System.exit`. Up to you whether it's worth it.

mprimi: This is really a last resort mechanism that should never kick in. 
If it ends up being used, it's because the Spring shutdown somehow got stuck.

genie-agent/src/main/java/com/netflix/genie/agent/spring/autoconfigure/AgentAutoConfiguration.java
mprimi: @tgianos added a commit in this PR that adds validation to all agent properties.

PR #1020, opened 2020-07-15
Improvements to signal handling

mprimi:


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -540,6 +540,39 @@ String composeScript() {
 
             sb.append(NEWLINE);
 
+            // Script Signal Handling Section
+            final String toStdErr = " >&2";
+            final List<String> trappedSignals = Lists.newArrayList("SIGTERM", "SIGINT", "SIGHUP");
+            sb
+                .append("# Trap exit signals to ensure children processes are dead before returning").append(NEWLINE)
```
tgianos, line 547: I don't really see any variables in here (may have missed one) can we just make this whole block a constant rather than appending them all every time?

mprimi: `trappedSignals` should be a constant. I was toying with the idea of making it a Property but it's not worth it. Can't imagine a case where it's set to anything other than these 3 signals.

PR #1022, opened 2020-07-21
Track job files archive status in database

mprimi:
Track the status of archived files in the database. And use said status when serving requests.

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/ChangeJobArchiveStatusException.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/ChangeJobArchiveStatusException.java
@@ -1,0 +1,46 @@
+/*
+ *
+ *  Copyright 2020 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.genie.agent.execution.exceptions;
+
+/**
+ * Failure to update the job archive status remotely.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public class ChangeJobArchiveStatusException extends Exception {
+
+    /**
+     * Constructor with message.
+     *
+     * @param message message
+     */
+    public ChangeJobArchiveStatusException(final String message) {
+        super(message);
+    }
+
+    /**
+     * Constructor with message and cause.
+     *
+     * @param message message
+     * @param cause   cause
+     */
+    public ChangeJobArchiveStatusException(final String message, final Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/genie-common/src/main/java/com/netflix/genie/common/dto/JobExecution.java
+++ b/genie-common/src/main/java/com/netflix/genie/common/dto/JobExecution.java
@@ -19,6 +19,7 @@ package com.netflix.genie.common.dto;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.netflix.genie.common.external.dtos.v4.ArchiveStatus;
 import lombok.Getter;
 
 import javax.annotation.Nullable;
@@ -69,6 +70,7 @@ public class JobExecution extends BaseDTO {
         message = "The amount of memory this job is set to use on the system"
     )
     private final Integer memory;
+    private final ArchiveStatus archiveStatus;
 
     /**
      * Constructor used by the builder build() method.
@@ -83,6 +85,7 @@ public class JobExecution extends BaseDTO {
         this.exitCode = builder.bExitCode;
         this.memory = builder.bMemory;
         this.timeout = builder.bTimeout;
+        this.archiveStatus = builder.bArchiveStatus;
     }
 
     /**
@@ -131,6 +134,15 @@ public class JobExecution extends BaseDTO {
     }
 
     /**
+     * Get the archival status of job files.
+     *
+     * @return the archival status as Optional as it could be null
+     */
+    public Optional<ArchiveStatus> getArchiveStatus() {
+        return Optional.ofNullable(this.archiveStatus);
+    }
+
+    /**
      * A builder to create job requests.
      *
      * @author tgianos
@@ -144,6 +156,7 @@ public class JobExecution extends BaseDTO {
         private Instant bTimeout;
         private Integer bExitCode;
         private Integer bMemory;
+        private ArchiveStatus bArchiveStatus;
 
         /**
          * Constructor which has required fields.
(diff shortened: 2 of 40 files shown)


genie-common/src/main/java/com/netflix/genie/common/dto/JobExecution.java
```diff
@@ -130,6 +135,15 @@ protected JobExecution(final Builder builder) {
         return Optional.ofNullable(this.memory);
     }
 
+    /**
+     * Get the archival status of job files.
+     *
+     * @return the archival status as Optional as it could be null
```
tgianos, line 141: you have the `private final` variable as `@NotNull` which doesn't line up with this comment or with the fact this returns and `Optional`

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/converters/EntityV3DtoConverters.java
```diff
@@ -162,6 +163,12 @@ public static JobExecution toJobExecutionDto(final JobExecutionProjection jobExe
         }
         jobExecutionProjection.getExitCode().ifPresent(builder::withExitCode);
         jobExecutionProjection.getMemoryUsed().ifPresent(builder::withMemory);
+        if (jobExecutionProjection.getArchiveStatus().isPresent()) {
+            // May throw IllegalArgumentException. Trusting the DB to have only valid values.
```
tgianos, line 167: could instead catch and return `UNKNOWN`

genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImplJobsIntegrationTest.java
```diff
@@ -1224,6 +1226,25 @@ void canGetJobHostInformation() {
         Assertions.assertThat(randomInfo.getTotalMemoryUsed()).isEqualTo(0L);
     }
 
+    @Test
+    @DatabaseSetup("persistence/jobs/init.xml")
```
tgianos, line 1230: you may want to break this `setup` away from this global init as it make it fragile if someone else accidently changes that. I've been trying to scope them to specific tests -> specific init files that way one change to support test A doesn't break tests B, C, D

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
```diff
@@ -1912,6 +1920,32 @@ public void updateJobStatus(
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void updateJobArchiveStatus(
+        @NotBlank(message = "No job id entered. Unable to update.") final String id,
+        @NotNull(message = "Status cannot be null.") final ArchiveStatus archiveStatus
+    ) throws NotFoundException {
+        log.debug(
+            "[updateJobArchiveStatus] Requested to chang the archive status of job {} to {}",
```
tgianos, line 1932: typo in `chang` -> `change`

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
```diff
@@ -1937,6 +1971,29 @@ public JobStatus getJobStatus(@NotBlank final String id) throws NotFoundExceptio
...
+            );
+        } catch (IllegalArgumentException e) {
+            return ArchiveStatus.UNKNOWN;
+        }
+//
+//        return ArchiveStatus.valueOf(
+//            this.jobRepository.findByUniqueId(id, JobExecutionProjection.class)
+//                .orElseThrow(() -> new NotFoundException("No job with id " + id + " exists"))
+//                .getArchiveStatus().orElse(ArchiveStatus.UNKNOWN.name())
+//        );
```
tgianos, line 1994: left in purposely?

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/repositories/JpaJobRepository.java
```diff
@@ -221,6 +221,15 @@ JobInfoAggregate getHostJobInfo(
     @Query("SELECT j.agentHostname FROM JobEntity j WHERE j.uniqueId = :id")
     Optional<String> getJobHostname(@Param("id") String id);
 
+    /**
+     * Get only the archive status of a job.
+     *
+     * @param id The id of the job to get the archive status for
+     * @return The job archive status or {@link Optional#empty()} if no job with the given id exists
+     */
+    @Query("SELECT COALESCE(j.archiveStatus, 'UNKNOWN') FROM JobEntity j WHERE j.uniqueId = :id")
```
tgianos, line 230: 👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/exceptions/ChangeJobArchiveStatusException.java
```diff
@@ -0,0 +1,46 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
```
tgianos, line 3: copy paste

genie-web/src/main/java/com/netflix/genie/web/services/JobDirectoryServerService.java
```diff
@@ -39,18 +41,24 @@
...
     void serveResource(
         String jobId,
         URL baseUrl,
         String relativePath,
         HttpServletRequest request,
-        HttpServletResponse response
+        HttpServletResponse response,
+        JobStatus jobStatus,
+        ArchiveStatus jobArchiveStatus,
+        boolean isV4
```
tgianos, line 62: I understand this optimization but I've avoided it as 1. it puts onus on calling services to know these pieces of information 2. Pushing the checks for them down as far as possible means we minimize the risk of looking in the wrong place due to a change in the time since original check and when we get here

genie-common/src/main/java/com/netflix/genie/common/dto/JobExecution.java
mprimi: Removing the @NotNull and sticking to the convention followed by other fields.

genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImplJobsIntegrationTest.java
mprimi: This `init.xml` is used and shared by numerous other tests in the same file. For example get/set status.
They're all very similar and I think it's appropriate they share the same test data.
If i was to clone it, it would be an identical copy of the existing one, and next time something changes one has to go and update multiple identical files.

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
mprimi: 😳 ... yes?

genie-web/src/main/java/com/netflix/genie/web/services/JobDirectoryServerService.java
mprimi: I see your point in (1), tho one could argue this is telling the service where to look / how to behave so it's not completely inappropriate.

But (2) I don't think is a valid reason. The calls in their original place happen 10 or so lines of java after where they are right now. That's probably microseconds.

Regardless. Got more changes coming to this so I may put them back where they were.

PR #1025, opened 2020-07-28
Update controller and DirectoryServer service to use new archive status

mprimi:


genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
```diff
@@ -692,54 +693,69 @@ public void getJobOutput(
...
+                throw new GenieServerException("Job files are not local, but forwarding is disabled");
             }
         }
 
+        // In any other case, delegate the request to the service
         log.debug("Fetching requested resource \"{}\" for job \"{}\"", path, id);
         this.jobDirectoryServerService.serveResource(id, baseUrl, path, request, response);
     }
 
+    private void forwardRequest(
```
tgianos, line 725: This is generically named `forwardRequest` but we also forward kill requests and it looks like this only happens for output requests

genie-web/src/test/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestControllerTest.java
```diff
@@ -628,6 +631,70 @@ void wontForwardJobOutputRequestIfOnCorrectHost() throws GenieException, GenieCh
...
+
+        Assertions.assertThatThrownBy(
+            () -> this.controller.getJobOutput(jobId, null, request, response)
+        ).isInstanceOf(GenieServerException.class);
+
+        Mockito.verify(this.agentRoutingService, Mockito.times(1)).getHostnameForAgentConnection(Mockito.eq(jobId));
+    }
+
+    @Test
+    void wontForwardJobOutputRequestIfForwardingRequestWasForwarded() throws GenieException, GenieCheckedException {
```
tgianos, line 657: could use `@DisplayName` for these JUnit 5 tests to shorten the actual method names if you want FYI

PR #1026, opened 2020-07-30
Fix entity search errors

mprimi:
Under the right circumstances, searching for Application/Cluster/Command can throw an error


genie-web/src/integTest/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceIntegrationTestBase.java
```diff
@@ -65,6 +65,13 @@
         PersistedJobStatusObserver.class //TODO: Needed for JobEntityListener but should be in DataAutoConfiguration
     }
 )
+//@TestPropertySource(
+//    properties = {
+//        "logging.level.com.netflix.genie.web.data.services.impl.jpa=DEBUG", // Genie JPA package
+//        "logging.level.org.hibernate.SQL=DEBUG",                            // Show SQL queries
```
tgianos, line 71: I thought `@DataJpaTest` turned on SQL logging by default already?

mprimi: It does print hibernate queries by default, but not in full detail. These do.

PR #1030, opened 2020-08-03
Bug: forward stdin to command at launch

mprimi:
The recent change of logic to better handle signals (PR #1020) did not take into account interactive commands.
After this change, any command that tried to read from stdin, would immediately find EOF.

Update the run script so that the bash process stdin is forwarded to the actual running command.

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
@@ -695,7 +695,7 @@ class JobSetupServiceImpl implements JobSetupService {
             sb
                 .append("# Launch the command")
                 .append(NEWLINE)
-                .append(this.commandLine).append(" &").append(NEWLINE)
+                .append(this.commandLine).append(" <&0 &").append(NEWLINE)
                 .append("wait %1").append(NEWLINE)
                 .append("exit $?").append(NEWLINE)
                 .append(NEWLINE);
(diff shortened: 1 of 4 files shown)


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -695,7 +695,7 @@ String composeScript() {
             sb
                 .append("# Launch the command")
                 .append(NEWLINE)
-                .append(this.commandLine).append(" &").append(NEWLINE)
+                .append(this.commandLine).append(" <&0 &").append(NEWLINE)
```
jzhuge, line 698: What was reason in #1020 to turn
```bash
command
```
into
```bash
command &
wait %1
```

jzhuge, line 698: Any doc on this `<&0`? I googled but only found vague explanations.

mprimi: https://github.com/Netflix/genie/pull/1020/commits/6a7c0d77c18d6bb950916e731f8f548860626e2c
Depending on the command, the bash process running `run` was terminating but leaving the underlying command running.
That change introduces a bash trap so that the job command is always terminated before a job is marked finished.

mprimi: It's just bash redirection (https://www.tldp.org/LDP/abs/html/io-redirection.html, https://www.gnu.org/software/bash/manual/html_node/Redirections.html).

Specifically, the `stdin` of `bash run` process is forwarded to the actual command.
So in case of interactive CLI commands, when you type `select * from foo`, that line goes to command, rather than bash.

akira:
👍

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
jzhuge: Understand the redirection part, but never seen `<&0` to redirect parent stdin to child. Good to know.

PR #1031, opened 2020-08-06
Expose page and size parameters for job search

rboorgapally:
Currently, the getJobs method doesn't expose the page and the search results size parameter in the client library. As a result we can only fetch the first 10 search results. There is no way to request additional search results using the client library. The API supports retrieving additional search results using the page and the size parameters. 

This pull request change exposes those parameters in the getJobs method so that users can retrieve all the results.

### DIFF
--- a/genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
+++ b/genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
@@ -219,6 +219,7 @@ abstract class JobClientIntegrationTest extends ClusterClientIntegrationTest {
                     null,
                     null,
                     null,
+                    null,
                     null
                 )
             )
--- a/genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
+++ b/genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
@@ -219,9 +219,6 @@ abstract class JobClientIntegrationTest extends ClusterClientIntegrationTest {
                     null,
                     null,
                     null,
-                    null,
-                    null,
-                    null,
                     null
                 )
             )
--- a/genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
+++ b/genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
@@ -218,6 +218,10 @@ abstract class JobClientIntegrationTest extends ClusterClientIntegrationTest {
                     null,
                     null,
                     null,
+                    null,
+                    null,
+                    null,
+                    null,
                     null
                 )
             )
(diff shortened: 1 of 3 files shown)


genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
```diff
@@ -218,6 +218,8 @@ void canSubmitJob() throws Exception {
                     null,
                     null,
                     null,
+                    null,
+                    null,
```
mprimi, line 222: A test that uses the newly added page size would be nice.
i.e. iterate over 3 jobs using page size 1

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -191,7 +191,9 @@ public void writeTo(final BufferedSink sink) throws IOException {
      * @throws IOException          For Network and other IO issues.
      */
     public List<JobSearchResult> getJobs() throws IOException, GenieClientException {
-        return this.getJobs(null, null, null, null, null, null, null, null, null, null, null, null, null, null, null);
+        return this.getJobs(null, null, null, null, null, null,
+            null, null, null, null, null, null,
```
mprimi, line 195: Can you do one parameter per line (repository style convention, I'm surprised checkstyle is ok with this.

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
```diff
@@ -305,7 +314,9 @@ public void writeTo(final BufferedSink sink) throws IOException {
         @Nullable final Long minFinished,
         @Nullable final Long maxFinished,
         @Nullable final String grouping,
-        @Nullable final String groupingInstance
+        @Nullable final String groupingInstance,
+        @Nullable final Integer page,
+        @Nullable final Integer size
```
mprimi, line 319: This is an API breaking change. Anyone using `getJobs` today will get compilation error after updating.
Can you instead create a new method and telescope?

genie-client/src/main/java/com/netflix/genie/client/apis/JobService.java
```diff
@@ -112,7 +115,9 @@
         @Query("minFinished") Long minFinished,
         @Query("maxFinished") Long maxFinished,
         @Query("grouping") String grouping,
-        @Query("groupingInstance") String groupingInstance
+        @Query("groupingInstance") String groupingInstance,
+        @Query("page") Integer page,
+        @Query("size") Integer size
```
mprimi, line 120: Any particular reason for not including the other 2 parameters (order/direction)

genie-client/src/main/java/com/netflix/genie/client/JobClient.java
rboorgapally: It complained about line length, so I split this way. Will make it one parameter per line.

mprimi:
Deprecated by #1033

PR #1032, opened 2020-08-08
Add leader task ArchiveStatusCleanup

mprimi:
Mutate archive status from PENDING to UNKNOWN if the job has completed more than N minutes ago and the agent disappeared.

This covers primarily 2 cases:
 - Agent crashing or partitioned from server after updating the final job status
 - Old agents that do not include the new logic to update archive status after archiving

### DIFF
--- a/genie-web/src/integTest/resources/com/netflix/genie/web/data/services/impl/jpa/persistence/jobs/archive_status.xml
+++ b/genie-web/src/integTest/resources/com/netflix/genie/web/data/services/impl/jpa/persistence/jobs/archive_status.xml
@@ -1,0 +1,228 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright 2020 Netflix, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<dataset>
+
+    <commands
+        id="1"
+        created="2016-03-21 01:47:00"
+        updated="2016-03-21 01:59:00"
+        entity_version="0"
+        unique_id="command1"
+        genie_user="tgianos"
+        name="spark"
+        version="1.6.0"
+        check_delay="10000"
+        status="ACTIVE"
+    />
+
+    <command_executable_arguments
+        command_id="1"
+        argument="spark"
+        argument_order="0"
+    />
+
+    <clusters
+        id="1"
+        created="2016-03-21 01:49:00"
+        updated="2016-03-21 02:59:00"
+        entity_version="0"
+        unique_id="cluster1"
+        genie_user="tgianos"
+        name="h2query"
+        version="2.4.0"
+        status="UP"
+    />
+
+    <criteria
+        id="0"
+        name="spark"
+    />
+
+    <!-- Should match with the right threshold -->
+    <jobs
+        id="1"
+        created="2020-01-01 00:00:00"
+        updated="2020-01-01 00:00:00"
+        entity_version="1"
+        unique_id="agentJob1"
+        genie_user="tgianos"
+        name="testSparkJob"
+        version="2.4"
+        command_criterion="0"
+        requested_cpu="1"
+        requested_memory="1560"
+        requested_timeout="608400"
+        status="SUCCEEDED"
+        cluster_id="1"
+        command_id="1"
+        check_delay="10000"
+        timeout_used="608400"
+        requested_id="true"
+        job_directory_location="/tmp/genie/jobs/1"
+        resolved="true"
+        v4="true"
+        cluster_name="h2query"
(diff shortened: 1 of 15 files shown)


genie-web/src/integTest/resources/com/netflix/genie/web/data/services/impl/jpa/persistence/jobs/archive_status.xml
```diff
@@ -0,0 +1,228 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright 2016 Netflix, Inc.
```
tgianos, line 3: Copy paste

genie-web/src/integTest/resources/com/netflix/genie/web/data/services/impl/jpa/persistence/jobs/archive_status.xml
```diff
@@ -0,0 +1,228 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<dataset>
+
+    <commands
```
tgianos, line 19: Note not all this data is needed. You can parse it down a lot if you want.

genie-web/src/main/java/com/netflix/genie/web/data/services/PersistenceService.java
```diff
@@ -970,6 +970,18 @@ void updateJobArchiveStatus(
...
+     * Get the set of agent jobs that are satisfy the following criteria.
+     * - Agent job
+     * - Final state
+     * - Finished time earlier than N seconds ago
+     * - Archive status is PENDING
+     *
+     * @param updatedThreshold only select jobs update before this threshold
+     * @return The list of job ids
+     */
+    Set<String> getFinishedJobsWithPendingArchiveStatus(@NotNull Instant updatedThreshold);
```
tgianos, line 984: Any thought to making this a bit more generic? `getJobsWithStatusAndArchiveStatus` or something? That way it could be used for more use cases in future if necessary without developing entirely new API

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
```diff
@@ -2212,6 +2222,21 @@ public JobInfoAggregate getHostJobInformation(@NotBlank final String hostname) {
         log.debug("[getHostJobInformation] Called for hostname {}", hostname);
         return this.jobRepository.getHostJobInfo(hostname, ACTIVE_STATUS_SET, USING_MEMORY_JOB_SET);
     }
+
+    /**
+     * {@inheritDoc}
+     * @param updatedThreshold
```
tgianos, line 2228: redundant `@param` tag here

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ArchiveStatusCleanupTask.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void run() {
+        final Set<Tag> tags = Sets.newHashSet();
+        final long start = System.nanoTime();
+        try {
+            final Instant updatedThreshold = Instant.now().minus(this.properties.getGracePeriod());
```
tgianos, line 86: These won't support fast property changes at the moment. Just pointing that out in case you could think of any reasons we might want to change these without a redeploy

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ArchiveStatusCleanupTask.java
```diff
@@ -0,0 +1,140 @@
...
+            final Set<String> jobIds =
+                this.persistenceService.getFinishedJobsWithPendingArchiveStatus(updatedThreshold);
+            if (!jobIds.isEmpty()) {
+                log.debug("Found {} finished jobs with PENDING archive status", jobIds.size());
+                this.clearJobsArchiveStatus(jobIds);
+            }
+            MetricsUtils.addSuccessTags(tags);
```
tgianos, line 93: Not specific to this PR but I wonder if we can roll this adding of success or failure tags into an aspect somehow so we don't have to write the same thing all over the place.

genie-web/src/main/java/com/netflix/genie/web/data/services/PersistenceService.java
mprimi: Done

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ArchiveStatusCleanupTask.java
mprimi: If one day the properties class start supporting fast properties, this will need no change.
Manually bypassing properties classes and using raw `Environment` is ugly and I'm avoiding it unless strictly necessary.

genie-web/src/main/java/com/netflix/genie/web/tasks/leader/ArchiveStatusCleanupTask.java
mprimi: Would be nice, maybe annotate such metods with `@GenieMetric(name="...")` or something like that.

PR #1033, opened 2020-08-10
Java client pagination API

mprimi:


genie-client/src/main/java/com/netflix/genie/client/ClusterClient.java
```diff
@@ -129,21 +132,66 @@ public String createCluster(
...
+    /**
+     * Method to get a list of all the clusters from Genie for the query parameters specified.
+     *
+     * @param name          The name of the cluster.
+     * @param statusList    The list of statuses.
+     * @param tagList       The list of tags.
+     * @param minUpdateTime Minimum Time after which cluster was updated.
+     * @param maxUpdateTime Maximum Time before which cluster was updated.
+     * @param pageSize      The maximum number of results returned
+     * @param sortAttribute The Application attribute used to sort
```
raveeram, line 167: Nit: 
```suggestion
     * @param sortAttribute The Cluster attribute used to sort
```
in this case? Same for command below as well.

genie-client/src/main/java/com/netflix/genie/client/apis/SortAttribute.java
```diff
@@ -0,0 +1,53 @@
...
+package com.netflix.genie.client.apis;
+
+/**
+ * Sort key for entity search results.
+ * This is a subset of all sort attributes supported by the server.
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+public enum SortAttribute {
```
tgianos, line 27: FYI the server allows sorting by multiple attributes at once and this mechanism limits it to singular

genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
```diff
@@ -143,6 +145,34 @@ void canSubmitJob() throws Exception {
...
+            .build();
+
+        final long waitStart = System.currentTimeMillis();
+        final long maxTotalWait = 120000;
+
+        for (final Map.Entry<String, JobStatus> entry : expectedStatuses.entrySet()) {
+            final String jobId = entry.getKey();
+            final JobStatus status = entry.getValue();
+
+            System.out.println("Waiting for job: " + jobId + " (expected final status: " + status.name() + ")");
```
tgianos, line 163: `log.warn` or even `info` seems better than leaving `System.out` in.

genie-client/src/main/java/com/netflix/genie/client/apis/SortAttribute.java
mprimi: Also only allows sorting by 3 statically defined attributes. Seems like a good tradeoff in terms of convenience to me.
Anyone using this API doesn't have to think too hard because the enums make it impossible to misuse it.

genie-client/src/integTest/java/com/netflix/genie/client/JobClientIntegrationTest.java
mprimi: The reason for stdout here is that it shows up in the test reports. So when smoke tests fail we can at least learn the job ID and look at server logs. Whereas lines printed with slf4j logger are forever lost.

PR #1035, opened 2020-08-18
Make Agent routing more responsive

mprimi:
With the previous curator-based implementation, the right timing of events was resulting in the routing table lagging seconds behind reality.
With enough jobs, these unlikely circumstances manifest multiple times a day, resulting in requests failing to route.

Re-implement AgentRoutingService to respond faster to change.

### DIFF
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -132,6 +132,11 @@ https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-featu
 |false
 |no
 
+|genie.agent.routing.refresh-interval
+|Interval at which individual connections are refreshed
+|3s
+|no
+
 |genie.agent.runtime.*
 |Properties with this prefix are forwarded to each agent during startup
 |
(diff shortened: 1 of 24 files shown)


genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -132,6 +132,11 @@ https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-featu
 |false
 |no
 
+|genie.agent.routing.refresh-interval
+|Interval at which individual connections are refreshed
+|3s
```
tgianos, line 137: You said in the PR message that it could be seconds behind with the curator solution but now this says 3 seconds as a default. Is that actually better than whatever curator was doing? I ask just cause some of the logic in here seems to be getting relatively complicated and I thought we introduced the zk stuff to make it simple

mprimi: There is some complexity here that can't be avoided... ZK connection status can change at any time. Other nodes with a stale view overwriting valid routing data, agents connecting, disconnecting, reconnecting to the same or another host. etc.
Plus there's other complexity such as making the service API non-blocking, so it doesn't tie up gRRP threads.

The main difference from this implementation to the previous is:
 - Old: Latent changes are periodically applied by a thread that is on an exponential backoff idle delay. In addition to the delay, it was performing a full sync-refresh, which could take hundreds of milliseconds.
 - New: changes are consumed one by one in priority order by a thread that is waiting on a blocking queue. Normally the change is applied as soon as it lands in the queue.

The 3 second delay in properties is just how often an entry is checked and refreshed.

PR #1038, opened 2020-09-02
Sketch: Integrate S3 attachment service

mprimi:
@tgianos poked around a bit about implementing S3 attachments. 
This seems to be the most straightforward and simple way.

 - It does not mark Files as attachment (could be added later)
 - It does not asynchronously upload while the job gets saved (could be added later)

I think the main contention point is going to be using attachment service directly in the JobRestController.
To me it's not too bad in terms of encapsulation, but maybe you disagree.
Pushing it down into job launch service requires more interfaces to change (launch service).

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
@@ -57,7 +57,7 @@ import com.netflix.genie.web.data.services.PersistenceService;
 import com.netflix.genie.web.dtos.JobSubmission;
 import com.netflix.genie.web.exceptions.checked.NotFoundException;
 import com.netflix.genie.web.properties.JobsProperties;
-import com.netflix.genie.web.services.AttachmentService;
+import com.netflix.genie.web.services.LegacyAttachmentService;
 import com.netflix.genie.web.services.JobCoordinatorService;
 import com.netflix.genie.web.services.JobDirectoryServerService;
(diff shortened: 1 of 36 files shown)


genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public class S3AttachmentService implements AttachmentService {
+
+    @Override
+    public Set<URI> saveAttachments(final String jobId, final Set<Resource> attachments) throws SaveAttachmentException {
+        // Check for limits (individual attachments size and total size)
```
tgianos, line 45: The limits should already have been handled by spring before it's even handed to our controller i believe

genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
```diff
@@ -946,6 +946,8 @@ private String agentExecution(
 
         if (attachments != null) {
             jobSubmissionBuilder.withAttachments(
+                // Call AttachmentService.save here and stuff URIs into jobSubmission
```
tgianos, line 949: One of the main reasons I have for pushing this down into the `JobLaunchService` now is so that when the V4 API comes we don't have to either copy/paste all this into that controller or move it then and modify a bunch of existing code.

PR #1039, opened 2020-09-04
Job environment debug dump filter (and more)

mprimi:
* Adjust verbosity of GRpcHeartBeatServiceImpl
* Configurable filtering of job environment debug dump
* Reorder properties

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
@@ -25,6 +25,8 @@ import com.netflix.genie.agent.execution.exceptions.DownloadException;
 import com.netflix.genie.agent.execution.exceptions.SetUpJobException;
 import com.netflix.genie.agent.execution.services.DownloadService;
 import com.netflix.genie.agent.execution.services.JobSetupService;
+import com.netflix.genie.agent.properties.AgentProperties;
+import com.netflix.genie.agent.properties.JobSetupServiceProperties;
 import com.netflix.genie.agent.utils.PathUtils;
 import com.netflix.genie.common.external.dtos.v4.ExecutionEnvironment;
 import com.netflix.genie.common.external.dtos.v4.JobSpecification;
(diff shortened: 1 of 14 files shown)


genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -382,14 +389,18 @@ private void addEntitiesFilesToManifest(
...
         private final List<Pair<String, String>> setupFileReferences;
         private final String commandLine;
         private final List<Pair<String, String>> localEnvironmentVariables;
         private final List<Pair<String, String>> serverEnvironmentVariables;
 
-        JobScriptComposer(final JobSpecification jobSpecification, final File jobDirectory) {
+        JobScriptComposer(
+            final JobSpecification jobSpecification,
+            final File jobDirectory
+            ) {
```
tgianos, line 403: Nit: formatting looks off here

travis/buildViaTravis.sh
```diff
@@ -17,6 +17,9 @@ if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
   INTEGRATION_TEST_DB=postgresql ${GRADLE} ${GRADLE_OPTIONS} genie-web:integrationTest
   # Build Docker images and compile documentation
   ${GRADLE} ${GRADLE_OPTIONS} javadoc asciidoc dockerBuildAllImages
+elif [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_TAG" == "" ] && [ "$TRAVIS_TAG" == "dev-snapshot" ]; then
```
tgianos, line 20: shouldn't this be `"$TRAVIS-BRANCH" == "dev-snapshot"`?

I'd prefer to look into ways of speeding up the regular build than building in hack arounds like this. This will still cause `latest.snapshot` to end up being these artifacts even if they would have failed some check

tgianos (approved):

travis/buildViaTravis.sh
mprimi: The purpose here is to make snapshots available for downstream testing without having to merge into master.
It's a mechanism to use in a selected set of situations (it requires a force push to a specific upstream branch).
The shortened build was just a convenience.

PR #1042, opened 2020-09-11
Implement S3 file attachment service

mprimi:
Rather than storing attachments on local disk of the Genie node that received the request, upload the file to S3 where it can be later retrieved by the agent (possibly executing elsewhere).

### DIFF
--- a/genie-docs/src/docs/asciidoc/_metrics.adoc
+++ b/genie-docs/src/docs/asciidoc/_metrics.adoc
@@ -197,6 +197,31 @@ It will have to be added if one is desired otherwise metrics are just published 
 |ArchiveStatusCleanupTask
 |status, exceptionClass
 
+|genie.jobs.attachments.s3.count.distribution
+|Distribution summary of the number of files attached
+|count
+|S3AttachmentServiceImpl
+|
+
+|genie.jobs.attachments.s3.largestSize.distribution
+|Distribution summary of the size of the largest file attached
+|bytes
+|S3AttachmentServiceImpl
+|
+
+|genie.jobs.attachments.s3.totalSize.distribution
+|Distribution summary of the total size of the files attached
+|bytes
+|S3AttachmentServiceImpl
+|
+
+|genie.jobs.attachments.s3.upload.timer
+|genie.jobs.attachments.s3.upload.timer
+|Time taken to upload job attachments to S3 (only measured for jobs with attachments)
+|nanoseconds
+|S3AttachmentServiceImpl
+|status, exceptionClass
+
 |genie.jobs.clusters.selectors.script.select.timer
 |Time taken by the loaded script to select a cluster among the one passed as input
 |nanoseconds
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -17,6 +17,16 @@ Whereas static properties values are bound during application startup and do not
 |===
 |Property |Description |Default Value |Dynamic
 
+|genie.agent.connection-tracking.cleanup-interval
+|Interval at which the cleanup task runs
+|2s
+|no
+
+|genie.agent.connection-tracking.connection-expiration-period
+|How long after the last heartbeat an agent connection is marked expired
+|10s
+|no
+
 |genie.agent.configuration.agent-properties-filter-pattern
 |Regular expression applied to filter server properties that are forwarded to the agent
 |^genie\.agent\.runtime\..*
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -230,6 +230,21 @@ Health of the system is marked unhealthy if the CPU load of a system goes beyond
 |null
 |yes
 
+|genie.jobs.attachments.location-prefix
+|Common prefix where attachments are stored
+|s3://genie/attachments
+|no
+
+|genie.jobs.attachments.max-size
+|Maximum size of an attachment
+|100MB
+|no
+
+|genie.jobs.attachments.max-total-size
+|Maximum size of all attachments combined (Spring and Tomcat may also independently limit the size of upload)
+|150MB
+|no
+
 |genie.jobs.cleanup.deleteDependencies
 |Whether or not to delete the dependencies directories for applications, cluster, command to save disk space after job completion
 |true
@@ -293,7 +308,7 @@ cases is specified as part of the `Command` entity for a particular job.
 
 |genie.jobs.locations.attachments
 |The default root location where job attachments will be temporarily stored. Scheme should be included. Created if
-doesn't exist.
+doesn't exist (deprecated, see genie.jobs.attachments.* properties)
 |file://${java.io.tmpdir}genie/attachments/
 |no
 
--- a/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
@@ -57,7 +57,7 @@ import com.netflix.genie.web.data.services.PersistenceService;
 import com.netflix.genie.web.dtos.JobSubmission;
 import com.netflix.genie.web.exceptions.checked.NotFoundException;
 import com.netflix.genie.web.properties.JobsProperties;
-import com.netflix.genie.web.services.AttachmentService;
+import com.netflix.genie.web.services.LegacyAttachmentService;
 import com.netflix.genie.web.services.JobCoordinatorService;
 import com.netflix.genie.web.services.JobDirectoryServerService;
 import com.netflix.genie.web.services.JobLaunchService;
@@ -159,7 +159,7 @@ public class JobRestController {
     private final Environment environment;
 
     // TODO: V3 Execution only
-    private final AttachmentService attachmentService;
+    private final LegacyAttachmentService legacyAttachmentService;
     private final JobExecutionModeSelector jobExecutionModeSelector;
 
     // Metrics
@@ -180,7 +180,7 @@ public class JobRestController {
      * @param registry                  The metrics registry to use
      * @param agentRoutingService       Agent routing service
      * @param environment               The application environment to pull dynamic properties from
-     * @param attachmentService         The attachment service to use to save attachments.
+     * @param legacyAttachmentService   The attachment service to use to save attachments.
      * @param jobExecutionModeSelector  The execution mode (agent vs. embedded) mode selector
      */
     @Autowired
@@ -197,7 +197,7 @@ public class JobRestController {
         final MeterRegistry registry,
         final AgentRoutingService agentRoutingService,
         final Environment environment,
-        final AttachmentService attachmentService,
+        final LegacyAttachmentService legacyAttachmentService,
         final JobExecutionModeSelector jobExecutionModeSelector
     ) {
         this.jobLaunchService = jobLaunchService;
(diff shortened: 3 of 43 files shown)


genie-docs/src/docs/asciidoc/_metrics.adoc
```diff
@@ -197,6 +197,30 @@ It will have to be added if one is desired otherwise metrics are just published
 |ArchiveStatusCleanupTask
 |status, exceptionClass
 
+|genie.jobs.attachments.s3.count.distribution
+|Distribution summary of for the number of files attached
```
tgianos, line 201: Nit: `of for` seems a typo in this and the one below

genie-docs/src/docs/asciidoc/_metrics.adoc
```diff
@@ -197,6 +197,30 @@ It will have to be added if one is desired otherwise metrics are just published
...
+|count
+|S3AttachmentService
```
tgianos, line 203: By convention we've generally added `Impl` to the end of these implementations of various interfaces or at least we've tried to I think

genie-web/src/integTest/java/com/netflix/genie/web/apis/rest/v3/controllers/RestControllerIntegrationTestBase.java
```diff
@@ -650,4 +660,29 @@ public void describeTo(final Description description) {
             description.appendText(this.mismatchDescription);
         }
     }
+
+    @Configuration
+    static class TestConfiguration {
+
+        private final TemporaryFolder attachmentsDirectory;
```
tgianos, line 667: I believe these tests have all been moved to JUnit 5. With that in mind this is from Junit 4 so should move to [one of the mechanisms](https://www.baeldung.com/junit-5-temporary-directory) for using JUnit 5 temp dir so we can eventually rip out junit 4.

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -230,6 +230,21 @@ Health of the system is marked unhealthy if the CPU load of a system goes beyond
...
+|s3://genie/attachments
+|no
+
+|genie.jobs.attachments.max-size
+|Maximum size of an attachment
+|100MB
+|no
+
+|genie.jobs.attachments.max-total-size
+|Maximum size of all attachments combined (Spring and Tomcat may also independently limit the size of upload)
```
tgianos, line 244: This is true that until now we've used 
```
spring.servlet.multipart.max-file-size | 1MB | Max file size.
spring.servlet.multipart.max-request-size | 10MB | Max request size.
```
from [here](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html#web-properties).

With the addition of these properties someone would now have to tune 4 different properties and make them line up or get confused when it doesn't behave how they want (e.g. spring lets it through but then genie errors or vice versa). 

Curious what are you trying to accomplish by adding these additional properties?

genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
```diff
@@ -946,10 +950,13 @@ private String agentExecution(
 
         if (attachments != null) {
             jobSubmissionBuilder.withAttachments(
-                Arrays
-                    .stream(attachments)
-                    .map(MultipartFile::getResource)
-                    .collect(Collectors.toSet())
+                this.attachmentService.saveAttachments(
```
tgianos, line 953: I think we discussed this offline but curious if you tried putting this in service tier instead and ran into a problem?

genie-web/src/main/java/com/netflix/genie/web/properties/AttachmentServiceProperties.java
```diff
@@ -0,0 +1,55 @@
+/*
+ *
+ *  Copyright 2016 Netflix, Inc.
```
tgianos, line 3: copy paste

genie-web/src/main/java/com/netflix/genie/web/properties/AttachmentServiceProperties.java
```diff
@@ -0,0 +1,55 @@
...
+@Validated
+public class AttachmentServiceProperties {
+
+    /**
+     * The property prefix for job user limiting.
+     */
+    public static final String PROPERTY_PREFIX = "genie.jobs.attachments";
+
+    @NotNull(message = "Attachment location prefix is required")
+    private URI locationPrefix = URI.create("s3://genie/attachments");
```
tgianos, line 47: I think the default should be on disk somewhere probably

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalFileSystemAttachmentService.java
```diff
@@ -0,0 +1,123 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Set<URI> saveAttachments(
+        @Nullable final String jobId,
+        final Set<Resource> attachments
+    ) throws SaveAttachmentException {
+
+        if (attachments.isEmpty()) {
+            return Sets.newHashSet();
```
tgianos, line 79: where possible we should get away from using this Guava way of instantiating collections and just use `new HashSet<>()` due to the outdated version of guava we're forced to use

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
```diff
@@ -0,0 +1,225 @@
...
+     */
+    public S3AttachmentService(
+        final S3ClientFactory s3ClientFactory,
+        final AttachmentServiceProperties attachmentServiceProperties,
+        final MeterRegistry meterRegistry
+    ) {
+        this.s3ClientFactory = s3ClientFactory;
+        this.properties = attachmentServiceProperties;
+        this.meterRegistry = meterRegistry;
+        this.s3BaseURI = new AmazonS3URI(attachmentServiceProperties.getLocationPrefix());
```
tgianos, line 84: You have it as `AttachmentServiceProperties` but then make an assumption here that it's a S3 path. This will throw a runtime exception if it's not a valid S3 URI so the system will fail to start so I guess that's fine just pointing out the assumption that I guess you're saying the deployment needs to be consistent

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
```diff
@@ -0,0 +1,225 @@
...
+        this.meterRegistry.summary(COUNT_DISTRIBUTION).record(attachments.size());
+
+        log.debug("Saving {} attachments for job request with id: {}", attachments.size(), jobId);
+
+        if (attachments.size() == 0) {
+            return EMPTY_SET;
+        }
+
+        // Check for attachment size limits
+        this.checkLimits(attachments);
```
tgianos, line 106: shouldn't this all be within the try? as it is you've missed these in some counter metrics below since you return before they're logged

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
```diff
@@ -0,0 +1,225 @@
...
+            MetricsUtils.addFailureTagsWithException(tags, e);
+            throw e;
+        } finally {
+            this.meterRegistry
+                .timer(SAVE_TIMER, tags)
+                .record(System.nanoTime() - start, TimeUnit.NANOSECONDS);
+        }
+    }
+
+    private void checkLimits(final Set<Resource> attachments) throws SaveAttachmentException {
```
tgianos, line 126: Just pointing out that depending on how the lengths are calculated by a given `Resource` implementation this could result in 2 or 3 read throughs of all the bytes of attachments:
1. Spring at ingestion
2. This check
3. Saving to eventual destination

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
```diff
@@ -0,0 +1,225 @@
...
+        log.debug(
+            "Uploading {} attachments for job request with id {} to: {}",
+            attachments.size(),
+            jobId,
+            commonPrefix
+        );
+
+        final Set<URI> attachmentURIs = Sets.newHashSet();
+
+        for (final Resource attachment : attachments) {
```
tgianos, line 189: Perhaps we could parallelize this?

genie-web/src/main/java/com/netflix/genie/web/spring/autoconfigure/services/ServicesAutoConfiguration.java
```diff
@@ -319,6 +324,24 @@ public FileSystemAttachmentService legacyAttachmentService(final JobsProperties
...
+     * @return The attachment service to use
+     */
+    @Bean
+    @ConditionalOnMissingBean(AttachmentService.class)
+    public S3AttachmentService s3AttachmentService(
+        final S3ClientFactory s3ClientFactory,
+        final AttachmentServiceProperties attachmentServiceProperties,
+        final MeterRegistry meterRegistry
+    ) {
+        return new S3AttachmentService(s3ClientFactory, attachmentServiceProperties, meterRegistry);
```
tgianos, line 342: I think our default should be local file system

genie-web/src/integTest/java/com/netflix/genie/web/apis/rest/v3/controllers/RestControllerIntegrationTestBase.java
mprimi: Ah dang i got confused, i should use @TempDir

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: More flexible configuration.
Rather than relying on the underlying server limit, this limits specifically the attachments.
For example one may not want to touch the upload limit, but still limit attachments to 5MB. This is now possible.

Tuning 4 different properties is not necessary. One can keep the spring/tomcat (maybe not even know they exist) and limit the size of attachments with a simple clearly-named property.

genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
mprimi: Refactoring was complex, specifically updating the embedded codepath.
Since we are about to delete it, I placed this here, which is not completely wrong IMO.
Once embedded is deleted, we can push this down as you suggest.

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalFileSystemAttachmentService.java
mprimi: Just `Sets.newHashSet` appears ~500 times. Until we decide to change convention and rip them all out, i'm gonna stick with it.

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
mprimi: Personally I wouldn't call this an assumption. It's a bean validating a configuration parameter, and failing to initialize if the value is misconfigured for the bean type.

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
mprimi: That's intentional. The upload timer is only recorded if we reach the upload part.
And the distribution counters are only updated if the attachments are not rejected due to their size.

But maybe i should add a metric for limit exceeded.
(i didn't because it's easy to see with other existing metrics)

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentService.java
mprimi: Hmmm i assumed the multi-part upload has to provide size of parts so there's never any need to scan. Gonna double check.

PR #1047, opened 2020-10-05
Add support for multiple agent launchers

mprimi:


genie-web/src/integTest/resources/application-integration.yml
```diff
@@ -29,9 +29,6 @@ genie:
     cluster-selector:
       source: file:///tmp/genie/scripts/selectCluster.js
       auto-load-enabled: true
-    execution-mode-filter:
```
tgianos, line 32: Were these documented as well? If so that documentation should be deleted

genie-web/src/main/java/com/netflix/genie/web/selectors/AgentLauncherSelectionContext.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param apiJob             whether this job was submitted through the API
+     * @param jobRequestMetadata the job request metadata
+     * @param agentLaunchers     the list of available launchers
+     */
+    public AgentLauncherSelectionContext(
+        @NotBlank final String jobId,
+        final JobRequest jobRequest,
+        final boolean apiJob,
+        final JobRequestMetadata jobRequestMetadata,
+        final Collection<AgentLauncher> agentLaunchers
```
tgianos, line 59: Perhaps we explicitly should make this a `List` so that there an be a defined priority fallback order if so desired?

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
```diff
@@ -2021,6 +2022,72 @@ public JobInfoAggregate getHostJobInformation(@NotBlank final String hostname) {
...
+     */
+    @Override
+    @Transactional(readOnly = true)
+    public Optional<JsonNode> getLauncherExt(@NotBlank final String id) throws NotFoundException {
+        log.debug(
+            "[getLauncherExt] Requested for job {}", id);
+        return this.jobRepository
+            .findByUniqueId(id)
+            .orElseThrow(() -> new NotFoundException("No job exists for the id specified"))
+            .getLauncherExt();
```
tgianos, line 2088: Doing it in this manner will load more data than you likely want as you're reusing the generic `findByUniqueId` method rather than just projecting out the only field you want.

genie-web/src/integTest/resources/application-integration.yml
mprimi: All of this was deleted in a previous PR: https://github.com/Netflix/genie/commit/491387a33aa7128e2a143c77d955323c76c6b581 these default just got left behind

genie-web/src/main/java/com/netflix/genie/web/selectors/AgentLauncherSelectionContext.java
mprimi: I think that logic belongs in the script (similarly to how candidate `Cluster` are not passed in order.
Not having order makes the configuration simpler to override: just add launcher and update the script to pick the right one.

genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/JpaPersistenceServiceImpl.java
mprimi: Ok, I didn't realize, i duplicated this from other persistence that also loads a single field.
I'll update both to just pull the necessary.

PR #1054, opened 2020-11-03
Implementation of Titus agent launcher

mprimi:
Implement an AgentLauncher that leverages [Titus](https://netflix.github.io/titus/) to launch each job in a dedicated container, rather than launching on the same host.

### DIFF
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -102,6 +102,11 @@ Whereas static properties values are bound during application startup and do not
 |/tmp/genie-agent.jar
 |no
 
+|genie.agent.launcher.local.enabled
+|Enable or disable the corresponding launcher.
+|true
+|no
+
 |genie.agent.launcher.local.host-info-expire-after
 |How long after the job information for this host is written into a local cache is it evicted. See
 https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion[Spring Docs]
--- a/genie-docs/src/docs/asciidoc/_properties.adoc
+++ b/genie-docs/src/docs/asciidoc/_properties.adoc
@@ -142,6 +142,106 @@ https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-featu
 |false
 |no
 
+|genie.agent.launcher.titus.enabled
+|Whether the Titus agent launcher is active
+|false
+|no
+
+|genie.agent.launcher.titus.endpoint
+|The Titus API endpoint URI
+|https://example-titus-endpoint.tld:1234
+|no
+
+|genie.agent.launcher.titus.entry-point-template
+|The container entry point expression, placeholder values are substituted at runtime
+|[/bin/genie-agent, exec, --api-job, --launchInJobDirectory, --job-id, <JOB_ID>, --server-host, <SERVER_HOST>, --server-port, <SERVER_PORT>]
+|no
+
+|genie.agent.launcher.titus.owner-email
+|The team email of the owners of the Titus job
+|owners@genie.tld
+|no
+
+|genie.agent.launcher.titus.application-name
+|The name of the application that launches the Titus job
+|genie
+|no
+
+|genie.agent.launcher.titus.capacity-group
+|The name of the capacity group to target
+|default
+|no
+
+|genie.agent.launcher.titus.security-attributes
+|A map of security attributes
+|
+|no
+
+|genie.agent.launcher.titus.security-groups
+|A list of security groups for the job
+|default
+|no
+
+|genie.agent.launcher.titus.i-am-role
+|The IAM role to launch the job as
+|arn:aws:iam::000000000:role/SomeProfile
+|no
+
+|genie.agent.launcher.titus.image-name
+|The name of the container image
+|image-name
+|no
+
+|genie.agent.launcher.titus.image-tag
+|The image tag
+|latest
+|no
+
+|genie.agent.launcher.titus.disk-size
+|The size of the disk volume to attach to the job container
+|10G
+|no
+
+|genie.agent.launcher.titus.network-bandwidth
+|The network bandwidth to allocate to the container
+|7M
+|no
+
+|genie.agent.launcher.titus.retries
+|How many times to retry launch if the job fails
+|0
+|no
+
+|genie.agent.launcher.titus.runtime-limit
+|The maximum runtime of the job
+|12H
+|no
+
+|genie.agent.launcher.titus.genie-server-host
+|The hostname of the Genie server or cluster for the agent to connect to
+|example.genie.tld
+|no
+
+|genie.agent.launcher.titus.genie-server-port
+|The port number of the Genie server or cluster for the agent to connect to
+|9090
+|no
+
+|genie.agent.launcher.titus.healthIndicator-max-size
+|Maximum number of recently launched jobs displayed in the health indicator details
+|100
+|no
+
+|genie.agent.launcher.titus.health-indicator-expiration
+|Maximum time a job is displayed in the health indicator details after launch
+|30m
+|no
+
+|genie.agent.launcher.titus.additional-environment
+|Map of additional environment variables to set for the job
+|default
+|no
+
 |genie.agent.routing.refresh-interval
 |Interval at which individual connections are refreshed
 |3s
--- a/genie-common/src/main/java/com/netflix/genie/common/dto/JobExecution.java
+++ b/genie-common/src/main/java/com/netflix/genie/common/dto/JobExecution.java
@@ -18,6 +18,7 @@
 package com.netflix.genie.common.dto;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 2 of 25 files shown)


genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -142,6 +142,106 @@ https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-featu
...
 |no
 
+|genie.agent.launcher.titus.enabled
+|Whether the Titus agent launcher is active
+|false
+|no
+
+|genie.agent.launcher.titus.endpoint
+|The Titus API endpoint URI
+|https://example-titus-endpoint.tld:1234/api/v3/jobs
```
tgianos, line 157: should you bake the `/api/v3/jobs` part into the code and just let the URL be set since I assume the implementation depends on which version of the API it's using

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -142,6 +142,106 @@ https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-featu
...
+|Maximum number of recently launched jobs displayed in the health indicator details
+|100
+|no
+
+|genie.agent.launcher.titus.health-indicator-expiration
+|Maximum time a job is displayed in the health indicator details after launch
+|30m
+|no
+
+|genie.agent.launcher.titus.additionalEnvironment
```
tgianos, line 245: just a note that the rest of these all have `-` separators and this one is camel case

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Agent launcher that spawns a job in a dedicated container through Titus (https://netflix.github.io/titus/).
+ *
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public class TitusAgentLauncher implements AgentLauncher {
```
tgianos, line 55: By convention should this be `TitusAgentLauncherImpl`?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+        try {
+            final TitusBatchJobResponse titusResponse = this.restTemplate.postForObject(
+                this.titusAgentLauncherProperties.getEndpoint(),
+                titusJobRequest,
+                TitusBatchJobResponse.class
+            );
+
+            if (titusResponse == null) {
+                throw new AgentLaunchException("Failed to request creation of Titus job for job " + jobId);
+
```
tgianos, line 118: Extra line here

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+            } else if (StringUtils.isBlank(titusResponse.getId())) {
+                throw new AgentLaunchException(
+                    "Failed to create titus job for job " + jobId
+                        + " - " + titusResponse.getStatusCode() + ": "
+                        + titusResponse.getMessage()
+                );
+
+            }
+
+            titusJobId = titusResponse.getId();
```
tgianos, line 128: Not a big deal but you're calling `getId` twice between this and the check above. Could reorder it so that this is just set once and then checked for the error case.

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+
+            log.info("Created Titus job {} to execute job {}", titusJobId, jobId);
+
+            MetricsUtils.addSuccessTags(tags);
+
+            final JsonNode launcherExt = JsonNodeFactory.instance.objectNode()
+                .put("launcher_class", this.getClass().getCanonicalName())
```
tgianos, line 135: `this.getClass().getCononnicalName()` could be a static constant as could the various keys used here

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+
+        final TitusBatchJobRequest titusJobRequest = this.createJobRequest(resolvedJob);
+        final Set<Tag> tags = Sets.newHashSet();
+
+        String titusJobId = "unknown";
+
+        final long start = System.nanoTime();
+        try {
+            final TitusBatchJobResponse titusResponse = this.restTemplate.postForObject(
+                this.titusAgentLauncherProperties.getEndpoint(),
```
tgianos, line 111: This could be a constant assigned in the constructor since it's not allowed to change right?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+            return Optional.of(launcherExt);
+
+        } catch (Exception e) {
+            log.error("Failed to launch job on Titus", e);
+            MetricsUtils.addFailureTagsWithException(tags, e);
+            throw new AgentLaunchException("Failed to create titus job for job " + jobId, e);
+        } finally {
+            final long end = System.nanoTime();
+            this.registry.timer(LAUNCH_TIMER, tags).record(end - start, TimeUnit.NANOSECONDS);
+            this.healthIndicatorCache.put(jobId, titusJobId);
```
tgianos, line 149: What removes things from this cache?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+                    this.titusAgentLauncherProperties.getNetworkBandwidth().toMegabytes() * 8 //MB to Mb
+                ),
+                new TitusBatchJobRequest.SecurityProfile(
+                    this.titusAgentLauncherProperties.getSecurityAttributes(),
+                    this.titusAgentLauncherProperties.getSecurityGroups(),
+                    this.titusAgentLauncherProperties.getIAmRole()
+                ),
+                new TitusBatchJobRequest.Image(
+                    this.titusAgentLauncherProperties.getImageName(),
+                    this.titusAgentLauncherProperties.getImageTag()
```
tgianos, line 209: It would be nice if a bunch of these properties were dynamic. In particular these two as if we have a bad image deployment using a fast property to pin it back to previous version quickly would be useful

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+                ),
+                entryPoint,
+                this.titusAgentLauncherProperties.getAdditionalEnvironment()
+            ),
+            new TitusBatchJobRequest.Batch(
+                1,
```
tgianos, line 215: this `1` is a magic number with no real context to it, perhaps a constant would be helpful to know what it means?

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+            this.healthIndicatorCache.put(jobId, titusJobId);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Health health() {
+        return Health.up()
```
tgianos, line 158: I think longer term we'd want to do something that maybe pinged the Titus health endpoint periodically for its health rather than always returning up

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
mprimi: At the moment, they get removed after a fixed amount of time. 
Or they fall off if at capacity and newer ones are added.

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
mprimi: It may even go further than that.. for example you may want to use different images for different job types.

For now, I'm making the smallest step from immutable host to immutable property image.
But i can see this evolving and this property going away completely.

genie-web/src/main/java/com/netflix/genie/web/agent/launchers/impl/TitusAgentLauncher.java
mprimi: Agreed. This launcher has no business being a health indicator, but it needs to be because `AgentLauncher` interface requires it. This kind of info belongs into an `InfoContributor`. 
I may detangle them in a later commit.

PR #1055, opened 2020-11-13
Improve handling of command/cluster selector runtime errors

mprimi:
Motivation: allow a client or user to distinguish between 2 failure cases:

 - The job is submitted with criteria that cannot be satisfied, criteria must be updated for this job to be accepted.
 - The selector encountered a runtime error, resubmitting the job with the same criteria may work

Until now, both cases would be handled the same leading to: 
 * Error `412: Precondition Failed' via API
 * Fatal error resolving job request via Agent
And the job would be marked FAILED with status message JobStatusMessages.FAILED_TO_RESOLVE_JOB

This change better distinguishes the 2 error conditions such that:
 - Selector runtime error returns a `500` via API
 - Selector runtime error produces a retryable error in the agent
 - Selector runtime error sets a different message with the final job status
 - Cluster selector runtime error does not proceed to the next selector (for better determinism)

This allows clients to retry, if appropriate.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
@@ -32,11 +32,13 @@ import com.netflix.genie.common.external.dtos.v4.JobRequest;
 import com.netflix.genie.common.external.dtos.v4.JobSpecification;
 import com.netflix.genie.common.external.dtos.v4.JobStatus;
 import com.netflix.genie.common.internal.exceptions.checked.GenieJobResolutionException;
+import com.netflix.genie.common.internal.exceptions.unchecked.GenieJobResolutionRuntimeException;
 import com.netflix.genie.common.internal.jobs.JobConstants;
 import com.netflix.genie.web.data.services.DataServices;
 import com.netflix.genie.web.data.services.PersistenceService;
 import com.netflix.genie.web.dtos.ResolvedJob;
 import com.netflix.genie.web.dtos.ResourceSelectionResult;
+import com.netflix.genie.web.exceptions.checked.ResourceSelectionException;
 import com.netflix.genie.web.properties.JobsProperties;
 import com.netflix.genie.web.selectors.ClusterSelectionContext;
 import com.netflix.genie.web.selectors.ClusterSelector;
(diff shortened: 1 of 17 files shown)


genie-web/src/main/java/com/netflix/genie/web/services/impl/JobResolverServiceImpl.java
```diff
@@ -466,15 +472,16 @@ private void resolveCluster(final JobResolutionContext context) throws GenieJobR
                         );
                     }
                 } catch (final Exception e) {
-                    selectorTags.add(NO_CLUSTER_RESOLVED_ID);
-                    selectorTags.add(NO_CLUSTER_RESOLVED_NAME);
+                    // Selector runtime exception
+                    // To avoid non-determinism, don't proceed to the next one
```
tgianos, line 476: I'm not sure if this is our best option. We're basically choosing aid in debugging over prioritizing getting the users job to run in any manner possible. I kind of feel like it should be the other way around if a user has given a valid request and we have another option available to us

mprimi: The case here is:
1. Script selector throws runtime exception
2. Fallback to random selector
3. Random resource gets selected 

It's certainly a complex tradeoff with no clear winner.
Non-determinism in general seems pretty bad, so I opted for it here, in the specific case of runtime errors.

Related: why does this resolver iterates over selectors, while the other one does not?
Seems to me like they should both do it or neither. What am i missing?

tgianos: Originally the command one was intended to do the same fallback mechanism however our users stated that they didn't want this behavior. If I recall it was because they felt while cluster selection could afford to be random as generally they're selecting something that just provides resources and the users job will run one way or another in the case of commands changing the actual binaries that a job runs with could have unintended consequences (e.g user gets a buggy version or a version their code isn't compatible with etc). They didn't want to take that risk and this allowed them to use the script to blacklist versions or whatnot. I believe the original design doc has some of this discussion in the notes though I haven't checked.

mprimi: Thanks for the clarification.

Would you like me to reinstate the logic to "in case of script runtime error, proceed to the next selector"?

tgianos: yeah lets for now

mprimi: Done

PR #1057, opened 2020-11-23
Pass dynamic properties into resource selection script

mprimi:
Enable passing parameters to the script resource selection scripts.
Among other things, this enable dynamically tuning parameters such as probabilistic load (e.g., the percentage of job routed to a "canary" cluster or command).


genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/DynamicPropertiesMapCache.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param refreshInterval        the interval after which the map is re-calculated
+     * @param environment            the environment
+     * @param propertiesNamesPattern the pattern that properties need to match to get included in the map
+     */
+    public DynamicPropertiesMapCache(
+        final @NonNull Duration refreshInterval,
```
tgianos, line 62: Generally we put the annotations before `final`

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/DynamicPropertiesMapCache.java
```diff
@@ -0,0 +1,142 @@
...
+     * @return an immutable map of property name and property value, for any property that matches the pattern that has
+     * a non-blank value
+     */
+    public Map<String, String> get() {
+        return this.cache.get(CACHE_KEY);
+    }
+
+    private Map<String, String> loadProperties(@NonNull final String propertiesKey) {
+        // There's only 1 item in this cache, so key is redundant
+        return DynamicPropertiesMapCache.createDynamicPropertiesMap(
```
tgianos, line 87: why the extra redirect to a static method?

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/DynamicPropertiesMapCache.java
```diff
@@ -0,0 +1,142 @@
...
+ * @author mprimi
+ * @since 4.0.0
+ */
+@Slf4j
+public class DynamicPropertiesMapCache {
+
+    private static final String CACHE_KEY = "properties-map";
+    private final Environment environment;
+    private final Pattern propertiesNamesPattern;
+    private final LoadingCache<String, Map<String, String>> cache;
```
tgianos, line 52: Seems like the cache instead of a single key "properties-map" could instead of multiple keys which are the actual unique patterns?

genie-docs/src/docs/asciidoc/_metrics.adoc
```diff
@@ -301,7 +301,7 @@ It will have to be added if one is desired otherwise metrics are just published
 |status, exceptionClass, scriptUri
 
 |genie.services.agentConfiguration.reloadProperties.timer
-|Time taken to refresh the set of properties forwarded to bootstrapping agents
+|Time taken to retrieve the set of properties forwarded to bootstrapping agents
```
tgianos, line 304: nit: These property change fixes aren't really related to the map change commit

genie-docs/src/docs/asciidoc/_properties.adoc
```diff
@@ -483,6 +483,16 @@ of 2.0
 |300000
 |no
 
+|genie.scripts.agent-launcher-selector.properties-pattern
+|The regular expression that filters which properties get passed to the script.
+|^agent-launcher-selector\..*
```
tgianos, line 488: While I understand the desire for flexibility here I worry that we're just asking for trouble or adding complexity to admins by introducing regex. I feel like when I initially saw this pull request it seemed easier to do something like `genie.scripts.agent-launcher-selector.extra-properties` or something and just say anything under that root will be passed to the script. This is similar to what [spring does with hibernate properties](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html#spring.jpa.properties) for example under `spring.jpa.properties.*`

mprimi: The downside of forcing a prefix like that is that this mechanism cannot be used to 'forward' generic properties.

For example, i cannot pass a loglevel change property to a component used by the script, because the loglevel property doesn't match `genie.scripts.agent-launcher-selector....`.

This may not be a big deal for the selector scripts, but this same mechanism is used by the agent configuration service.
And for the agent it seems more important to be able to forward arbitrary properties we dynamically set.

The default configuration for this component seems to hide the complexity (which is there only if needed).
In the most common case, just set `agent-launcher-selector.foo` and it's gonna make its way to the scripts context.
Super easy for canaries, toggles, etc.

I'll add some doc to to clarify usage and tuning.

tgianos: yeah but do what hibernate does and strip the prefix off and pass the properties in as is... e.g. `genie.blah.script.logging.level.debug` with a prefix of `genie.blah.script` would pass in `logging.level.debug` to the final script

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/DynamicPropertiesMapCache.java
mprimi: The downside of a single shared cache is that all users need to agree on a single eviction policy.
For example: all properties sets must be refreshed every 5 seconds.

Also requires more thinking in terms of concurrency if used and shared by multiple other components.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/DynamicPropertiesMapCache.java
mprimi: This is a utility class that takes snapshots of properties matching a given regex.
And also provides a utility object that does the same but caches.

Maybe i should disentangle the two.

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Isn't that more complexity for whoever is using this?
Rather than setting a property you have to come up with the right prefix, then set it then retrieve a property with a different name in the runtime..

genie-docs/src/docs/asciidoc/_metrics.adoc
mprimi: This documents the slight change of semantic for this metric, I think it's in the right commit.

genie-common-internal/src/main/java/com/netflix/genie/common/internal/util/DynamicPropertiesMapCache.java
mprimi: Not sure why this is even there. We set non-null at package level.

genie-docs/src/docs/asciidoc/_properties.adoc
mprimi: Switched to fixed property prefix that gets stripped before forwarding to agents and scripts.

tgianos (approved):

PR #1072, opened 2021-02-18
Improve agent kill logic

nvhoang:
* Process gets killed after completion should return `SUCCEEDED` result
* Add `destroyForcibly()` after failing to graceful `destroy()` job process
* Set killed by system status properly and detect parent got killed

### DIFF
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
@@ -55,6 +55,7 @@ public class JobProcessManagerImpl implements JobProcessManager {
     private final AtomicReference<KillService.KillSource> killSource = new AtomicReference<>();
     private final AtomicReference<ScheduledFuture> timeoutKillThread = new AtomicReference<>();
     private final AtomicReference<File> initFailedFileRef = new AtomicReference<>();
+    private Boolean isInteractiveMode;
 
     private final TaskScheduler taskScheduler;
 
@@ -80,6 +81,8 @@ public class JobProcessManagerImpl implements JobProcessManager {
         if (!this.launched.compareAndSet(false, true)) {
             throw new IllegalStateException("Job already launched");
         }
+
+        this.isInteractiveMode = interactive;
 
         final ProcessBuilder processBuilder = new ProcessBuilder();
 
@@ -255,8 +258,15 @@ public class JobProcessManagerImpl implements JobProcessManager {
                     return new JobProcessResult
                         .Builder(JobStatus.KILLED, JobStatusMessages.JOB_MARKED_FAILED, exitCode)
                         .build();
-                case API_KILL_REQUEST:
                 case SYSTEM_SIGNAL:
+                    // In interactive mode, killed by a system signal is mostly likely by a user (e.g. Ctrl-C)
+                    return new JobProcessResult
+                        .Builder(JobStatus.KILLED,
+                        isInteractiveMode
+                        ? JobStatusMessages.JOB_KILLED_BY_USER : JobStatusMessages.JOB_KILLED_BY_SYSTEM,
+                        exitCode)
+                        .build();
+                case API_KILL_REQUEST:
                 default:
                     return new JobProcessResult
                         .Builder(JobStatus.KILLED, JobStatusMessages.JOB_KILLED_BY_USER, exitCode)
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
@@ -200,6 +200,23 @@ public class JobProcessManagerImpl implements JobProcessManager {
         }
     }
 
+    private void gracefullyKill(final Process process) throws Exception {
+        final Instant graceKillEnd = Instant.now().plusSeconds(10);
+        process.destroy();
+        while (process.isAlive() && Instant.now().isBefore(graceKillEnd)) {
+            process.waitFor(100, TimeUnit.MILLISECONDS);
+        }
+    }
+
+    private void forcefullyKill(final Process process) throws Exception {
+        final Instant forceKillEnd = Instant.now().plusSeconds(10);
+        // In Java8, this is exactly destroy(). However, this behavior can be changed in future java.
+        process.destroyForcibly();
+        while (process.isAlive() && Instant.now().isBefore(forceKillEnd)) {
+            process.waitFor(100, TimeUnit.MILLISECONDS);
+        }
+    }
+
     /**
      * {@inheritDoc}
      */
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
@@ -67,7 +67,7 @@ public class JobProcessManagerImpl implements JobProcessManager {
      *
      * @param taskScheduler The {@link TaskScheduler} instance to use to run scheduled asynchronous tasks
      */
-    public JobProcessManagerImpl(TaskScheduler taskScheduler) {
+    public JobProcessManagerImpl(final TaskScheduler taskScheduler) {
         this.taskScheduler = taskScheduler;
     }
 
--- a/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
+++ b/genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
@@ -36,6 +36,7 @@ import java.lang.reflect.Field;
 import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -62,7 +63,7 @@ public class JobProcessManagerImpl implements JobProcessManager {
      *
      * @param taskScheduler The {@link TaskScheduler} instance to use to run scheduled asynchronous tasks
      */
-    public JobProcessManagerImpl(TaskScheduler taskScheduler) {
+    public JobProcessManagerImpl(final TaskScheduler taskScheduler) {
         this.taskScheduler = taskScheduler;
     }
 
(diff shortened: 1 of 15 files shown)


genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -62,21 +62,19 @@
...
-        final File jobScript,
-        final boolean interactive,
-        @Nullable final Integer timeout,
-        final boolean launchInJobDirectory
-    ) throws JobLaunchException {
+    public void launchProcess(File jobDirectory,
+                              File jobScript,
+                              boolean interactive,
+                              @Nullable final Integer timeout,
+                              boolean launchInJobDirectory) throws JobLaunchException {
```
tgianos: Please revert the formatting change here

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -104,11 +102,9 @@ public void launchProcess(
 
         this.initFailedFileRef.set(PathUtils.jobSetupErrorMarkerFilePath(jobDirectory).toFile());
 
-        log.info(
-            "Executing job script: {} (working directory: {})",
+        log.info("Executing job script: {} (working directory: {})",
             jobScript.getAbsolutePath(),
-            launchInJobDirectory ? jobDirectory : Paths.get("").toAbsolutePath().normalize().toString()
-        );
+            launchInJobDirectory ? jobDirectory : Paths.get("").toAbsolutePath().normalize().toString());
```
tgianos, line 112: This also seems like an odd formatting change where it pulled the `);` up a line and removed the first break. Generally as a "rule" we've gone with once we need multiple lines for a function call we try to keep one argument per line just for readability and then closing `);` on the subsequent line to delineate blocks

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -170,11 +166,11 @@ public void kill(final KillService.KillSource source) {
      */
     @Override
     public JobProcessResult waitFor() throws InterruptedException {
-        if (!this.launched.get()) {
+        if (!launched.get()) {
```
tgianos, line 205: I personally perfer to leave `this` in for member variables so that if I happen to look at the code in something that doesn't include syntax highlighting I'll know easily where it came from. I guess another way would be to do what some people do with `_` or `m` prefix to member variables but for consistency I just put `this.`

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -200,14 +196,19 @@ public JobProcessResult waitFor() throws InterruptedException {
...
-        // TODO: This doesn't seem quite right to me and is confirmed by an existing test. If the job completes
-        //       successfully but then the system calls kill before this method then the final job status will be
-        //       killed instead of successful. We should discuss this - TJG 9/9/2019
-        if (this.killed.get()) {
-            final KillService.KillSource source = this.killSource.get() != null
-                ? this.killSource.get()
-                : KillService.KillSource.API_KILL_REQUEST;
+        // Check exit code first to see if the job finishes successfully and returns SUCCEEDED as status,
+        // even the job gets a KILL request.
+        if (process != null && exitCode == SUCCESS_EXIT_CODE) {
```
tgianos, line 237: I wonder if we need to synchronize on this kind of check. I know above it dereferences the atomic get but by here we're out of the lock

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -146,19 +148,43 @@ public void launchProcess(File jobDirectory,
...
         final Process process = this.processReference.get();
+        if (process == null) {
+            return;
+        }
 
-        // TODO: This probably isn't enough. We need retries with a force at the end
-        if (process != null) {
+        try {
+            // Grace killing period
+            final Instant graceKillEnd = Instant.now().plusSeconds(10);
```
tgianos: this hardcoded 10 seconds could be a place for improvement via property or at least should be a static constant as it is used multiple times and you'd probably want to change them all and not miss one just cause you didn't scan the code

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -146,19 +148,43 @@ public void launchProcess(File jobDirectory,
...
             process.destroy();
+            while (process.isAlive() && Instant.now().isBefore(graceKillEnd)) {
+                process.waitFor(100, TimeUnit.MILLISECONDS);
```
tgianos: this 100 is also a magic number

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -146,19 +148,43 @@ public void launchProcess(File jobDirectory,
...
+            }
+            if (!process.isAlive()) {
+                log.info("Forcefully killed job process successfully");
+                return;
+            }
+        } catch (Throwable t) {
+            log.warn("Process kill with exception: {}", t.getMessage());
         }
+        log.warn("Failed to kill job process");
```
tgianos, line 180: should probably be `error` level thing if we didn't actually kill it.

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -146,19 +148,43 @@ public void launchProcess(File jobDirectory,
...
+            process.destroyForcibly();
+            while (process.isAlive() && Instant.now().isBefore(forceKillEnd)) {
+                process.waitFor(100, TimeUnit.MILLISECONDS);
+            }
+            if (!process.isAlive()) {
+                log.info("Forcefully killed job process successfully");
+                return;
+            }
+        } catch (Throwable t) {
+            log.warn("Process kill with exception: {}", t.getMessage());
```
tgianos, line 178: should this say `Process failed to kill due to exception` and maybe include the exception so we have the stack?

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImplSpec.groovy
```diff
@@ -546,4 +546,45 @@ class JobProcessManagerImplSpec extends Specification {
         cleanup:
         threadPoolScheduler.shutdown()
     }
+
+    def "Force kill diehard process"() {
```
tgianos, line 552: I can't really comment on the commit directly but the commit message is missing a space

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImplSpec.groovy
```diff
@@ -546,4 +546,45 @@ class JobProcessManagerImplSpec extends Specification {
...
+        this.manager.kill(KillService.KillSource.API_KILL_REQUEST)
+
+        then:
+        noExceptionThrown()
+
+        when:
+        JobProcessResult result = this.manager.waitFor()
+
+        then:
+        result.getFinalStatus() == JobStatus.KILLED
```
tgianos, line 594: Just a note that according to code coverage this test doesn't cover the exception cases if you were wanting to test them

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -187,6 +177,23 @@ public void kill(final KillService.KillSource source) {
         log.warn("Failed to kill job process");
     }
 
+    private void gracefullyKill(final Process process) throws Exception {
+        final Instant graceKillEnd = Instant.now().plusSeconds(10);
```
tgianos, line 184: Magic numbers for this and the 100 below

genie-agent/src/test/groovy/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImplSpec.groovy
```diff
@@ -547,12 +548,19 @@ class JobProcessManagerImplSpec extends Specification {
         threadPoolScheduler.shutdown()
     }
 
+
     def "Force kill diehard process"() {
         def future = Mock(ScheduledFuture)
 
+        Path touchedFile = temporaryFolder.resolve("touchz")
+
+        String script = new ClassPathResource("slowlyDie.test.sh")
```
tgianos, line 557: Not sure if you're aware but Groovy has nice utilities for reading files... I think there's just a simple `File.text` API that will read the contents into a string

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -55,6 +55,7 @@
     private final AtomicReference<KillService.KillSource> killSource = new AtomicReference<>();
     private final AtomicReference<ScheduledFuture> timeoutKillThread = new AtomicReference<>();
     private final AtomicReference<File> initFailedFileRef = new AtomicReference<>();
+    private Boolean isInteractiveMode;
```
tgianos, line 58: if this never changes after object is created it should be final

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -255,8 +258,15 @@ public JobProcessResult waitFor() throws InterruptedException {
                     return new JobProcessResult
                         .Builder(JobStatus.KILLED, JobStatusMessages.JOB_MARKED_FAILED, exitCode)
                         .build();
-                case API_KILL_REQUEST:
                 case SYSTEM_SIGNAL:
+                    // In interactive mode, killed by a system signal is mostly likely by a user (e.g. Ctrl-C)
+                    return new JobProcessResult
+                        .Builder(JobStatus.KILLED,
+                        isInteractiveMode
+                        ? JobStatusMessages.JOB_KILLED_BY_USER : JobStatusMessages.JOB_KILLED_BY_SYSTEM,
```
tgianos, line 266: Generally for ternary operators we've gone with this kind of formatting if it is unable to fit on a single line:
```
isInteractiveMode
    ? JobStatusMessages.JOB_KILLED_BY_USER
    : JobStatusMessages.JOB_KILLED_BY_SYSTEM;
```

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/JobSetupServiceImpl.java
```diff
@@ -710,6 +710,12 @@ String composeScript(final JobSetupServiceProperties jobSetupProperties) {
                 .append("# Launch the command")
                 .append(NEWLINE)
                 .append(this.commandLine).append(" <&0 &").append(NEWLINE)
+                // Spawn a new child process to watch if parent process dies, if so kill command forcefully
+                .append("pid=$!").append(NEWLINE)
+                .append("ppid=$$").append(NEWLINE)
+                .append("{ while kill -0 $ppid &> /dev/null; do sleep 30; done; "
```
tgianos, line 716: Do you think 30s is too long of a gap? Should it possibly be shorter? We also could make this a property at some point

nvhoang (commented):
I addressed your comments @tgianos . PTAL!

PR #1074, opened 2021-02-25
Add spot to plug-in some custom Titus modifications

tgianos:
Needed way to manipulate some of the security attributes of a Titus job request. This will let us (and others) do that specific to whatever needs they have in their Titus installation environment.

Other minor changes included in separate commits


nvhoang (approved):
LGTM. I think it's reasonable for an initial interface.

PR #1081, opened 2021-03-08
Upgrade to netflixoss 9.2.2 to publish candidates to maven central

rpalcolea:
Hi folks,

This upgrades our netflixoss plugin to allow publishing candidates to maven central via:

* `netflixossAltCandidateRepo=false` (to support the old flag)
* `netflixossPublishCandidatesToMavenCentral=true`

Here is a sample project using this change: https://github.com/Netflix/netflixoss-sample-project/runs/2059598228?check_suite_focus=true

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ buildscript {
 
 plugins {
     id "com.github.kt3k.coveralls" version "2.10.2"
-    id "nebula.netflixoss" version "9.1.0"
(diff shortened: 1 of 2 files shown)


tgianos:
@rpalcolea Should was just go ahead and switch to the new flag during this pr as well?

rpalcolea:
done @tgianos

tgianos (approved):

PR #1082, opened 2021-03-11
Fix state bug for job group info object

tgianos:
Previously the Titus request was immutable so sharing a static job info object wasn't a problem. Now that we allow the request to be mutated, particularly by the adapter implementation, this shared object represents a problem. This commit simply switches to creating a new JobGroupInfo object for every request so that state is isolated and not reused and subject to race conditions between request threads.


nvhoang (approved):
LGTM. 🚀

PR #1086, opened 2021-03-12
Use GitHub actions for publishing

rpalcolea:
We haven't explore using services such as postgres / mysql for tests but this change is one step closer to be on GHA

The idea is to replicate the logic from Travis to GHA

Also, GHA supports long outputs and allows to download zip files so we can provide more insight to JFrog when the issue shows up

thoughts @tgianos ?

I can spend some time early next week on the integration test side of things so we can consolidate the PR experience


rpalcolea:
Disregard this for now. It needs more work for docker and services for integration tests

PR #1087, opened 2021-03-12
Use GitHub actions

rpalcolea:
This replaces TravisCI with Github Actions

Interesting bits:

# Service containers

This uses [Service Containers](https://docs.github.com/en/actions/guides/about-service-containers) to bootstrap mysql and postgres. My understanding is that changing configuration is not as easy as with TravisCI services. For this reason, I added the MySQL config as a build step:

```
      - name: Configure MySQL database
        run: |
          sudo apt-get install -y mysql-client
          mysql --host 127.0.0.1 --port 3306 -uroot -ppassword -e "SET GLOBAL innodb_file_format = barracuda;"
          mysql --host 127.0.0.1 --port 3306 -uroot -ppassword -e "SET GLOBAL innodb_file_per_table = 'on';"
          mysql --host 127.0.0.1 --port 3306 -uroot -ppassword -e "SET GLOBAL innodb_large_prefix = 'on'"
          mysql --host 127.0.0.1 --port 3306 -uroot -ppassword -e "show variables like '%innodb_file%';"
```

This is the replacement for:

```
- cat "$TRAVIS_BUILD_DIR/travis/config/mysql/my.cnf" | sudo tee -a /etc/mysql/my.cnf
- sudo service mysql restart
```

In addition, learned that the services use official Dockerhub images. So in this case, postgres and mysql can not be initialized without passwords. More info in:

* https://github.community/t/postgres-with-an-empty-password/18171
* https://hub.docker.com/_/postgres
* https://hub.docker.com/_/mysql

I used `password` for `POSTGRES_PASSWORD` and `MYSQL_ROOT_PASSWORD` env variables. Also modified `application-db-postgresql.yml ` and `application-db-mysql.yml ` to reflect this

`MYSQL_DATABASE` and `POSTGRES_DB` allows to bootstrap the initial Databases so no need to create them as a step.


.travis.yml
```diff
@@ -1,48 +1,8 @@
...
-addons:
-  postgresql: '9.6'
 language: java
 jdk:
-- openjdk8
-services:
-- mysql
-- postgresql
-- docker
+  - openjdk8
```
rpalcolea, line 4: We can remove this file and disable in TravisCI.com once we move this to actions

rpalcolea:
Hi @tgianos 
Wondering if we could introduce this change to move away from TravisCI and then work on getting slack notifications internally

tgianos (approved):

rpalcolea:
closing in favor of https://github.com/Netflix/genie/pull/1113

PR #1089, opened 2021-03-16
Add retry logic to Titus API calls

tgianos:
Will retry on 408 and 503 by default 3 times with exponential backoff. All beans are configurable/overridable


nvhoang (approved):
LGTM 🚀

PR #1090, opened 2021-03-17
Update dependencies

tgianos:
- Update OkHTTP to version matching internal recommendation
- Update SpringFox from 2.x to 3.x
- Move Swagger/SpringFox into its own module that can be ignored or taken as needed


nvhoang (approved):
LGTM 🚀

PR #1091, opened 2021-03-18
Replace JCenter with Maven Central

rpalcolea:


build.gradle
```diff
@@ -3,7 +3,7 @@ import org.springframework.boot.gradle.plugin.SpringBootPlugin
 
 buildscript {
     repositories {
-        jcenter()
+        mavenCentral()
```
tgianos, line 6: Isn't mavenCentral() there be default from Gradle itself @rpalcolea ?

tgianos (approved):

build.gradle
rpalcolea: No, if you don't declare a repository, it won't resolve anything 😓

PR #1092, opened 2021-03-29
Mark job FAILED instead of KILLED

nvhoang:


genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -267,7 +267,7 @@ public JobProcessResult waitFor() throws InterruptedException {
                         .Builder(JobStatus.KILLED,
```
tgianos, line 267: This is what we need changed not the message. The message can stay and actually should stay because it will allow analysis to say the reason it "failed"

nvhoang: Fixed. Thanks.

tgianos (approved):

PR #1093, opened 2021-04-12
Initial Request Tracing Implementation

tgianos:
Initial effort into supporting request tracing in Genie

- Leverage [Spring Cloud Sleuth](https://spring.io/projects/spring-cloud-sleuth) and [Brave](https://github.com/openzipkin/brave)
- Cross process boundaries (server -> agent, agent -> job) via environment variables
- Annotate important spans (job launch, agent state machine)
- Tag job launch and agent exec spans with job id

Part of larger internal platform tracing effort as PoC


tgianos:
@narayaruna @karkum @nvhoang this is a semi-large PR. I've tried to break it down into commits that should be self contained for easier review/understanding so I'd suggest navigating by commit to get a picture of what I did

genie-common-internal/src/main/java/com/netflix/genie/common/internal/tracing/brave/impl/EnvVarBraveTracePropagatorImpl.java
```diff
@@ -0,0 +1,224 @@
...
+     * @see TraceContext.Builder#sampled(Boolean)
+     */
+    static final String GENIE_AGENT_B3_SAMPLED_KEY = "GENIE_AGENT_B3_SAMPLED";
+
+    /**
+     * The key for the active trace id low set of bits.
+     *
+     * @see TraceContext.Builder#traceId(long)
+     */
+    static final String GENIE_JOB_B3_TRACE_ID_LOW_KEY = "GENIE_B3_TRACE_ID_LOW";
```
xiao-chen, line 77: Why call it 'trace_id_low' (instead of 'trace_id'), when B3 just calls it traceId?

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
```diff
@@ -284,6 +282,23 @@ public JobProcessResult waitFor() throws InterruptedException {
...
+
+    private void forcefullyKill(final Process process) throws Exception {
+        final Instant forceKillEnd = Instant.now().plusSeconds(KILL_WAIT_SECS);
+        // In Java8, this is exactly destroy(). However, this behavior can be changed in future java.
+        process.destroyForcibly();
+        while (process.isAlive() && Instant.now().isBefore(forceKillEnd)) {
+            process.waitFor(KILL_CHECK_INTERVAL_MS, TimeUnit.MILLISECONDS);
+        }
+    }
+
```
xiao-chen, line 301: nit: these 2 method moves seem unnecessary

xiao-chen (approved):
LGTM pending 2 minor comments. Thanks for the good work!

genie-agent/src/main/java/com/netflix/genie/agent/execution/process/impl/JobProcessManagerImpl.java
tgianos: Yeah. Intellij auto formatting. oh well

genie-common-internal/src/main/java/com/netflix/genie/common/internal/tracing/brave/impl/EnvVarBraveTracePropagatorImpl.java
tgianos: I went back and forth on this a few times (and it's certainly not set in stone). Ideally I'd have just done `GENIE_B3_TRACE_ID` and used the hex string however the builder for `TraceContext` [doesn't expose the method](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java#L416-L417) which would allow me to reconstitute the context from that single value. Hence I had to break it up and just felt it was a little clearer to use `LOW` and `HIGH` rather than absence of `LOW` representing low. That felt to me like a historical thing where they went from supporting only 64 bit identifiers to both that and 128 bit.

genie-agent/src/main/java/com/netflix/genie/agent/execution/services/impl/grpc/GRpcServicesAutoConfiguration.java
```diff
@@ -63,7 +63,7 @@
     @Bean
     @Lazy
     @ConditionalOnMissingBean(AgentHeartBeatService.class)
-    public AgentHeartBeatService agentHeartBeatService(
+    public GrpcAgentHeartBeatServiceImpl agentHeartBeatService(
```
nvhoang, line 66: Just curious why changing to `*Impl` instead of the interface? Will this change tie it to a specific implementation?

tgianos: In general for Spring it is a better practice to make the return type as specific as possible. This allows it to be used in the most injection scenarios. If for example someone had a `@ConditionalOnBean(GrpcAgentHeartBeatServiceImpl)` on some other bean definition because they explicitly needed that one even though technically there was a bean of this type available Spring would only know it as `AgentHeartBeatService` and thus that condition would fail.

It doesn't tie anything to a specific implementation just makes the type more concrete in the spring context.

PR #1105, opened 2021-06-17
Fixing memory leak issue causes by parked jobs

natadzen:
Fixing the memory leak issue caused by parked jobs not getting cleaned up when exception happens during onComplete.
I've added 2 try-catch blocks: one wrapping around onComplete to specifically catch known exceptions that we saw cause this issue. And second try-catch block is wrapping the whole block of the for loop, in case there are other unknown exceptions we have during cleanup.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobKillServiceImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobKillServiceImpl.java
@@ -210,14 +210,7 @@ public class GRpcJobKillServiceImpl extends JobKillServiceGrpc.JobKillServiceImp
                     final StreamObserver<JobKillRegistrationResponse> observer = this.parkedJobKillResponseObservers
                         .remove(jobId);
 
-                    if (observer != null && !isStreamObserverCancelled(observer)) {
-                        try {
-                            observer.onCompleted();
-                        } catch (final Exception observerException) {
-                            log.error("Got exception while trying to complete streamObserver during cleanup"
-                                + "for jobID {}. Exception: {}", jobId, observerException);
-                        }
-                    }
+                    cancelObserverIfNecessary(observer);
                 }
             } catch (final Exception unexpectedException) {
                 log.error("Got unexpected exception while trying to cleanup jobID {}. Moving on. "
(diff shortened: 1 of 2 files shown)


natadzen:
Need to do an integration test to make sure that isCancelled actually works for real observers. If it doesn't, we can remove that part.

nvhoang (approved):
LGTM. Thanks @natadzen .

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobKillServiceImpl.java
```diff
@@ -204,14 +205,36 @@ public void killJob(
...
-                    jobId
-                );
-                if (observer != null) {
-                    observer.onCompleted();
+            try {
+                if (!this.agentRoutingService.isAgentConnectionLocal(jobId)) {
+                    final StreamObserver<JobKillRegistrationResponse> observer = this.parkedJobKillResponseObservers
+                        .remove(jobId);
+
+                    if (observer != null && !isStreamObserverCancelled(observer)) {
```
tgianos, line 213: You could move the null check and all this logic into your function and just call it `cancelObserverIfNecessary`

genie-web/src/test/groovy/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobKillServiceImplSpec.groovy
```diff
@@ -65,31 +66,39 @@ class GRpcJobKillServiceImplSpec extends Specification {
             getPersistenceService() >> this.persistenceService
         }
         this.service = new GRpcJobKillServiceImpl(dataServices, this.agentRoutingService, this.requestForwardingService)
+
+        this.serviceSpy = Spy(service);
```
tgianos, line 70: given this (I haven't used this feature so don't know) but do we still need the actual `service` as a member or can we make it a local variable within setup?

tgianos (approved):
I would squash the `checkstyle` commit into one of the other commits if possible

genie-web/src/test/groovy/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobKillServiceImplSpec.groovy
natadzen: I kept it in case other tests don't need a Spy, but I think I'll just keep the Spy, and whoever wants a regular object can expose it later.

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcJobKillServiceImpl.java
natadzen: Sounds good!

PR #1109, opened 2021-07-25
refactor: Polish use of JUnit 5

jkschneider:
👋 @tgianos hope all is well! `@ExtendWith` is no longer necessary for `@SpringBootTest`. Also moving a handful of asserts to JUnit 5. Let me know if you'd like to move to AssertJ and we can generate that change as well.

Co-authored-by: Moderne <team@moderne.io>


tgianos:
hey @jkschneider! I think your Junit5 assertions change isn't working. In general as time has gone on we've standardized on AssertJ anyway so if you want to change those files to be AssertJ that would be fine. The build broke as it is anyway as the Junit5 assertions aren't on the classpath for that module I guess.

Hope things are well!

PR #1114, opened 2021-09-20
Split command deactivation into two batched queries

tgianos:
Reduce risk for locking


nvhoang (approved):
LGTM. Thanks @tgianos !

PR #1115, opened 2021-09-20
Split command deactivation into two batched queries

tgianos:
Reduce risk for locking

Cherry picks from #1114 into Master

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/repositories/JpaCommandRepository.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/repositories/JpaCommandRepository.java
@@ -48,6 +48,23 @@ public interface JpaCommandRepository extends JpaBaseRepository<CommandEntity> {
             + ")";
 
     /**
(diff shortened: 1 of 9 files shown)


genie-web/src/main/java/com/netflix/genie/web/data/services/impl/jpa/repositories/JpaCommandRepository.java
```diff
@@ -47,6 +47,23 @@
...
+            + " FROM commands"
+            + " WHERE status IN (:currentStatuses)"
+            + " AND created < :commandCreatedThreshold"
+            + " AND id NOT IN ("
+            + "SELECT DISTINCT(command_id)"
+            + " FROM jobs"
+            + " WHERE command_id IS NOT NULL"
+            + " AND created >= :jobCreatedThreshold"
+            + ")"
+            + " LIMIT :limit";
```
nvhoang, line 65: I was thinking whether we should `ORDER BY` but that'd probably be unnecessary. The commands will be marked as `INACTIVE` and will be excluded in the next batch.

nvhoang (approved):
LGTM. Thanks for the quickfix @tgianos !

PR #1126, opened 2021-12-02
add jobId in the logs of file transfer related activities, whenever possible

nvhoang:


genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
```diff
@@ -453,7 +457,7 @@ private synchronized FileTransfer startFileTransfer(
         ) throws NotFoundException, LimitExceededException {
             // Create a unique ID for this file transfer
             final String fileTransferId = UUID.randomUUID().toString();
-            log.debug(
+            log.info(
```
tgianos, line 460: Why change this level? We should just be able to raise the level of the logger internally to make sure these debug statements are printed, it's going to be noisy

tgianos (approved):
minor comment on log level

genie-web/src/main/java/com/netflix/genie/web/agent/apis/rpc/v4/endpoints/GRpcAgentFileStreamServiceImpl.java
nvhoang: I mostly wanted to see when the transfer starts and when it gets failed/stalled or completed.

tgianos (approved):

PR #1134, opened 2021-12-21
Update log4j to 2.17

loganaden:
This fixes this security issue: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45105


tgianos:
Thanks @loganaden but looks like by now we should be on `2.17.1` as there is still a vulnerability on `2.17.0`. I'll upgrade to that in separate PR.

PR #1141, opened 2022-01-18
Make attachments for job requests optional

tgianos:
For backwards compatibility some internal clients were sending requests to the multipart endpoint but not sending an attachment field. This worked on the 2.3.x line but on 2.5.x it is enforced more strictly and is throwing 400. The proper thing to do would be to make the clients do the right thing but it would be such a long tail that releasing this fix is more effective short term while that logic is slowly phased out.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
@@ -263,7 +263,7 @@ public class JobRestController {
     @ResponseStatus(HttpStatus.ACCEPTED)
     public ResponseEntity<Void> submitJob(
         @Valid @RequestPart("request") final JobRequest jobRequest,
-        @RequestPart("attachment") final MultipartFile[] attachments,
+        @RequestPart(value = "attachment", required = false) @Nullable final MultipartFile[] attachments,
         @RequestHeader(value = FORWARDED_FOR_HEADER, required = false) @Nullable final String clientHost,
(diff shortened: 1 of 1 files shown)


genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
```diff
@@ -263,15 +263,15 @@ public JobRestController(
...
+        @RequestPart(value = "attachment", required = false) @Nullable final MultipartFile[] attachments,
         @RequestHeader(value = FORWARDED_FOR_HEADER, required = false) @Nullable final String clientHost,
         @RequestHeader(value = HttpHeaders.USER_AGENT, required = false) @Nullable final String userAgent,
         final HttpServletRequest httpServletRequest
     ) throws GenieException, GenieCheckedException {
         log.info(
             "[submitJob] Called multipart method to submit job: {}, with {} attachments",
             jobRequest,
-            attachments.length
+            attachments == null ? 0 : attachments.length
```
nvhoang, line 274: Would this be an issue if the attachment is actually null? I assume the `handleSubmitJob` at line `277` can take null just fine.

nvhoang (approved):
LGTM. Thanks @tgianos .

genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
tgianos: This specific line is for the array so it shouldn't matter if the internal attachments are null. The `handle` method was already re-used between both so should be ok but we'll test all this with our repro's before we actually release.

PR #1144, opened 2022-01-20
Align grpc and protobuf dependencies with internal versions

tgianos:
Also put in patch fix to support M1 Macs


nvhoang (approved):
LGTM. Thanks @tgianos .

PR #1147, opened 2022-01-26
Upgrade Server Modules to Java 11

tgianos:
Make Java 11 the baseline for the server modules going forward


AlphaVS-76 (approved):
You could've commented out the unwarranted lines, would've helped if they were required for future references

nvhoang (approved):

PR #1152, opened 2022-02-09
Fix exposing inconsistency in job status outside of persistence API

tgianos:
Within job launch logic there was a helper method which would query the job status and based on the returned value proceed with some logic to either update it or fall back to other logic. This works ok if all requests to the persistence service implementation go to a single cosistent backend. If, however, read only queries go to a read replica which may have lag or some other implementation entirely this breaks down without the service actually knowing why or how.

Moving the logic for this behind the persistence API and letting the launch service only act the returned job status from the source of truth api should fix this problem.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,14 +9,16 @@ buildscript {
 
     dependencies {
         classpath "com.netflix.nebula:gradle-aggregate-javadocs-plugin:3.0.1"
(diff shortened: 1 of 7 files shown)


nvhoang (approved):
LGTM. Thanks for the fix @tgianos !

Issue #1164, opened 2022-03-26
prestosql/presto image is not found in docker hub


SamirAllahverdi:
Hi, when I run docker-compose in 4.2.1-SNAPSHOT, it can not find the image.
How can I run it alternatively?

PR #1166, opened 2022-03-28
Fix Demo

tgianos:
Presto split into presto and trino. Looks like the presto containers are gone from docker. Switch to Trino.

Related to #1164


nvhoang (approved):
LGTM. Thanks for the fix @tgianos .

PR #1168, opened 2022-03-28
Intercept 429 http error code and raise an explicit GenieClientTooManyRequestException exception

nvhoang:
It is up to the client how to handle 429 error code by catching `GenieClientTooManyRequestException`.

### DIFF
--- a/genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
+++ b/genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
@@ -20,6 +20,7 @@ package com.netflix.genie.client.interceptors;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.netflix.genie.client.exceptions.GenieClientException;
+import com.netflix.genie.client.exceptions.GenieClientTooManyRequestsException;
 import com.netflix.genie.common.external.util.GenieObjectMapper;
 import okhttp3.Interceptor;
 import okhttp3.Response;
@@ -38,6 +39,7 @@ public class ResponseMappingInterceptor implements Interceptor {
 
     private static final String NO_MESSAGE_FALLBACK = "No error detailed message in server response";
     private static final String ERROR_MESSAGE_KEY = "message";
+    private static final int HTTP_TOO_MANY_REQUESTS = 429;
 
     /**
      * Constructor.
@@ -66,8 +68,12 @@ public class ResponseMappingInterceptor implements Interceptor {
                             responseBody == null || !responseBody.has(ERROR_MESSAGE_KEY)
                                 ? NO_MESSAGE_FALLBACK
                                 : responseBody.get(ERROR_MESSAGE_KEY).asText();
+                        final int responseCode = response.code();
+                        if (responseCode == HTTP_TOO_MANY_REQUESTS) {
+                            throw new GenieClientTooManyRequestsException(errorMessage);
+                        }
                         throw new GenieClientException(
-                            response.code(),
+                            responseCode,
                             errorMessage
                         );
                     } catch (final JsonProcessingException jpe) {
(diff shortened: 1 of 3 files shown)


genie-client/src/main/java/com/netflix/genie/client/exceptions/GenieClientTooManyRequestException.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package com.netflix.genie.client.exceptions;
+
+/**
+ * An exception class that represents 429 - Too Many Requests received from the server.
+ *
+ * @author andrew
+ * @since 4.0.0
+ */
+public class GenieClientTooManyRequestException extends RuntimeException {
```
tgianos, line 26: Should extend `GenieClientException` and be plural `GenieTooManyRequestsException` or `GenieClientTooManyRequestsException` (the latter gets pretty long)

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
```diff
@@ -55,6 +56,10 @@ public Response intercept(final Chain chain) throws IOException {
         if (response.isSuccessful()) {
             return response;
         } else {
+            if (response.code() == 429) {
```
tgianos, line 59: there should be a constant from either okhttp or UrlConnection you can use for this number.

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
```diff
@@ -55,6 +56,10 @@ public Response intercept(final Chain chain) throws IOException {
...
+                throw new GenieClientTooManyRequestException();
```
tgianos, line 60: There should be an error message in the body you can parse out and use here instead of just bland. Can probably move this down into the below body parsing section and have the if somewhere else before final exception is thrown

genie-client/src/main/java/com/netflix/genie/client/exceptions/GenieClientTooManyRequestException.java
nvhoang: Done. Will go with `GenieClientTooManyRequestsException`.

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
nvhoang: you'd be surprised 429 is not on those lists, unless i am mistaken.

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
nvhoang: Done.

nvhoang (commented):
PTAL! Thanks.

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
tgianos: haha wow. All right, you can make it your own constant if you want but not a huge deal

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
```diff
@@ -66,6 +68,9 @@ public Response intercept(final Chain chain) throws IOException {
                             responseBody == null || !responseBody.has(ERROR_MESSAGE_KEY)
                                 ? NO_MESSAGE_FALLBACK
                                 : responseBody.get(ERROR_MESSAGE_KEY).asText();
+                        if (response.code() == HTTP_TOO_MANY_REQUESTS) {
```
tgianos, line 71: nit: There's now 2 possible calls to `response.code()` here and the line below in the original throw. Up to you if you want to break it out.

tgianos (approved):

genie-client/src/main/java/com/netflix/genie/client/interceptors/ResponseMappingInterceptor.java
nvhoang: Done.

PR #1171, opened 2022-04-29
Initial support for expanding configurable execution environments for jobs

tgianos:
Expose ability for defaults, command defaults and job request overrides for compute resources:
- cpu
- gpu
- memory
- disk
- network

Also create new mechanism that allows admins and users to define sets of container images that should be resolved into a final payload to send to the agent launcher. These images can then be refined (into a Kubernetes pod for example) before launching the final job. TitusAgentLauncherImpl was modified to currently only take the main image but will be extended via adapter internally to launch as something similar to a pod.

Images can have:
- name
- tag
- arguments

Priority order of resolution:
job request -> selected command -> system defaults


enicloom (approved):
LGTM. The commit descriptions are well documented and clearly explained the changes included.

But overall, this is a large pr that is hard to review thoroughly. The high level changes and db migrations make sense. By browsing through the code, there was no obvious thing to call out.

PR #1177, opened 2022-09-15
Enable v4.3.0 in place migration

enicloom:
This diff enables v4.3.0 in place migration to mysql. In order to achieve that, the column type modifications need to be removed.


enicloom:
Closing this pr since it is implemented by #1178

PR #1178, opened 2022-09-15
use converter to switch value between int and long

enicloom:
Instead of changing the memory from long to int in application, we can use a customized converter to achieve the same thing without creating a mess. related to https://github.com/Netflix/genie/pull/1177


akashdw (approved):

enicloom:
moving forward with this approach since it is inline with the future path of genie.

PR #1184, opened 2023-02-01
Files deletion cleanup improvements

xiao-chen:
Break down the file deletion clean up task to perform clean up and look up in smaller trunks.


enicloom:
The implementation LGTM. Performance tests are needed before merge.
This change does break the consistency guarantee during file cleanup since the job submitted during the cleanup process can experience inconsistent states between submitted and executed.

enicloom:
> This change does break the consistency guarantee during file cleanup since the job submitted during the cleanup process can experience inconsistent states between submitted and executed.
This is incorrect. It won't break the consistency guarantee for jobs submitted then executed between cleanups.

xiao-chen:
Thank you @enicloom , yes this will be consistent, but delete (and lookup for deletion) in smaller batches. The distinct queries are still looking for all, which guarantees consistency. Those tables are small enough to not cause problems, but if they do, then a different approach would be needed.

Did performance testing offline on mysql, performance insights looking good.

enicloom (approved):

PR #1191, opened 2023-09-22
add upstream security information pass thru mechanism

enicloom:
This pr adds a way for the clients that use the genie java-client to pass through customized security information.

Ideally, the security or authorization should be handled by the framework rather than app logic. However, due to the fact that the upstream clients of genie normally are some type of scheduler that runs job on user's behalf. The job authorization may need information beyond the client's own identity.


bhou2 (approved):
LGTM. The code is clear and clean.

PR #1193, opened 2023-11-15
bump spring_boot_version to 2.7.+

bhou2:
This PR is to migrate Genie OSS SBN from 2.6.7 to 2.7.+ (i.e., bump spring_boot_version to 2.7.+). This migration is not backwards compatible (see https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes). 

In particular, for Genie: (1) H2database from 1.4 to 2.1.; (2) MySQL from 5.7 to 8.0; (3) PostgreSQL from 9.6 to 10; (4) Flyway from 7.x to 9.x. Accordingly, the related sql scripts / implementations, configurations, and dependencies need upgrades.

Tests have been done: (1) local gradle build; (2) genie build on github actions; (3) deployment and tests on personal genie stack.


genie-web/src/main/resources/genie-web-defaults.yml
```diff
@@ -141,7 +141,7 @@ spring:
       repositories:
         enabled: false
   datasource:
-    url: jdbc:h2:mem:genie
+    url: jdbc:h2:mem:genie;MODE=MySQL;DB_CLOSE_DELAY=-1;DATABASE_TO_LOWER=TRUE;CASE_INSENSITIVE_IDENTIFIERS=TRUE;NON_KEYWORDS=value,limit;
```
stevie9868, line 144: I assume our test case is comprehensive, and since all test cases are passed, I think it should be safe but want to bring this up again to double confirm that there should have no risk with such change.

stevie9868 (approved):
Left a small comment, overall lgtm!

genie-web/src/main/resources/genie-web-defaults.yml
bhou2: Yes. It is safe. This change is required as H2 is upgraded from 1.4 to 2.1.

PR #1199, opened 2024-02-20
Save scheduler job metadata entries to env

jzhuge:
Save scheduler job metadata entries to env so that they are available for Spark run.py.


jzhuge:
TODO

- Add unit tests
- Remove the hardcoded suffix

PR #1211, opened 2024-04-08
genie client integration tests: throw the exception of setting up the docker container

bhou2:
genie client integration tests: throw the exception of setting up the docker container


zdong2 (approved):
LGTM

PR #1212, opened 2024-04-27
Add sanity check on attachment filename

bhou2:
When using the local filesystem to store the job attachments, the attachment filename should not contain '/' as this allows saving and potentially executing files at any path of the genie server. To avoid this security issue, we reject the attachment if its filename is illegal, i.e., containing '/'.


zdong2 (approved):
LGTM

PR #1215, opened 2024-05-01

bhou2:
In addition to blocking the char / in the filename, we should also block the char \ as it can be used in the window environment as a part of file path.


genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalFileSystemAttachmentServiceImpl.java
```diff
@@ -94,9 +94,9 @@ public Set<URI> saveAttachments(
                 final long attachmentSize = attachment.contentLength();
                 final String filename = attachment.getFilename();
 
-                if (filename != null && filename.contains("/")) {
+                if (filename != null && (filename.contains("/") || filename.contains("\\"))) {
                     throw new IllegalAttachmentFileNameException("Attachment filename " + filename + " is illegal. "
-                        + "It should not contain the char: /.");
+                        + "It should not contain the char: / or \\.");
```
zdong2, line 99: nit: maybe `"Filenames should not contain '/' or '\\'."`  would be more clear?

zdong2 (approved):
LGTM

PR #1216, opened 2024-05-03

bhou2:
The purpose of checking the attachment filename is to avoid the filename to be . or .. or containing / or \. So that users cannot put the attachments to any location via a hacking manner.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalFileSystemAttachmentServiceImpl.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalFileSystemAttachmentServiceImpl.java
@@ -27,6 +27,7 @@ import com.netflix.genie.web.services.AttachmentService;
 import org.springframework.core.io.Resource;
 
 import javax.annotation.Nullable;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


Shagunflix (approved):
fix looks good

genie-web/src/main/java/com/netflix/genie/web/services/impl/LocalFileSystemAttachmentServiceImpl.java
```diff
@@ -94,9 +95,18 @@ public Set<URI> saveAttachments(
...
-                        + "Filenames should not contain / or \\.");
+                if (filename != null) {
+                    if ((filename.contains("/") || filename.contains("\\") || filename.equals(".") || filename.equals(".."))) {
+                        throw new IllegalAttachmentFileNameException("Attachment filename " + filename + " is illegal. "
+                            + "Filenames should not be . or .., or contain /, \\.");
+                    }
+
+                    final File file = new File(String.valueOf(attachmentsBasePath), filename);
+                    if (!file.getCanonicalPath().startsWith(String.valueOf(attachmentsBasePath))
+                        || file.getCanonicalPath().equals(String.valueOf(attachmentsBasePath))) {
```
zdong2, line 106: Can we add a test case for this?

bhou2: Updated. Thanks.

zdong2 (approved):
LGTM

PR #1217, opened 2024-05-07

bhou2:
Revise the check on attachment filename from denying `filename.equals("..")` to `filename.contains("..")`


Shagunflix (approved):
changing from filename.equals("..") to filename.contains("..")

PR #1223, opened 2025-01-27
Genie agent JDK17 upgrade

bhou2:
Genie agent JDK17 upgrade

### DIFF
--- a/genie-agent/build.gradle
+++ b/genie-agent/build.gradle
@@ -1,4 +1,5 @@
 import org.apache.tools.ant.filters.ReplaceTokens
+import org.gradle.plugins.ide.idea.model.IdeaLanguageLevel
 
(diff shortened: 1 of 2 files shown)


genie-agent/build.gradle
```diff
@@ -1,4 +1,5 @@
 import org.apache.tools.ant.filters.ReplaceTokens
+import org.gradle.plugins.ide.idea.model.IdeaLanguageLevel
 
 apply plugin: "java-library"
```
zdong2, line 4: `genie-agent` is a lib but not an app. IIUC it's only consumed by genie-netflix-agent, which is already on JDK17, so this change should be fine, but testing is still required before the release.

zdong2 (approved):
LGTM

PR #1225, opened 2025-02-04
Upgrade genie-common-internal to spring cloud aws 3

bhou2:
1. upgrade io.awspring.cloud:spring-cloud-aws-dependencies to 3.2.1

### DIFF
--- a/genie-agent-app/build.gradle
+++ b/genie-agent-app/build.gradle
@@ -25,6 +25,14 @@ license {
     exclude "*.txt"
 }
 
+dependencyManagement {
+    imports {
+        mavenBom "com.amazonaws:aws-java-sdk-bom:1.12.99"
+        mavenBom "io.awspring.cloud:spring-cloud-aws-dependencies:3.2.1"
+        mavenBom "io.zipkin.brave:brave-bom:5.13.3"
+    }
+}
+
 dependencies {
     /*******************************
      * Implementation Dependencies
--- a/genie-agent/build.gradle
+++ b/genie-agent/build.gradle
@@ -23,6 +23,14 @@ java {
     }
 }
 
+dependencyManagement {
+    imports {
+        mavenBom "com.amazonaws:aws-java-sdk-bom:1.12.99"
+        mavenBom "io.awspring.cloud:spring-cloud-aws-dependencies:3.2.1"
+        mavenBom "io.zipkin.brave:brave-bom:5.13.3"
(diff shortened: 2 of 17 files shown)


genie-agent-app/build.gradle
```diff
@@ -25,6 +25,14 @@ license {
     exclude "*.txt"
 }
 
+dependencyManagement {
+    imports {
+        mavenBom "com.amazonaws:aws-java-sdk-bom:1.12.99"
```
iuliiasobolevska, line 30: `com.amazonaws` is AWS SDK V1 and with Amazon's [announcement of end-of-support for AWS SDK for Java v1.x](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/) you'll have to move to V2 eventually
(jFYI, both V1 and V2 will continue being supported with upgrade to Spring Boot for now)

genie-agent/build.gradle
```diff
@@ -23,6 +23,14 @@ java {
     }
 }
 
+dependencyManagement {
```
iuliiasobolevska, line 26: you should be able to define it in the root `build.gradle` to reduce duplication (with configure subprojects each)

genie-agent/build.gradle
```diff
@@ -53,7 +61,6 @@ dependencies {
      *******************************/
 
     implementation("io.grpc:grpc-netty")
-    implementation("io.awspring.cloud:spring-cloud-starter-aws")
```
iuliiasobolevska, line 56: you sure this is safe to drop? for reference, Spring Cloud 3 starter name is `io.awspring.cloud:spring-cloud-aws-starter`

genie-agent-app/build.gradle
```diff
@@ -25,6 +25,14 @@ license {
...
+        mavenBom "io.awspring.cloud:spring-cloud-aws-dependencies:3.2.1"
```
iuliiasobolevska, line 31: you should also 
* update Spring Boot version in https://github.com/Netflix/genie/blob/master/gradle.properties#L23-L24
* either reuse that `spring_cloud_version` var or remove it
* Spring Boot 3 integrates with flyway 10: https://github.com/Netflix/genie/blob/master/gradle.properties#L30, check if you can use Spring Boot BOM recommendation instead of hardcoded version

PR #1245, opened 2025-07-07
Genie Web SBN3 Minor Fixes

bhou2:
This PR is created to -- 
1. handle IllegalStateException of S3AttachmentServiceImpl. The issue is that for large attachment files, `inputStream.available()` can be smaller than the content length of the file to upload and accordingly causes IllegalStateException. This PR is to replace `inputStream.available()`  with the actual content file length to create the request body for S3 putObject.
2. bypass Spring 3 ContentCachingResponseWrapper 2GB limit for transferring large files.

### DIFF
--- a/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
+++ b/genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
@@ -835,15 +835,36 @@ public class JobRestController {
         headers.add(JobConstants.GENIE_FORWARDED_FROM_HEADER, request.getRequestURL().toString());
     }
 
-    private void copyResponseHeaders(final HttpServletResponse response, final ClientHttpResponse forwardResponse) {
+    /**
+     * Copies response headers from the forwarded response to the current response.
+     * This method has package-private access for testing purposes.
+     */
+    void copyResponseHeaders(final HttpServletResponse response, final ClientHttpResponse forwardResponse) {
         final HttpHeaders headers = forwardResponse.getHeaders();
         for (final Map.Entry<String, String> header : headers.toSingleValueMap().entrySet()) {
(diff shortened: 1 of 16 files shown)


genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
```diff
@@ -835,15 +835,32 @@ private void copyRequestHeaders(final HttpServletRequest request, final ClientHt
         headers.add(JobConstants.GENIE_FORWARDED_FROM_HEADER, request.getRequestURL().toString());
     }
 
-    private void copyResponseHeaders(final HttpServletResponse response, final ClientHttpResponse forwardResponse) {
+    /*package-private*/ void copyResponseHeaders(final HttpServletResponse response, final ClientHttpResponse forwardResponse) {
```
zdong2, line 838: Do we need `/*package-private*/` here?

genie-web/src/main/java/com/netflix/genie/web/services/impl/S3AttachmentServiceImpl.java
```diff
@@ -197,25 +197,35 @@ private Set<URI> uploadAllAttachments(
 
             URI attachmentURI = null;
 
-            try (InputStream inputStream = attachment.getInputStream()) {
-                // Prepare object
+            try {
```
zdong2, line 200: Do we need this outer try?

zdong2 (approved):
LGTM

genie-web/src/main/java/com/netflix/genie/web/apis/rest/v3/controllers/JobRestController.java
bhou2: Updated to make the comment more formal.

PR #1258, opened 2026-07-07
Upgrade tika-core to address CVE-2025-66516

zdong2:
### Summary
- Bumps `tika-core` 2.3.0 → 3.2.2 to fix CVE-2025-66516
- Bumps `commons-io` 2.11.0 → 2.20.0, a transitive requirement of tika-core 3.2.2 (without it, the pin downgrades commons-io and causes `NoSuchMethodError`)
- Updates `JobRestControllerIntegrationTest#testResponseContentType`: Tika 3.2.2 fixes a UTF-8 detection bug, so a non-UTF-8 test file is now correctly classified as `application/octet-stream` instead of `text/plain`. Only affects the `Content-Type` header for served job files; charset is still forced to UTF-8 regardless.

### Test plan
- [x] `./gradlew build` passes
- [x] `genie-netflix-smoke-tests` passes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,7 +137,7 @@ configure((Set<Project>) ext.javaProjects) {
                 entry "converter-jackson"
             }
             dependency("com.google.guava:guava:20.0")
-            dependency("commons-io:commons-io:2.11.0")
+            dependency("commons-io:commons-io:2.20.0")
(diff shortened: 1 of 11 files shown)


bhou2 (approved):
LGTM. Thanks.

